PR #43, opened 2022-11-12
Add basic intellij plugin

jacksonthea:


ggleyzer (commented):
Please remove everything we don't know we neeed

.idea/misc.xml
```diff
@@ -4,7 +4,7 @@
   <component name="FrameworkDetectionExcludesConfiguration">
     <file type="web" url="file://$PROJECT_DIR$" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_19" default="true" project-jdk-name="azul-18" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_18" default="true" project-jdk-name="azul-18" project-jdk-type="JavaSDK" />
```
jacksonthea, line 7: I don't know why the jdk version would be different from the language level. How's that possible?

plugin_intellij/src/main/java/org/xtclang/sdk/language/EcstasyFileType.java
```diff
@@ -0,0 +1,45 @@
+package org.xtclang.sdk.language;
+
+import com.intellij.openapi.fileTypes.LanguageFileType;
```
ggleyzer, line 3: Please take at any .java in javatools project to seeur coding conventions (e.g. org.xvm.compiler.Compiler.java). We are usually separating the import packages with a new line, trying not to use wildcard imports unless it's more than 20 imports from the same package and comment anything that is not absolutely obvious.

ggleyzer (approved):

jacksonthea:
Cleaned up the imports/packages. Want to take one more look?

PR #45, opened 2022-11-13
Turn build searchable options off

jacksonthea:
This option was creating a build warning, so we'll turn it off for now because we don't need it yet.


ggleyzer (commented):
I think the uiDesigner.xml file got here by mistake; the rest looks good

ggleyzer (approved):

PR #49, opened 2023-03-15
Buildsystem tweaks and encodings

lagergren:
The default build did not work on my machine, as it was not set up for UTF-8 in its environment. Changed the build to force UTF-8 encodings, to minimize the need for external configuration for a new user.

Also added some "best practice" gradle.properties to speed up the build, cache it, and make it parallel. 

Also generated and checked in gradle wrapper scripts, which is another "best practice", with the same purpose as the changes above: make the system buildable on a virgin OS install without any external tools, and explicitly control which versions of the build tools should be downloaded and used.


ggleyzer (approved):
Looks good

ggleyzer (approved):

ggleyzer (commented):
I'm not sure I understand the implications; we used to have those files in the repo and then removed them. Let's dicsuss...

lagergren:
Sure thing, let's talk about the implications you wondered about in a bit.  I can also remove any changes you feel just add redundance.  I'll ping back in a bit.

PR #63, opened 2023-05-20
Add a random seed to Compiler to help with testing.

cliffclick:
Currently only shuffles the file visitation order.
In the future, could shuffle any collection order where the final compiler results are invariant.
Added compiler option "-seed=N". Defaults to 0 seed and no shuffle if missing.

### DIFF
--- a/javatools/src/main/java/org/xvm/tool/Launcher.java
+++ b/javatools/src/main/java/org/xvm/tool/Launcher.java
@@ -662,6 +662,14 @@ public abstract class Launcher
             }
 
         /**
+          * @return the seed for psuedo-random
+          */
+        public int seed()
+            {
+              return values().get("seed") instanceof Integer i ? i : 0;
+            }
+
+        /**
          * Parse the command line arguments into
          *
          * @param asArgs  the command line arguments to parse
@@ -2229,6 +2237,13 @@ public abstract class Launcher
         {
         File[] aFile = dir.listFiles();
         Arrays.sort(aFile, Comparator.comparing(File::getName));
+        // Psuedo random shuffle files, in an effort to trigger bugs
+        int seed = options().seed();
+        if( seed!=0 ) System.err.println("Compiling with seed="+seed);
+        for( int i=0; i<aFile.length; i++ ) {
+          int j = (i+seed)%aFile.length;
+          File tmp = aFile[i];  aFile[i] = aFile[j];  aFile[j] = tmp; // Swap i and j
+        }
         return aFile;
         }
 
--- a/xdk/build.gradle.kts
+++ b/xdk/build.gradle.kts
@@ -48,6 +48,7 @@ val bridgeLib       = "$javaDir/javatools_bridge.xtc"
 
 val distDir         = "$buildDir/dist"
 
+val seed            = "" // -seed=N to change the random seed, or blank ("") for no seed
 val xdkVersion      = rootProject.version
 var distName        = xdkVersion
 val isCI            = System.getenv("CI")
@@ -105,7 +106,7 @@ val compileEcstasy = tasks.register<JavaExec>("compileEcstasy") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "$ecstasyMain/x/ecstasy.x",
          "$turtleMain/x/mack.x")
@@ -129,7 +130,7 @@ val compileAggregate = tasks.register<JavaExec>("compileAggregate") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -148,7 +149,7 @@ val compileCollections = tasks.register<JavaExec>("compileCollections") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -167,7 +168,7 @@ val compileCrypto = tasks.register<JavaExec>("compileCrypto") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -186,7 +187,7 @@ val compileNet  = tasks.register<JavaExec>("compileNet") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -206,7 +207,7 @@ val compileJson = tasks.register<JavaExec>("compileJson") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -225,7 +226,7 @@ val compileOODB = tasks.register<JavaExec>("compileOODB") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -244,7 +245,7 @@ val compileIMDB = tasks.register<JavaExec>("compileIMDB") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -264,7 +265,7 @@ val compileJsonDB = tasks.register<JavaExec>("compileJsonDB") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -284,7 +285,7 @@ val compileWeb  = tasks.register<JavaExec>("compileWeb") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -304,7 +305,7 @@ val compileWebauth = tasks.register<JavaExec>("compileWebauth") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -324,7 +325,7 @@ val compileXenia = tasks.register<JavaExec>("compileXenia") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
@@ -344,7 +345,7 @@ val compileBridge = tasks.register<JavaExec>("compileBridge") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
          "-L", "$coreLib",
          "-L", "$turtleLib",
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ prj/
 .idea/
 .gradle/
 !gradle-wrapper.jar
+TAGS
--- a/javatools/src/main/java/org/xvm/tool/Compiler.java
+++ b/javatools/src/main/java/org/xvm/tool/Compiler.java
@@ -77,6 +77,7 @@ import org.xvm.util.Severity;
  * <li>{@code -strict} - convert warnings to errors</li>
  * <li>{@code -nowarn} - suppress warnings</li>
  * <li>{@code -verbose} - provide information about the work being done by the compilation process</li>
+ * <li>{@code -seed=n} - set the random seed to n</li>
  * </ul>
  */
 public class Compiler
@@ -558,6 +559,7 @@ public class Compiler
             addOption("o",       Form.File,   false, "File or directory to write output to");
             addOption("qualify", Form.Name,   false, "Use full module name for the output file name");
             addOption("version", Form.String, false, "Use full module name for the output file name");
+            addOption("seed",    Form.Int,    false, "Set the internal compiler random seed for testing purposes");
             addOption(Trailing,  Form.File,   true , "Source file name(s) and/or module location(s) to"
                                                  + " compile");
             }


javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -2229,6 +2237,13 @@ private File[] listFiles(File dir)
         {
         File[] aFile = dir.listFiles();
         Arrays.sort(aFile, Comparator.comparing(File::getName));
+        // Psuedo random shuffle files, in an effort to trigger bugs
+        int seed = options().seed();
+        if( seed!=0 ) System.err.println("Compiling with seed="+seed);
```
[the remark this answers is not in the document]
ggleyzer: I agree on the style. As Cliff said himself earlier, a change in an existing file should follow the current style used. Otherwise it's going to be a mess.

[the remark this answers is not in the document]
lagergren: You are introducing your own sorting logic for the seed, when there is sorting logic in the JDK. 

private final Random random = seed == 0 ? null : new Random(seed)

... 
if (seed != 0) {
    Collections.shuffle(dir, random)
}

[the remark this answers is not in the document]
ggleyzer: On the substance, my concern is that 

- this solution can only cover functionality in Launcher and derived classes, while there is no visibility of "seed" to the rest of the compiler/runtime
- even more importantly there is no visibility into the Ecstasy code, which I think would have to be exposed via some injection

xdk/build.gradle.kts
```diff
@@ -129,7 +130,7 @@ val compileAggregate = tasks.register<JavaExec>("compileAggregate") {
     jvmArgs("-Xms1024m", "-Xmx1024m", "-ea")
 
     classpath(javatoolsJar)
-    args("-o", "$libDir",
+    args("-o", "$libDir", "$seed",
          "-version", "$xdkVersion",
```
lagergren, line 134: Apologies that there is no compileXtc task yet, but I am working on unifying this and should be done very soon (TM) now, after having spent time trying to understand how Gradle really wants things. 

So - either way - this is the master build script for the actual XTC platform. This is the wrong place to add a source of non determinism even if it's a user defined one. 
 
We really don't want to change how we build the actual XTC platform in random ways.  This is another argument that seed is something we add from the unit tests. Even if the unit tests are meta enough to just create the XTC platform with random seeds.

ggleyzer: That is a good point

xdk/build.gradle.kts
```diff
@@ -264,7 +265,7 @@ val compileJsonDB = tasks.register<JavaExec>("compileJsonDB") {
...
```
lagergren, line 269: Except for actually adding several changes for the master build script for XTC, making it possible to run it with a seed - currently something you HARD CODE in the build logic is bad.

1) This is the master build. Its concern is not to be testable
2) Should you still require it to be testable, then use the test configuration . 
3) The seed is a hard coded variable, which means that if we change it, we add a dependency that basically triggers a manual full rebuild and gradle having to update graphs, and redo whatever isn't cached. That's not OK. If I want to pass random seeds to gradle (which is doubtful), that is done through a property -Pseed on the command line, or as an environment variable or as a java SystemProperty or something. Gradle then takes that value and uses it as input the compiler task, or has the compiler task just read it using some standard environment mechanism. 
4) -seed as a first class citizen of compiler arguments, placing it in the same category as -o, -L, -version and other compiler flags that actually have to do with the compilation. Compiler options are documented and public. Documenting that seed exists and publicising it not what we want to do. 
5) If we want to run the compiler on anything, at the moment, in the test suite, so as not to have to deal with gradle, that works fine. Add a unit test class under javatools/src/test/org/vm/compiler/RandomListFilesTest, and have it call compile jobs there... e.g. new Compiler(...arguments..).process() or something. Do not involve anything where a test state causes a project rebuild. There is no end to the trouble that will cause, including unnecessary rebuilds.

ggleyzer: All are valid and in my view warrant a discussion. I was thinking about #4 quite a bit myself, but in terms of providing some more generic "global inject-like" facility (kind of Java's -D flag). All my previous attempts were stricken down by Cam's critics though )

javatools/src/main/java/org/xvm/tool/Launcher.java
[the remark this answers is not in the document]
ggleyzer: Also, I think using of the random would be better. The proposed algorithm doesn't really shuffle much. I wrote a simple test for it:

    module TestSimple
        {
        @Inject Console console;
    
        void run()
            {
            String[] strings = ["a", "b", "c", "d", "e", "f", "g"];
    
            console.print(shuffle(strings, 1));
            console.print(shuffle(strings, 2));
            console.print(shuffle(strings, 3));
            console.print(shuffle(strings, 4));
            console.print(shuffle(strings, 5));
            console.print(shuffle(strings, 6));
            }
    
        String[] shuffle(String[] strings, Int seed)
            {
            strings = strings.reify(Mutable);
            for (Int i : 0 ..< strings.size)
                {
                Int j = (i + seed) % strings.size;
                String tmp = strings[i]; strings[i] = strings[j]; strings[j] = tmp;
                }
            return strings;
            }
        }

Here's the output:

[a, c, d, e, f, g, b]
[b, a, e, f, g, c, d]
[b, c, a, g, d, e, f]
[d, a, b, c, f, g, e]
[c, d, e, a, b, g, f]
[b, c, d, e, f, a, g]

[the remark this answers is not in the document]
ggleyzer: Cam, what do you think of adding `List shuffle(Random random)` to List.x?

xdk/build.gradle.kts
cliffclick: Yeah, well I got the conversation rolling anyways.
- "Write in the local style", totally agree, will change.
- No IDEA integration for me (yet).  Emacs does not grok the Tangosol style, and fights me towards a more normal Java format.  So no "Ctrl-Alt-L and it sorts itself out".
- Using the JDK shuffle: Sounds good to me.  Didn't know it was there.
- "Not random enough".  Disagree; in my experience it doesn't take very much random to trigger these things.  But the JDK shuffle should do a good job in any case.
- Seed logic in the master gradle: Dunno where else to put it, so feedback welcome.
- "NO visibility of seed to other parts".  That wasn't a goal of this PR.  This PR was simply to (1) test the PR process, and (2) more complete file-order shuffle to see if there was another lurker, and (3) start the process (but not complete it) for better testing.

Markus's long comment gets a long reply in another comment.

cliffclick: Markus: 
(1),(2) - keep master build clean of test state: 100% agree
(3) - First class argument citizen "-seed" is bad, 100% agree.  But I don't see a secret "-XXseed=123" parser either.  Not sure we need a secret arg parser.  But still we need a way to inject a seed for testing... which falls under (4a) below.
(5) "There are several listFiles... Extending this mechanism...".  Nah, slippery slope argument, does not apply.  You don't want to extend this mechanism to all things, and neither do I.  This is a specific targeted place where we hit a bug in the past.  If you like, consider it a targeted regression test that happens to use a very flexible mechanism we might want to use in other places in the future.  
(4a) "Instead, add a unit test to ...", first time I got somebody to admit we had a place for unit tests!  This is how I do it for my AA project as well.

(4b) "is supposed to be done" ... 
There's NO "supposed to be done" without a team agreement, and at the moment, I do not agree to the use of mock's where a trivial line of code does the job.  We can have that team-wide discussion and come to a mutual agreement, but we have not yet do so.

cpurdy:
Hard to reason about everything in one go, but let me make a few suggestions to consider:

- We do not want to conflate testing of our Ecstasy code (e.g. class libraries) with the testing of the compiler (Java-based tool chain). So when we're fuzzing (randomizing, shuffling, whatever) the compiler, the purpose there is to automate the fuzzing so that we can pummel the crap out of the compiler (using some form of randomization in a loop from outside the compiler, i.e. compile with seed 1234, compile with seed 1235, etc.) But the goal is always the same: Regardless of the presence of fuzzing or not, and regardless of the value of the random seed, the compiler output should always be identical. So I'd suggest this as "rule number 1": the compiler must withstand fuzzing without impacting its output.
- To get the fuzzing in place, I don't _personally_ mind recompiling the compiler with a different dead-code-elimination result, e.g. based on `static boolean FUZZING = true;` to test with fuzzing vs. `false` for normal (published) builds. I acknowledge that reasonable people may differ with my opinion on this (changing the code to test), and they instead desire to use the same exact binaries to test fuzzing vs. not fuzzing; I'm not putting a thumb on that scale, so whatever we come to agreement on after getting consensus on the end goals, I'm good with.
- Everyone here already groks this next bit, but I want to put it down in writing as a good historical record for some future reader: There needs to be a simple way to pass a seed from the outside (from the test harness) into the compiler, and (as above) I'd personally suggest that it doesn't necessarily have to be something that is possible in published builds. Inside the compiler code (probably in the `Handy` class), we will store the seed in some `public static final int FUZZY_SEED` field, and we will have some singleton `public static final Random FUZZER` object instantiated using that seed. Finally, there are helper functions e.g. `public static List fuzz(List list) { if (FUZZING) { return Collections.shuffle(list, FUZZER); } else { return list; }` eventually called from lots of places. I say "eventually" because we don't have the luxury of just stopping everything to e.g. add fuzzing perfectly everywhere, but we'll get an approach nailed down, add it initially to the obvious places (like the one we just fixed), and then we will have this approach ready to use elsewhere else in the future when we see opportunities to do the same.
- For this specific use, I concur on using the `public static void shuffle(List<?> list, Random rnd)` from Java's `Collections` API. I think we've all written one of those before, so it's nice that it's in the standard libs now.

cliffclick:
"to use the same exact binaries to test fuzzing vs. not fuzzing"...

I at least can add some color here.  You pick your invariant points and test there.  You want the fuzzing to fuzz as much as possible between the invariant points, but otherwise be deterministic.  In the case of a compiler + test program (be it XTC, Java, AA, or COBOL) and the final program result - an obvious pair of invariants is the compiler itself, and the final program result.  So if a fuzz'd compiler produces a fuzz'd binary, which runs on a fuzz'd VM and still produces the same output, that's a very strong result.  Flopping bytecodes about (because XTC parses modules in a random order) should be legit, and shouldn't very the final test code's output.

So yes, you want the compiler to be fuzzed, and produced fuzz'd bytecodes, to run in the fuzz'd XVM to a identical result.

Generally speaking, once the fuzzer has a result, its generally pretty quick to binary search the tool-chain to figure out the interactions.  So "long times" between invariants is fine (from a fuzzer's POV).

lagergren:
Hey!

So regarding shuffling file lists for compiler testing - before I started - here is a code review comment: 

There is an issue with Collections.sort being used as the universal guarantee that every platform and every run will return fields in deterministic order.  It is not a complete fix. 

The issue is, that Collection.sort(files) needs to explicitly use the same locale everywhere in the world to do what we intended. The current sort method will use the default locale. This means that if you are in Germany. An alphabetical sort puts “A and Ä” first in the alphabet, but if you are in Sweden, an alphabetical sort will put Ä late in the alphabet, after, Z and Å.  Thus, there is still no guarantee that Collections.sort in alphabetical order returns the identical order for listFiles of a directory with identical contents at different places in the universe. If you force the sort to use a specific Locale, it doesn’t matter which one, then this will not be a problem. You can pick Turkish for that locale, for all we care, it doesn’t matter, as long as the order is the same for every conceivable run, which they will be, no mater if all your file names are ASCII 7 or Mandarin. 


Regarding the test methodology of something like this, my 2c: 

* Our test functionality requirement that we need to implement this test:  "A call site of listFiles in the Compiler needs to generate random orders for the collection returned, based on some deterministic seed value"

* Current change request problem: If you change value of the seed constant, that you put in xkd/buld.gradle.kts, Gradle will notice that all cached portions of the build now hash differently, and it has to rebuild everything. This is a slow process, compared to an integration test that has a compiler bootstrapped already. So don’t do this.

* To shake loose ordering problems, given the relatively small size of our compile jobs, it is likely that you have to test a rather large number of different seeds and xtc code files, and do a lot of recompilation to build confidence that we really have no ordering issues.  So you likely want to change the seed many times, and rerun compilation, as it’s not very common with specific ordering issues causing bugs. If this is done through some gradle hack of the master build file, you will burn a lot of CPU time just waiting for Gradle to rebuild stuff after you touch one of the things in the build with the way largest number of dependencies. So for every seed you want to run some test compile job, you are looking at tens of seconds.  So don’t do that. 

* If we insist on keeping the seed in the Gradle build file, then it should still certainly not go into the main configuration, but in a test configuration of your choice.  But don’t do this. 

* Hence, messing with Gradle is a bad idea. We want to test a modified compiler on different inputs, and we can do that programmatically instead, by instantiating new Compiler( … args … ) instances in the unit tests and calling Compiler.process(). If you can run 5-10 Gradle recompilations with different seeds in a minute, as described above, you can certainly run 100,000 compilations of the same workload from the unit test this way, since you don’t need to recompute anything or involve Gradle at all.

* Of course, if you still want the unit test to invoke Gradle build on something, that’s perfectly possible from the Unit test class, as it can speak Gradle tooling API an Gradle exists in its classpath by default.   This is still a horrible idea, but it’s possible. Then you could also invoke a Gradle context with extra {  seed = …. Programmatically or from a test build file / build context / test configuration.  But don’t do this. 

* So if you are satisfied that you can create a few compilation jobs for the unit test and running them in different Compiler instances with different seeds, you could e.g. add a package private constructor to the Compiler class that accepts a seed and the standard arguments, marshalling the argument parsing parts to the public constructor. The unit test by convention, is in the same package as the code with JUnit. This means that you have modified the Compiler functionality, which IMHO still is a very bad idea and an anti-pattern, for testing purposes. But at least the rest of the world now can’t see it, and don’t have to know about it. Even better - you have not created anything public, with the implications that it MUST be there forever, marked as deprecated well in advance of changes or removals, and also requires documentation. We really don’t want that. But my point is, that you can now do what you did with the gradle build kts file for the Compiler, from the unit test suite, by just adding a package private way to put the seed into the Compiler class. Or a Random. Or any source of repeatable ordering based on a seed. It doesn’t matter. With this solution, you can also add a @VisibleForTesting annotation, as best practice for people who come after you who will work on the compiler. It’s neighbourly.  But don’t do this! 

* Your hypothetical polymorphic Compiler constructor that takes a seed, should ABSOLUTELY not be public, but it can be package private, which means that only the compiler package and its tests are allowed to call it. Now we have removed all need for randomised order testing being visible to anyone, and avoiding having to document any new compiler flags or public APIs we have created to do this. 

* However, even though we have drastically minimised the complexity and drastically increased the throughput for the test case already, and only added a very small package private constructor to the Compiler, and some private shuffle logic, we do not want to do it this way, because even though we have eliminated and drastically simplified the previous version of reordered collection testing, we are still modifying product code to facilitate testing. In the bullet point above, this would still be accepted as a good encapsulated solution by most code reviewers, including myself, a little bit depending on what it looks like. But if you don’t need to modify the Compiler at all, to be able to test this, that would be ideal. And you can. Typically this involves adding a dependency from the project to Mockito, but ONLY in the test configuration and test environment, like so: 

build.gradke.kts: 

dependencies {
  testImplementation("org.mockito:mockito-core:2.1.0")
}

* Now you have the ability to use Mockito in the unit test. The product will not care about it, try to install it, or plug it in. Just the tests.

* There are several listFiles calls in the Compiler. There are explicit calls, and implicit calls, perhaps even in the JDK, that we don’t know about, as they are called by some API function we use. Through injection, we can replace any return value of any method (in our case listFiles) with a one of our own, known only to the test. So you can now decide if you want to replace all the listFiles calls in the entire Java ecosystem that to shuffle return values, or you can do it for one particular call site in the Compiler, or for every call site in the compiler package or however you want to do it, depending on when feel that the granularity gives you maximum relevant testability. This is how you do it: https://www.toptal.com/java/unit-integration-junit-tests

* Now we have a unit test with throughput thousands of orders of magnitude better than some awkward gradle build bootstrapping. We really don’t want to test Gradle here at all, we want to test the compiler on our x files, for the XDK. Because Gradle itself was not the issue. It would also be enough to test a smaller set of x project compilations (with some dependencies, to provoke ordering issues). Setting that up in Java in the unit test is trivial. If you run 100000 compilations of this with different seeds, and verifying equality between results, ensuring they are identical, this should give us more than enough confidence that we aren’t sensitive to ordering issues anymore. We can’t fully guarantee it of course, but we can feel very very confident about it, if this test does its job.  We have now done all this without adding a single line of code to the Compiler or to the gradle build system, and we have the added benefit that the test methodology is eminently extensible to pretty much everything else that requires return value replacements for verification. This is pretty close to, or identical to, industry best practice, but I mostly care about the hugely decreased complexity. 

/M

lagergren:
I thought of another one that is pretty strictly enforced, wherever I have been these last ten years, and that is forbidding wildcard imports in Java.
It leads to code bloat, that you somehow seem to psychologically avoid, if you have to explicitly import any class that you want to use in your code. There is a lot of people, including me, who feel it’s a code simplifying strategy. It’s very easy to grab anything from the kitchen sink,  even if all you need is a spatula and a knife, so to speak … 

/M

cpurdy:
> forbidding wildcard imports in Java.

Let's put those types of comments onto the thread where we're discussing code style; this thread is already a bit too long to digest :) 

> Your hypothetical polymorphic Compiler constructor that takes a seed, should ABSOLUTELY not be public, but it can be package private, which means that only the compiler package and its tests are allowed to call it. Now we have removed all need for randomised order testing being visible to anyone, and avoiding having to document any new compiler flags or public APIs we have created to do this.

I'm going to backtrack a bit on what I said earlier, because there is a good point that you're trying to make here (that will get lost in the amount of text we've all posted), which is that we don't want this kind of garbage showing up in the build file, in the command line documentation, in the public APIs, etc. This is where I tend to agree with the general idea that Cliff appear to be trying to communicate, which is that a small amount of hyper targeted tests built into the code base itself are quite acceptable if they significantly reduce the complexity surfaced otherwise.

For example (_the bike shed should be blue_), we could as easily use an OS env variable `xvm.internal.test.seed=12345` (or whatever), which is easy enough to add to the ether in test-harness land, and easy enough to pluck from the ether in Java land. It only needs to be doc'd internally, and it's an example of (for lack of a better word) a hack that helps us to achieve our stated goals without introducing compositional entropy. (It is entropy, but it does not become a "mathematical factor" in the total entropy explosion of a growing project.)

Remember, that paragraph ^ is bike-shedding, not a "we must do it this way" explanation. It's important to remember that this is a test area that is not a part of the standard unit testing, nor the pre-checkin testing, nor the CI testing, etc. It's specific to a dedicated fuzzing effort that has a predictable input (the XDK sources) and a predictable output (the XDK image), and we're willing to burn hours or days of CPU time on an occasional basis to validate that we haven't regressed anything since the last big run, but we're not willing to bleed any of the complexity into the public space.

So I take back my earlier suggestion that we consider using a different build and "dead code elimination" to yank the test code out of the regular build. As long as we can avoid the harness mess showing up in the doc, in the command lines, in the APIs, etc., and as long as there's no runtime performance impact, I think we'll be fine having the support for the fuzzing "always baked in".

lagergren:
Let me add just one final leaf thin wall of text to this discussion, and I will stop creating e-mails after that. 

We can certainly do as you suggest, Cam!

The most important thing is that we do not do anything that is visible as publically accessible code, or that changes the behaviour of such code. That would get our built up tech debt to the level of Thread.stop, worst case, in the end. 

Still worth noting, however, is that there is a pretty universal consensus with Java projects these days that mocks/code injection instead of modifying product code for any purpose but writing better product code is the way to go. That is also why JUnit and Maven/Gradle have easily extensible mechanisms to do that from the test suite. 

Our use case is “overriding return values from one or more method calls in the product code to see if said product code breaks”. 

This is *LITERALLY* the use case for code injection on top of JUnit tests. 

I want to make ABSOLUTELY clear that I’m in no way in favour of bending our bodies into strange shapes to conform to a third party test framework that is even aaaalllmost a good fit for what we want to test. A lot of people are in favour of that these days, but I am not one of them. In this case, however, I can’t thing of a more perfect fit than code injection from the test suite. 

I’ll just summarize quickly why I am stubborn with this and then get back to my real job - programming. 

1) This is not going to be the only time we need to add a test that m modifies runtime behaviour for testability. There may be hundreds of those tests in the end. I suspect the compound diff of those 100 tests will have introduced a lot of stuff in the product that are harmless and that we can change at will, if things become too complex, as long as they are not public. This is still a bit of a complex operation. 

2) Not all use cases for modifying values ni the running product for testing purposes work with the “add an environment variable access, or subclass this thing” approach. We will very likely have places where we need to go deep into the internals, for example, into the JDK, where we can’t change the code, and do the same pattern. Hoping that you can subclass any third party class you need to modify for testing, and that its protection is open enough to e.g. inherit some method and replace it with our own, is far from given. This is exactly why Bob Lee started writing Guice in 2004. Code injection can modify virtual all parts of our runtime lifecycle. This is also why Guice pretty much transformed the Java application world. 

3) Regarding testig in general, writing tests should be encouraged at a near-manic level and coverage should be monitored continuously, with the utopian goal of 100%. There is almost only good that can come out of that. However, if all the tests require changes to the product code, even if they are small, non-public, and if we can change them whenever we feel like it, this still adds up. I want my product to be chiefly concerned with running XTC - not checking environment variables everywhere, and having to add package private constructors for stuff that the mission of compiling and running XTC does not need. 

4) This is not going to be the only case where code injection is a perfect fit for a test. For example - the functional test for the Gradle XTC extension does it, even though it’s not merged yet. It was the solution that involved adding the minimum amount of code in the minimum amount of files, and everyone in the world is familiar with the syntax and methodology. As a new XTC developer, I would feel a lot more confident if I add a test with a code replacement by injection, than having to understand if there is some hidden danger to augmenting production code to get the functionality I want. For all I know XTC does secret reflection things with reflection or something, and a package private constructor, or a new subclass (perhaps in a place where the original class was declared final for performance reasons), is a lot harder to understand if it has any potential side effects if you don’t have full familiarity with the XTC design. I am still at the level where I compare fine grained differences with master, if I feel I don’t know if I can change anything and go unpunished. So if I can add my wanted test functionality without modifying production code, there is a risk that I just don’t have the time or inclination to ask for help about something that I’m not sure is side effect free. Or that I have to prove it myself. This is a much higher threshold, and may even be the thing that makes me decide not to get into XTC, because I don’t know if a simple thing will have any subtle side effects. 

5) Finally - I can’t actually find a better way to implement the logic for this test than through code injection from JUnit, when it comes to a) minimum modified files and b) minimum added lines of code. 

I really don’t have a lot more to add here. This is dogfood of which I have eaten container loads, against my will, these last ten years or so. Some of if matters, some of it doesn’t. All that I know is that technical debt is best killed before it is even born. That public String getter and setter that you just added, for a value that probably just has three possible well defined values, might seem like an easy to change thing now, and it is. But it’s no fun anymore three years later, when it has bled into 50 different public APIs written by four different teams. No one certainly has any idea of the value domain for this String anymore. That I can promise you, because there are no comments and there is no “grammar” for anything. Now I need to change that String, which is really an implementation detail, because it’s absolutely critical to conserve footprint. This is a situation I have been in more than once, and unwillingly also having the software architect role for one of those four teams, was immensely irritating, since I had little time to do real work. A significant amount of my time was used to stomp this kind of thing out without changing 50 files, 48 which I am not familiar with and not even in the project for which I am responsible. Because otherwise we could not meet the service level agreements that the salespeople had promised. 

/M

PR #72, opened 2023-05-24
Upgrade Gradle build action to v2.4.2 to avoid future situation were we are vulnerable to Github configuration cache security issue for automated builds

lagergren:
Upgrade Gradle build action to v2.4.2 to avoid future situation were we are vulnerable to Github configuration cache security issue for automated builds


lagergren:
Ready to go in. @2.4.2 resolves and works.

PR #74, opened 2023-05-27
Fix binding parameters to functions with more than two parameters

thegridman:


ggleyzer (approved):

thegridman:
Was there a reason this was closed rather than merged after it had been approved?

PR #75, opened 2023-05-29
Missed one more emacs ignore

cliffclick:


lagergren (approved):

lagergren:
While you are at it, looking at the Emacs master ignore template - do you expect anything else to have to go in here? Might as well add it directly, then...

https://github.com/github/gitignore/blob/main/Global/Emacs.gitignore

cliffclick:
No more that I know of, until I find them.

PR #80, opened 2023-07-07
Updated readme to add section for bleeding edge developers

kabutz:


cpurdy (changes requested):
Two items:

1) The "dist-local" works after you've already done an installation of Ecstasy via the `brew` command. I think we should explain this briefly, i.e. that after installing via brew, it's possible to update that "cellar" copy of the XDK when you build the XDK locally, by adding the gradle target "dist-local".

2) "Ecstacy" should be "Ecstasy"

cpurdy (approved):
Looks good; thanks!

PR #85, opened 2023-07-24
Support for WOFF and WOFF2 media types (fonts)

azzazzel:
This is needed so web apps (UIs) can serve their own fonts (for example containing icons)


ggleyzer (approved):
Looks good)

PR #90, opened 2023-08-13
Add XDK Dockerfile

thegridman:
Added a `Dockerfile` to run the xdk in a container. This is obviously work-in-progress while we decide exactly how we work in containers, but it is a starting point for anyone to try it out.

## To build the image

First build the XDK
```
grade clean build
```
The build the image
```
cd xdk
docker build -t ghcr.io/xtclang/xdk:latest .
```

## Run the XDK in a Container

To run the image
```
docker run -it --rm ghcr.io/xtclang/xdk:latest
```

This will run the image interactively and drop you into a shell. All the XDK executables are on the path, as is Java and Node. 
There was no point doing anything more as the default entry point, as there is nothing to run.

The XDK is installed under the `/xqiz.it` directory.

## Notes 
The image is currently fixed to use Java 17.0.2. It also install Node 18.17.1 as that is what the image from https://github.com/azzazzel/xtc_platform uses. Eventually all these would ideally be parameterised. 

I could have done the whole build of the XDK inside the first part of the Dockerfile too but most people right now will already be building locally and this way you get whatever you have just built in the image.

### DIFF
--- a/xdk/Dockerfile
+++ b/xdk/Dockerfile
@@ -1,0 +1,19 @@
+FROM openjdk:17-slim as Builder
+
+RUN mkdir -p /tmp && mv ${JAVA_HOME} /tmp/jdk
+
+FROM node:18.17.1-slim
+
+ENV JAVA_HOME=/usr/local/jdk
+ENV ECSTASY_HOME=/xqiz.it
+ENV PATH $ECSTASY_HOME/bin:$JAVA_HOME/bin:$PATH
+
+COPY --from=builder /tmp/jdk ${JAVA_HOME}
+COPY build/xdk/ $ECSTASY_HOME/libexec/
+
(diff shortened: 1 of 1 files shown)


lagergren (approved):
Seems to work on both aarch64 and x86 mac and linux respectively for me. Is it a good idea to maybe check in a .dockerignore file too?

Having an extension of this also building the xdk can come later but that can be useful I think. And also staging this and the ui container together with buildah/docker compose. But this is a good start! Maybe document the build and run commands, or maybe you did but I'm on the subway with my cell and might have missed it. 

Feel free to correct the "grade" not "gradle" typo left in the master README.md

lagergren:
Converting this to draft. The world has changed quite a lot since this PR was put up, and I hope to finish the dockerization of the Platform this week. Then we can sync and decide what needs to be added to Dockerize the XDK.

PR #94, opened 2023-08-25
Ported tests to JUpiter API and finally deprecated JUnit

lagergren:
Got rid of JUnit 4 framework in order to smooth the way for the XTC plugin integration, which needs more modern approaches to its testability than JUnit 4, only provided in JUnit 4. 

Typically this can be done with Gradle so that: 
* Existing JUnit 4 tests still execute under JUnit 4, using "useJUnit()" test configuration. This, however, requires build logic that is not 100% shared for the test suites in the various subproject in the XDK.  
* Recommended and needed approach to forward port Junit tests, is to go through the Jupiter API, which is an abstraction with common APIs, no matter which one of it's supported tests suites are under the surface. 
* Thus, and since the change was automatic and IDE driven in nature, I took the liberty of making sure that our unit test payloads follow the Jupiter API, and moved them up to Junit 5 notation.

Tests run and work as before, and use the Jupiter Apis. All code coverage is identical, and we have decoupled JUnit 4 (and JUnit itself) as an explicit dependency from the project. We also no longer need to add JUnit 4 to the implementation configurations in various places.  We also do not need to version the test suite, the bundled/built-in useJUnitJupiterAPI() ensures us this works. 

The most significant change is the sad fact that JUnit 4 assertions and Jupiter API assertions use different parameter orders:

assertXXX("description", "expected", "actual") now has string last:
assertXXX("expected, "actua", "description")

This is the majority of the code line changes.  And smaller things of that nature.  They have all been generated programatically, and It's not hard to be convinced that we have the same test coverage, run the same tests and so on, as we did before. 

I have also taken the liberty to add @Disabled to unit test methods that were checked in as // @Test, because that provides us with metadata about tests still being disabled that we may want to revisit, and it gives us a true tally of which tests we have, disabled / parked /skipped /run or otherwise.

### DIFF
--- a/javatools/src/test/java/org/xvm/runtime/TestConnector.java
+++ b/javatools/src/test/java/org/xvm/runtime/TestConnector.java
@@ -2,7 +2,6 @@ package org.xvm.runtime;
 
 
 import java.io.File;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -16,11 +15,11 @@ import org.xvm.asm.ModuleRepository;
 import org.xvm.asm.ModuleStructure;
 
 import org.xvm.compiler.BuildRepository;
+
 import org.xvm.tool.Compiler;
 import org.xvm.tool.Disassembler;
 
 import org.xvm.util.Handy;
-
 
 /**
  * The connector test.
--- a/javatools/build.gradle.kts
+++ b/javatools/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.gradle.language.base.plugins.LifecycleBasePlugin.BUILD_GROUP
+
 /*
  * Build file for the Java tools portion of the XDK.
  */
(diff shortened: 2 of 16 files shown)


lagergren:
Temporarily converting to draft until I have removed the cruft at the side of the cherry pick from the build-aggregator branch so everything is back to normal. Will request review when this is the case and untag the draft notification.

lagergren:
Ping @ggleyzer @cpurdy - I think this is pretty much ready to go in.  While it doesn't affect anything hugely in master, merging this mostly machine-generated change now makes a lot of sense to me, as I want to keep the first pull request to master for the XTC Gradle plugin as small as possible. In order to do unit and integration tests of the XTC plugin, we need JUnit 5, and some other functionality that is not available in the legacy JUnit 4 configurations.  

The XTC plugin also adds language conventions for Java, among other things, in its build logic, so that all Java configuration settings for all Java subproject are in one place and one place only. I've chosen to include that functionality in the XTC plugin pull request instead, so here, we still have the copied build logic for our few Java projects. But it will soon go away. 

All tests seem to run and pass, and have the same coverage as before (but now also using @Disabled instead of // @Test, so that we get proper statistics of which tests have been written, and how large a fraction of disabled tests we currently have in the build. This helps us remember that we should reactivate these tests if the are relevant (and forward port them if needed), or review and delete them if they are deemed to be unnecessary.

The change request is wide and shallow, mostly machine generated, and every changed line that I don't need to put up as part of the XTC Plugin pull request, so that we can concentrate on its core functionality, during review, will save complexity, so please merge this to master ASAP, if you are happy with it.

javatools/src/test/java/org/xvm/runtime/TestConnector.java
ggleyzer, line 1: I don't think this file needed to change. For whatever reason IDEA re-ordered java.* and org.xvm.* imports in your commit, placing "java" after "org". Practically everywhere else we do it in an alphabetical order - java.* goes first, org.* follows and then static imports in the same order.

javatools_utils/src/test/java/org/xvm/util/ConstBitSetTest.java
ggleyzer, line 1: I don't think this file needed to be changed either. Our current codding convention is "package" followed by two empty lines, followed by the imports, followed by two empty lines again. We may change it in the near future, but there is no reason to change it piece meal. Same goes for other changes.

ggleyzer (changes requested):

ggleyzer (approved):

javatools/src/test/java/org/xvm/runtime/TestConnector.java
lagergren: I'm not sure what's going on with the imports, they ended up as they did on Save, with what I thought was the ecstasy check style.xml configuration.  I can revert them if it's important?

javatools_utils/src/test/java/org/xvm/util/ConstBitSetTest.java
lagergren: Sure, I can restore the import order.  I have never heard about a code standard for anything not Python that recommends more than one blank line for any kind of separation. I'll figure out why IntellIJ did not respect the XTC xml coding standard config.

javatools/src/test/java/org/xvm/runtime/TestConnector.java
lagergren: Changed and pushed.

PR #97, opened 2023-10-17
Added first version of the XTC plugin, and migrated the build.

lagergren:
Added first version of the XTC plugin. Gradle/IDEs now understand native XTC projects, and the XDK and the plugin is build through this plugin, as bootstrapping. 

We are now ready for first rounds of review and testing. 

TODO: Publish correct first artifacts for the XDK and the XTC Plugin for this version. Switch to SNAPSHOT version and document how development works with semantic versions described in the README.md.

TODO: Finish and put up a pull request for the 'examples' repository that uses the XTC plugin, and an XDK dependency to the abovementioned published version.


lagergren:
For some reason the build does not go through completely, because we can't resolve the "mack" module. Not sure why that doesn't work in the CI path, but it works locally with ./gradlew build on my machine. I will investigate tomorrow. 

To purge all build state, use the bin/purge... sh script. 

Suggested test methodology: clone XVM to a new directory, check out this branch and use the ./gradlew build command to build, before trying to import it into an IDE. 

If you import it into IntelliJ, be and anything important is in your ".idea" folder (or stale settings, or references to older XVM versions or parts thereof, it's probably a bad idea to import this branch into that environment. It's also probably a good idea to never rely in ".idea" for any state whatsoever. Those things tend to go away when someone coughs. 

;-) 

Goodnight, guys!

lagergren:
It seems to be working now. Great! Go forth and review, men!

lagergren:
Moved to the collapsed version to avoid rebase horrors for the build kts files that have been very different for three months. Almost there. Closing this. Please don't delete the git branch proper for a bit so I have a history record while we merge.

PR #98, opened 2023-11-08
Replaced raw Collections.EMPTY_LIST with preferred Collections.emptyList()

lagergren:
Fixes 52 of the circa 1200 Java warnings we have.


ggleyzer (approved):
I have not looked at all the files, but I believe you :)

PR #110, opened 2024-01-02
JSON enhancements

thegridman:
This PR is to add some functionality to the JSON module

* A fluent builder API  added to created JSON object and JSON array instances
* Support for JSON Pointer [RFC6901](http://tools.ietf.org/html/rfc6901)
* Support for JSON Patch [RFC6902](https://datatracker.ietf.org/doc/html/rfc6902)
* Support for JSON merge patch [RFC7396](http://tools.ietf.org/html/rfc7396)
* Support for deep merging JSON objects and arrays (not the same as RFC7396)


lib_json/src/main/x/json/mappings/JsonPatchMapping.x
```diff
@@ -0,0 +1,29 @@
...
+ * A mapping for a `JsonPatch`.
+ */
+const JsonPatchMapping
+        implements Mapping<JsonPatch> {
+
+    /**
+     * Construct the JsonPatchOperationMapping.
+     */
+    construct() {
+        this.typeName = $"JsonPatch";
```
ggleyzer, line 11: "$" is not necessary since there is nothing to compute

lib_json/src/main/x/json/mappings/JsonPatchOperationMapping.x
```diff
@@ -0,0 +1,47 @@
...
+ * A mapping for a `JsonPatch.Operation`.
+ */
+const JsonPatchOperationMapping
+        implements Mapping<JsonPatch.Operation> {
+
+    /**
+     * Construct the JsonPatchOperationMapping.
+     */
+    construct() {
+        this.typeName = $"JsonPatch.Operation";
```
ggleyzer, line 11: Ditto

lib_json/src/main/x/json/mappings/JsonPatchOperationMapping.x
```diff
@@ -0,0 +1,47 @@
...
+        this.actionMap = new HashMap();
```
ggleyzer, line 12: "this" is not required, since there is no name collision, but if used here should be used two line down as well. Otherwise it creates some kind of confusion to the reader

lib_json/src/main/x/json.x
```diff
@@ -25,4 +25,39 @@ module json.xtclang.org
...
+     */
+    typedef Array<Doc> as JsonArray;
+
+    /**
+     * A type representing a non-primitive JSON structure.
+     */
+    typedef JsonObject | JsonArray as JsonStruct;
+
+    /**
+     * @return a new instance of a mutable `JsonObject`.
```
ggleyzer, line 45: Trivial thing - we usually don't put the dot ('.') at the end of @return or @param

lib_json/src/main/x/json.x
```diff
@@ -25,4 +25,39 @@ module json.xtclang.org
...
+     */
+    JsonObject newObject() = new ListMap<String, Doc>();
+
+    /**
+     * @return a builder that can produce immutable JSON object instances.
```
ggleyzer, line 50: Ditto

lib_json/src/main/x/json/JsonArrayBuilder.x
```diff
@@ -0,0 +1,173 @@
...
+    @Override
+    protected void merge(Int index, Doc value) {
+        if (index < 0) {
+            add(value);
+        } else {
+            Doc existing = values[index];
+            switch (existing.is(_)) {
+            case JsonObject:
+                if (value.is(JsonStruct)) {
+                    JsonObject o = new JsonObjectBuilder(existing).deepMerge(value).build();
```
ggleyzer, line 122: Why not to inline this? Then you could further simplify using a ternary or `switch (value.is(_)`

lib_json/src/main/x/json/JsonArrayBuilder.x
```diff
@@ -0,0 +1,173 @@
...
+                    set(index, o);
+                } else if (value.is(Primitive)) {
+                    set(index, value);
+                } else {
+                    assert;
+                }
+                break;
+            case JsonArray:
+                if (value.is(JsonStruct)) {
+                    JsonArray a = new JsonArrayBuilder(existing).deepMerge(value).build();
```
ggleyzer, line 132: Ditto

lib_json/src/main/x/json/JsonArrayBuilder.x
```diff
@@ -0,0 +1,173 @@
...
+     * in the range between zero and the current size of the array being built.
+     */
+    @Override
+    protected void mergeObject(JsonObject o) {
+        Map<JsonPointer, Doc> map = new ListMap();
+        for (Map<String, Doc>.Entry entry : o.entries) {
+            JsonPointer pointer = JsonPointer.from(entry.key);
+            Int?        index   = pointer.index;
+            assert index != Null as "Cannot merge JSON Object with non-Int keys into a JSON array";
+            assert index >= 0 && index < values.size as
```
ggleyzer, line 162: This could be simplified to `0 <= index < value.size`

lib_json/src/main/x/json/JsonPointer.x
```diff
@@ -0,0 +1,240 @@
...
+ * An empty JSON Pointer defines a reference to the target itself.
+ */
+const JsonPointer {
+    /**
+     * Private `JsonPointer` constructor.
+     * Pointer instances are created via the static `from` method, which
+     * will properly validate the pointer string and create the correct
+     * pointer chain.
+     */
+    private construct (String pointer, String key = "", JsonPointer? remainder = Null) {
```
ggleyzer, line 15: We usually don't put a space there - `construct(String pointer...`

lib_json/src/main/x/json/JsonPointer.x
```diff
@@ -0,0 +1,240 @@
...
+            case JsonObject:
+                if (Doc child := doc.get(key)) {
+                    if (remainder.is(JsonPointer)) {
+                        return remainder.get(child, supportNegativeIndices);
+                    }
+                    return True, child;
+                }
+                break;
+            case JsonArray:
+@Inject Console console;
```
ggleyzer, line 187: I don't think you meant it to be here :)

lib_json/src/main/x/json/JsonPointer.x
```diff
@@ -0,0 +1,240 @@
...
+     * Ensure the specified index is a positive index into the array
+     *
+     * @param array  the array to check this pointer's index value against
+     *
+     * @return False if the array is empty, or True if this pointer's index is zero or positive in the
+     *         range 0 ..< array.size, or True if the index is negative in the range -array.size >.. -1
+     * @return a valid index into the array
+     */
+    conditional Int getValidIndex(JsonArray array, Boolean supportNegativeIndices) {
+@Inject Console console;
```
ggleyzer, line 218: Ditto

ggleyzer (commented):
It looks like there is still WIP. Please let me know when you're in a good state and we'll push it in

lib_json/src/main/x/json/JsonObjectBuilder.x
```diff
@@ -0,0 +1,139 @@
...
+    protected Doc get(String key) = values[key];
+
+    @Override
+    protected void update(String key, Doc doc) = add(key, doc);
+
+    @Override
+    protected void merge(String key, Doc value) {
+        Doc existing = values[key];
+        switch (existing.is(_)) {
+        case JsonObject:
```
ggleyzer, line 112: I think this could be refactored into "two-column" switch the same way as you did previosely

lib_json/src/main/x/json/JsonPointer.x
```diff
@@ -0,0 +1,233 @@
...
+            default: assert;
+        };
+    }
+
+    /**
+     * Obtain the value from the specified JSON `Doc` at the location
+     * pointed to by this `JsonPointer`.
+     *
+     * @param doc                   the JSON `Doc` to obtain the value from
+	 * @param allowNegativeIndices  support the non-standard use of negative indices for JSON arrays to mean indices
```
ggleyzer, line 168: The formatting is off

ggleyzer (approved):
Short of a couple of tiny things, everything looks great. Please submit when you're ready

lib_json/src/main/x/json/JsonObjectBuilder.x
```diff
@@ -0,0 +1,131 @@
...
+    protected Doc get(String key) = values[key];
+
+    @Override
+    protected void update(String key, Doc doc) = add(key, doc);
+
+    @Override
+    protected void merge(String key, Doc value) {
+        Doc existing = values[key];
+        switch (existing.is(_), value.is(_)) {
+        case (JsonObject, JsonStruct):
```
ggleyzer, line 112: Not sure if it would read better, but just wonder if the compiler is smart enough to merge the cases, for example:
`
 case (JsonObject, JsonStruct):
 case (JsonArray, JsonStruct):
            add(key, new JsonArrayBuilder(existing).deepMerge(value).build());
            break;
case (JsonObject, Primitive):
 case (JsonArray, Primitive):
  case (Primitive, Doc):
            add(key, value);
            break;
`

PR #113, opened 2024-01-19
XTC Kotlin build DSL + XTC aware Gradle builds

lagergren:
There are two known minor outages:

We can't rebuild the Unicode tables, but it's a small fix. Max half a day or less, and we haven't rebuilt them since the last release and can also do so manually.

XtcLauncherTask output stream redirection should allow adding and not just replacing streams, and also redirecting output streams does't work well. Likely an easy fix. We should also significantly simplify the launcher logic in the plugin, as it got a bit complex while still having to handle System.exit, which now is done in a safe manner by first propagating exceptions.

Hence, please go over this and add review comments everywhere you find something weird. It seems to work, though, and I believe I support 95% of your current use cases with no added friction (knock on wood).

For a reference use case, see the platform pull request that uses this.

The TODO file will be removed before merge and entered as separate issues into GitHub issues for this repo, with the hope that there will also be several small and well described starter tasks for anyone who wants to get into XDK platform development or understand XTC in general.


lagergren:
Added a cleanup commit, switched to new branch because I made a git error that I was too scared to resolve with the reflag.

javatools_launcher/src/main/resources/javatools/README.md
ggleyzer, line 1: The content of the README still refers to the "build" directory; need to update it

javatools_launcher/README.md
ggleyzer, line 1: There's also this paragraph that is no longer correct:

(The `build/javatools` directory exists solely for testing the build result; it contains a **fake**
`javatools.jar` with a "main class" that echos the command line arguments. Read the README.md in
that directory for a detailed explanation.)

lagergren:
Closed after cherry picking over changes to https://github.com/xtclang/xvm/pull/114

lagergren:
The last commit broke master after rebasing the build, but not before. Will add it incrementally later, and try to understand why.

PR #114, opened 2024-01-23
XTC Kotlin build DSL through XTC Plugin, XDK artifact publications.

lagergren:
There are two known minor outages:

We can't rebuild the Unicode tables, but it's a small fix. Max half a day or less, and we haven't rebuilt them since the last release and can also do so manually.

XtcLauncherTask output stream redirection should allow adding and not just replacing streams, and also redirecting output streams does't work well. Likely an easy fix. We should also significantly simplify the launcher logic in the plugin, as it got a bit complex while still having to handle System.exit, which now is done in a safe manner by first propagating exceptions.

Hence, please go over this and add review comments everywhere you find something weird. It seems to work, though, and I believe I support 95% of your current use cases with no added friction (knock on wood).

For a reference use case, see the platform pull request that uses this.

The TODO file will be removed before merge and entered as separate issues into GitHub issues for this repo, with the hope that there will also be several small and well described starter tasks for anyone who wants to get into XDK platform development or understand XTC in general.

-- PS: 
This branch was switched out, as for some reason, the last commit from xtc-plugin broke the build after rebasing on master. 

The last commit from xtc-plugin broke build dependencies, after rebasing on master. This isn't apparent from either master or from the xtc-plugin branch, and is likely a subtle build config resolution issue.  

This commit breaks master cd5d94571cb89f9194de1fecf62c230fe22ff866 on top of master commit 382ad3c95aa5fb0a49e67aba0699b9eb13e5e563. 

Will investigate further, but backed it out for now to have something that can still be merged.


build-logic/common-plugins/src/main/kotlin/org.xtclang.build.java.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+    doLast {
+        logger.info("$prefix JVM arguments: $jvmArgs")
+    }
+}
+
+//val assemble by tasks.existing
+val checkWarnings by tasks.registering {
+    if (!getXdkPropertyBoolean(lintProperty, false)) {
+        val lintPropertyHasValue = isXdkPropertySet(lintProperty)
+        if (lintPropertyHasValue) {
```
ggleyzer, line 59: What is the point of these warnings? How should I (or Dima) act on that? If you believe some compiler option needs to be on, let's turn it on

build-logic/aggregator/src/main/kotlin/org.xtclang.build.aggregator.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+
+    private fun checkStartParameterState() {
+        val startParameter = gradle.startParameter
+        with (startParameter) {
+            logger.info("""
+            $prefix Start parameter tasks: $taskNames
+            $prefix Start parameter init scripts: $allInitScripts
+        """.trimIndent())
+
+            if (taskNames.count { !it.startsWith("-") && !it.contains("taskTree") } > 1) {
```
ggleyzer, line 50: I'm pretty sure I don't understand the complexity of the issue, but how hard it is to declare that "build" _mustRunAfter_ "clean", so we can use the most commonly command "gradlew clean build". Even better, I'd like to do "gradlew clean build installLocalDist".

xdk/build.gradle.kts
```diff
@@ -1,602 +1,408 @@
...
+ *   nothing breaks in existing use cases. All in all, it's rather horrible to go in and replace
+ *   parts of an existing home brew (or other package manager) based installation, and should
+ *   really never be allowed, as it leads to bit skew.
+ *
+ * TODO: @aalmiray recommends application plugin run script generation, and that makes sense to me.
+ *   It should be possible to hook up JReleaser and/or something like launch4j to cross compile
+ *   binary launchers for different platforms, if that is what we want instead of the current
+ *   solution.
+ */
+val installLocalDist by tasks.registering {
```
ggleyzer, line 388: This task blew up - I posted the error to the Discord

ggleyzer (dismissed):
The failure of "installLocalDist" really is a show stopper for me

build-logic/common-plugins/src/main/kotlin/org.xtclang.build.java.gradle.kts
lagergren: The point is that we should fix the millions of Java warnings in our Java code base, for example use of Raw types. This is the reminder. The alternative is to spew them out on the screen, and die on error. All other alternatives (silently ignore Java code with lots of warnings) have gotten me, as someone who wants to work with the code base in trouble, repeatedly.  

If you don't want to see the message, use of the semi persistent ways to set the lint property of false in your environment, as described in the documentation.  Add an export of its env variable equivalent to your shell .rc profile or something. This is a default setting that is warranted for several reasons, and we should strive to write warning free Java code.

build-logic/aggregator/src/main/kotlin/org.xtclang.build.aggregator.gradle.kts
lagergren: It deadlocks due to something having to do with the builds still being resolved when starting. It's not a standard or recommended gradle way to run multiple tasks. If you want a talk that does clean build, then write one and call it rebuild. 

mustRunAfter guarantees nothing about task execution, btw.

xdk/build.gradle.kts
lagergren: Do you have a link to a build scan on how it blew up?

lagergren: I had a file resolution issue remaining. I'll add a check that this works to the GitHub CI runner when I have time.

build-logic/aggregator/src/main/kotlin/org.xtclang.build.aggregator.gradle.kts
lagergren: Why would you want to do a "clean build installLocalDist" and not just "installLocalDist"? The dependency resolution should be the same?

build-logic/common-plugins/src/main/kotlin/org.xtclang.build.java.gradle.kts
ggleyzer: I haven't had a single issues caused by ignoring any of those warnings, and Cam and I have been working on this for six years now. My vote is to preserve a status quo and let the build behave as it does today in this regard. I would be happy if we had a flag to turn those warnings on - may be at some point we can hire an internal to go and clean those up :)

build-logic/aggregator/src/main/kotlin/org.xtclang.build.aggregator.gradle.kts
ggleyzer: because I'm doing it about 80 times a day and it would me a bit more annoying

xdk/build.gradle.kts
ggleyzer: I did the command that you recommended, `./gradlew build --scan --stacktrace`, but it looked like it blew up before producing anything else but the output I posted to Discord.

build-logic/common-plugins/src/main/kotlin/org.xtclang.build.java.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    doLast {
+        logger.info("$prefix JVM arguments: $jvmArgs")
+    }
+}
+
+val checkWarnings by tasks.registering {
+    if (!getXdkPropertyBoolean(lintProperty, false)) {
+        val lintPropertyHasValue = isXdkPropertySet(lintProperty)
+        logger.warn("$prefix WARNING: Java warnings are ${if (lintPropertyHasValue) "explicitly" else ""} for project.!")
```
ggleyzer, line 58: Looks like the produced warning is missing a noun:

> WARNING: Java warnings are explicitly for project.!

Did you mean it to say "explicitely turned off"? Also, I think the dot is unnecessary (or an exclamation mark)

ggleyzer:
The xvm build seems to be working fine. Slight challenge for developing using separate stand-alone xtc repositories (platform or examples) is that there are two ways to publish things:
1. Using "installLocalDist", which updates the local brew cellar, making the standalone "xcc" and "xec" commands using the newly minted xdk
2. Using "publishLocal", which update the local maven repository (~/.m2), which allows the "build" command to use the newly minted compiler, and run it via "xtcRun" task, but not via simple "xec" from the terminal

lagergren:
> The xvm build seems to be working fine. Slight challenge for developing using separate stand-alone xtc repositories (platform or examples) is that there are two ways to publish things:
> 
> 1. Using "installLocalDist", which updates the local brew cellar, making the standalone "xcc" and "xec" commands using the newly minted xdk
> 2. Using "publishLocal", which update the local maven repository (~/.m2), which allows the "build" command to use the newly minted compiler, and run it via "xtcRun" task, but not via simple "xec" from the terminal

There is nothing "simple" with that "Xec from the terminal". I should really be a Maven Application Start Script generated from the installDist task, but I haven't had time for that. You are supposed to upgrade a version with the outputs from the installDist task, so it seems to me that you could just extend installLocal dist to take take the contents of installDist and put it somewhere in your home directory, which is also on the path. This a good property candidate.  Add some code to set up the symlinks to the native launchers, and have installDist write its full distort in a place in your home directory where you keep evolving distros

I haven't a single cycle to spend on the examples repo, or even better, I encourage that you do it to learn about the system. I intended to port them before merging the platform. This still doesn't get in the way of merging XVM though.

ggleyzer:
I'm not sure about the Maven Start Script and why is it better than brew, but you're right, it's a different topic. Just tested that the new xvm build using "installLocalDist" target works just fine with the old platform build, so we're good to go

cpurdy (approved):
Still lots of TODOs but Gene and I both have it running and it seems stable enough to move to the main line and polish it up there. It appears that the branch is slightly behind and needs to be brought up to date in order to merge it in.

lagergren:
> I'm not sure about the Maven Start Script and why is it better than brew, but you're right, it's a different topic. Just tested that the new xvm build using "installLocalDist" target works just fine with the old platform build, so we're good to go

The start script hasn't got anything to do with brew. The start script is better than building your own binary launcher, that does exactly what the start script does.

PR #116, opened 2024-02-02
Added some new github runner actions to try to trap any cache problem…

lagergren:
Completed porting native launcher code to Java, placing no-argument launchers (as the application plugin wants for its startup scripts) for the compiler and runner in Javatools.

Now this should be simple to hook up to script generation instead of binary/symlink launchers.


javatools/src/main/java/org/xvm/tool/XdkResolvingLauncher.java
```diff
@@ -0,0 +1,125 @@
...
+ * solution.
+ * <p>
+ * The logic implemented in this file is identical to the logic used by our
+ * legacy binary launchers for identifying and resolving if we are executing
+ * from and XTC installation, and the locations of its core libraries required
+ * to run are present.
+ * <p>
+ * The application plugin launch config will look something like this:
+ *   + applicationDefaultJvmArgs = "-ea" (as in the current native launchers)
+ *   + applicationDistribution = the xdk/build/install directory, or rather, the outputs of installDist
```
aalmiray, line 29: Do we really have to se this? Shouldn't this be part of the convention?

lagergren: It is totally part of the convention, at the moment the CR is rather much "work in progress and note to self".  Rest assured t will be cleaned up.  Btw - since the application plugin only assumes one application config per project, I'm probably just going to add two createStartScript type tasks to xdk/build.gradle.kts and have the distribution generation depend on those. The distribution can copy them to the bin dir . Right now the distribution only copies the three platform specific exe launchers to libexec/bin, and doesn't put anything in the upper bin directory for the distribution. This is where I want to place the start scripts, which should be trivial to extend the distribution configuration to do.    That, if I understand it correctly, supersedes both the need for symlinking anything anywhere, as well as requiring an installed home-brew distribution with the launcher links in the right place, where we overwrite arbitrary parts of that tree to "refresh" a local install .

lagergren: Working as fast as I can, but I got interrupted by debugability questions for a bit so I am a bit behind schedule, but not much left now hopefully.

lagergren:
This is draft, but I'm closing this to approach it from another angle later, or even better, let someone else do it, so that the build system knowledge has less of a bus factor. From 
@cpurdy's release pipeline work, it's a lot more obvious to me what we should keep and not.

PR #121, opened 2024-02-07
Gradle-ized the xtcrun and xtccompile tasks, so that we can extend configurations in any tasks of our choice. Fixed user feedback on DSL and "bootstrap quickly" stuff feedback from JK and Dima

lagergren:


manualTests/build.gradle.kts
```diff
@@ -173,29 +253,105 @@ xtcRun {
...
      */
+    val moduleTestName = getTestModuleName() // module name on command line or 'TestSimple' as default
+    // Special case - use the Runner to run all tests in parallel, as long as the xtcRunAll tasks does not
+    // automatically do this with everything in the module path of the source set. Also modify that task
+    // after bugfixing to have a parallel option, through the XTC runner, or whatever it may be.
     module {
-        moduleName = "TestFizzBuzz" // Will add other ways to resolve modules too.
-        showVersion = true // Overrides env showVersion flag.
-        moduleArgs("Hello, ", "World!")
+        if (!shouldRunAllTestsInParallel()) {
```
ggleyzer, line 271: I'd suggest to avoid a double-negative:

if (shouldRunAllTestsInParallel()) {
   ...
} else {
    moduleName = moduleTestName
}

settings.gradle.kts
```diff
@@ -38,4 +38,6 @@ includeBuild("plugin")
 includeBuild("xdk")
 includeBuild("manualTests")
```
ggleyzer, line 39: I would argue that your original assessment was correct and we don't need "manualTest" to be a part of the main build. If one needs such a "sanity check", he could trivially say:
    ./gradlew :manualTests:build

ggleyzer (changes requested):

settings.gradle.kts
lagergren: I'll look for a completely excluded way to lift manual tests out of the build, but still can be invoked somehow from the XVM repo build system. The point of pointing something to a build system is that the build system can (optionally) compile and run it. I'll see what i can do with a more complete detachment.

lagergren:
Flagging as draft. Now I have the run all task working again, with will run all the source set outputs of a source set in deterministic sequential order. The command line override -PmethodName=... is integrated into the build as a project property, and I will soon check in the "gradle del" way of doing the parallel runner, with arguments provided from the source set output, but I may have to do that tomorrow, as it took more time to clean this up than expected. However - the code got significantly healthier and there were several broken things. 

There are some printouts, and as I said, the parallel case missing but it's working , just needs to be decluttered and it's not long. This will show the the build del can be extended instead of a lot of weird functionality being added through if else in the scripts.

Also made source set handling more generic.

There are some code warnings, forgotten imports, printouts that should go away. Will get rid of those too, but hopefully this is good to go tomorrow and then I will deal with the other build issues, like the incremental issue with the platform ,and the xunit source set configuration with JK.

lagergren:
Other test examples that we could use after seeing people trip over things in manualTest:

Inherit the xtc-template-app into a hello world that uses the xdk as an included build instead of the repository. Should be simple as a starting point for this instead of poking at the manual build .

settings.gradle.kts
lagergren: Found a way to do it.

lagergren:
For unknown reason there was a merge created at 10:40 this morning, messing up any chance to keep this change history separate from continuously updated master CRs.

lagergren:
Moved to new PR.

PR #124, opened 2024-02-12
Update the packed integer (XIP) format

cpurdy:
Simplifies the XIP format to make packed integer numbers easier to read in a debugger. Expand support for larger packed integer values, primarily for crypto key usage. Update Unicode to latest 15.1.0 (Sept '23) version.


ggleyzer (approved):
It's very strange that some of the previous changes showed up in the file set...

PR #125, opened 2024-02-13
Fixed broken build. ManualTests had had the TestFizzBuzz.x deleted, and replaced with FizzBuzz.x, that defines a module with a different name. It also does not contain the parameter passing tests that the original code has. Reverting.

lagergren:
Fixed broken build. ManualTests had had the TestFizzBuzz.x deleted, and replaced with FizzBuzz.x, that defines a module with a different name. It also does not contain the parameter passing tests that the original code has. Reverting.

### DIFF
--- a/manualTests/src/main/x/FizzBuzz.x
+++ b/manualTests/src/main/x/FizzBuzz.x
@@ -1,6 +1,15 @@
+// TODO: Add unit tests for both broken and working code/runtime modules.
+// TODO: Add a "test debugger" task to manualTests with the assert:debug enabled, or something.
 module FizzBuzz {
-    void run() {
+    void run(String[] args = []) {
         @Inject Console console;
+
+        loop: for (String arg : args) {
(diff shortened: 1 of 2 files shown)


manualTests/src/main/x/FizzBuzz.x
```diff
@@ -1,6 +1,15 @@
+// TODO: Add unit tests for both broken and working code/runtime modules.
+// TODO: Add a "test debugger" task to manualTests with the assert:debug enabled, or something.
 module FizzBuzz {
-    void run() {
+    void run(String[] args = []) {
         @Inject Console console;
+
+        loop: for (String arg : args) {
+            console.print($"TestFizzBuzzArgument: (args[{loop.count}] = {arg})");
```
thegridman, line 8: Should this be `"FizzBuzzArgument:`

thegridman (approved):

PR #126, opened 2024-02-13
Make an echo test specifically for gradle build use

cpurdy:
Create a new test (in the manual tests sub-project) that can be used by the Gradle build in lieu of the FizzBuzz test.

### DIFF
--- a/manualTests/src/main/x/EchoTest.x
+++ b/manualTests/src/main/x/EchoTest.x
@@ -1,0 +1,15 @@
+/**
+ * This test is invoked by the build. It can also be invoked manually using the following command:
+ *
+ *     ./gradlew :manualTests:runXtc
+ */
+module EchoTest {
+    void run(String[] args = []) {
+        @Inject Console console;
+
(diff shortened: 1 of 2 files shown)


lagergren:
Did you also want to remove the argument parsing from FizzBuzz.x? Nevermind. This is approved. The question is why we are spending time here right now until I have managed to finish my PR?

manualTests/src/main/x/EchoTest.x
```diff
@@ -0,0 +1,15 @@
+/**
+ * This test is invoked by the build. It can also be invoked manually using the following command:
+ *
+ *     ./gradlew :manualTests:runXtc
+ */
+module EchoTest {
+    void run(String[] args = []) {
+        @Inject Console console;
```
lagergren, line 8: Nitpick. This comment will get stale fast, and is still kind of weird. _Every_ test can be invoked by the build, and with runXtc with different configurations and properties. If you have this comment here, you might as well copy/paste it to every single test in the ManualTests project src source.

lagergren (approved):

Issue #130, opened 2024-02-14
Launchers should respect XDK_HOME


cpurdy:
**Is your feature request related to a problem? Please describe.**
`xec` and `xcc` (etc.) commands should use the XDK specified by `XDK_HOME`, iff that env variable exists.

**Describe the solution you'd like**
Launchers should check for the XDK_HOME env var (e.g. using `getenv()`).

**Describe alternatives you've considered**

* set path explicitly
* specify full path of the `xec` (etc.) commands on the command line
* copy the built libs into the location that already holds the XDK that's in the path (e.g. the way that we used to overwrite the brew install from the build 🤮)

**Additional context**
N/A

lagergren:
The only thing that may be missing here (I don't know) is a way to make the launcher NOT respect XDK_HOME. You aren't always in a place where you can rewrite system variables, but it's no biggie. I just want to make it very easy for the user to understand what XDK is the "current" one.

thegridman:
A number of other CLI type tools I use will pick up their home/config directory location from an environment variable (if set) but also have a corresponding command line flag to set the location too, which would be used in preference to any environment variable.

cpurdy:
We can add the other suggested options later without a great deal of effort, if they are required.

Issue #129, opened 2024-02-14
Remove "libexec" from the build output of the xdk


cpurdy:
**Describe the bug**
The new build includes a subdirectory under the created "xdk" directory called "libexec".

**To Reproduce**
`./gradlew install`
Then look under `./xdk/build/xdk/` directory

**Expected behavior**
No `./xdk/build/xdk/libexec` directory should be created; the current contents of that directory should be moved on level up.

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

cpurdy:
The result is in mainline and everything seems to work. This fix was necessary for the XDK_HOME changes as well, and it is also working already and in mainline.

PR #133, opened 2024-02-15
Cpurdy/#130

cpurdy:
#130 -> Marcus : this will need to be tested on macOS. It depends on the #129 changes. Wherever your current xec is in your path (assuming its not under your XDK_HOME itself), you should copy the new macos_launcher binary to that location. This change will now use the XDK_HOME location (if one is spec'd) intead of location where the launcher is in the path.


manualTests/build.gradle.kts
```diff
@@ -187,9 +188,9 @@ xtcRun {
      * TODO: Add a nicer DSL syntax with a nested modules section.
      */
```
lagergren, line 189: You are doing something with the launcher in this branch, while I struggle to finish the build system. Can you please only modify the launcher code in this branch, and not do unrelated changes. I'm review ready in 30 minutes, and the manual tests do not look like that anymore.

lagergren: If you feel it is an important change that TestFizzBuzz changes names to EchoTest, please still do it in a separate small PR, so that it does not widen the impact area of conflict resolution.

ggleyzer (approved):

PR #141, opened 2024-02-18
Added debug launcher task property + launcher dsl property that will suspend any forked JavaExec task and wait for the debugger to attach. Also made it possible to enable with System environment variable XTC_DEBUG=true and XTC_DEBUG_PORT=<specific port> (default is 4711)

lagergren:
Added debug launcher task property + launcher dsl property that will suspend any forked JavaExec task and wait for the debugger to attach. Also made it possible to enable with System environment variable XTC_DEBUG=true and XTC_DEBUG_PORT=<specific port> (default is 4711)


VERSION
```diff
@@ -1 +1 @@
-0.4.4
+0.4.41
```
ggleyzer, line 1: Did you mean this version increase or this is just a remnant of your previous testing?

ggleyzer (approved):

VERSION
lagergren: I did mean to increase it, because I forgot when I changed public task methods in the ManualTest DSL resolution stuff.

Issue #140, opened 2024-02-16
Rename UnsupportedOperation to Unsupported


cpurdy:
**Describe the bug**
Exception names describe what went wrong, attempting to use nouns and avoiding ending the name with "Exception". There are of course exceptions to this rule, but we try to reduce them when possible.

In this case, we have the following usage:

```
throw new UnsupportedOperation();
```

While this is definitely better than `UnsupportedOperationException`, the problem that we're trying to describe is that something is _unsupported_, and "operation" doesn't add any quality information.

**To Reproduce**
n/a

**Expected behavior**
The result of this change will be code like:

```
throw new Unsupported();
```

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

lagergren:
That's while you typically send a string as an argument, or getClass().getSimpleName() or something like that?

lagergren:
To check alternatives in other ecosystems, the Kotlin TODO("...") is one example. Or its assert(condition) { lambda }

cpurdy:
I was unaware that Kotlin had a TODO. Just looked it up: It's a function that throws.

Ecstasy has had a `TODO` keyword (it's an expression, but it's normally used as a statement) since day one. e.g.

```
@Override
SimpleRequest with(Uri?        uri      = Null,
                   Protocol?   protocol = Null,
                   Path?       path     = Null,
                   AcceptList? accepts  = Null,
                   ) {
    TODO CP
}
```

... and ...

```
...
} else {
    TODO("delineation of non-delineated data");
}
```

... and ...

```
if (!(defaultHasher := algorithms.hasherFor("MD5"))) {
    defaultHasher = TODO create crypto.hashers.MD5 class
}
```

cpurdy:
Ecstasy also has an assert flavor of `TODO`:

```
assert:TODO format <= Huge as "Someone needs to implement the Huge format";
```

(And of course you can just throw the exception yourself.)

cpurdy:
> That's while you typically send a string as an argument, or getClass().getSimpleName() or something like that?

Two issues here:

1) the name is too long for no good reason (other than influence from e.g. Java)
2) yes! we definitely would pass a descriptive string if appropriate

lagergren:
[...]

PR #144, opened 2024-02-19
Rename UnsupportedOperation (#140)

cpurdy:
Rename UnsupportedOperation to Unsupported (and use its sub-class NotImplemented for TODOs)

### DIFF
--- a/lib_ecstasy/src/main/x/ecstasy.x
+++ b/lib_ecstasy/src/main/x/ecstasy.x
@@ -119,11 +119,18 @@ module ecstasy.xtclang.org {
             extends IllegalState(text, cause);
 
     /**
-     * An `UnsupportedOperation` exception is raised when an attempt is made to invoke functionality
-     * that is not present or has not yet been implemented.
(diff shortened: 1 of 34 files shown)


lib_ecstasy/src/main/x/ecstasy.x
```diff
@@ -119,12 +119,19 @@ module ecstasy.xtclang.org {
...
-     * that is not present or has not yet been implemented.
+     * An `Unsupported` exception is raised when an attempt is made to invoke functionality that is
+     * not present, has not yet been implemented, or is otherwise not available.
      */
-    const UnsupportedOperation(String? text = Null, Exception? cause = Null)
+    const Unsupported(String? text = Null, Exception? cause = Null)
             extends Exception(text, cause);
 
+    /**
+     * A `NotImplemented` exception is raised when a `TODO` is encountered or an `assert:TODO`
```
ggleyzer, line 129: Maybe the doc should use "T0D0" (zeros instead of Os) to avoid the unnecessary TODO highlight, since this is not a TODO

ggleyzer (approved):

PR #152, opened 2024-02-21
Change the log level to "info" for expected warnings

ggleyzer:


cpurdy (approved):
Reviewed with Gene. Build should have no extraneous messages under normal successful execution. In the future, use `--info` to see these messages.

PR #156, opened 2024-02-23
Implementing checkstyle for Sun/Oracle-like Java code standard, auto rewrite source code format, and a build flow. Enabled initially on the plugin and javatools_unicode packages.

lagergren:


gradle/checkstyle/xtc-idea-inspections.xml
```diff
@@ -0,0 +1,68 @@
...
+    <option name="m_ignoreNonPublicClasses" value="false" />
+  </inspection_tool>
+  <inspection_tool class="NonThreadSafeLazyInitialization" enabled="true" level="WARNING" enabled_by_default="true" />
+  <inspection_tool class="OverridableMethodCallDuringObjectConstruction" enabled="true" level="WARNING" enabled_by_default="true" />
+  <inspection_tool class="ReadResolveAndWriteReplaceProtected" enabled="true" level="WARNING" enabled_by_default="true" />
+  <inspection_tool class="RedundantFieldInitialization" enabled="true" level="WARNING" enabled_by_default="true" />
+  <inspection_tool class="SamePackageImport" enabled="true" level="WARNING" enabled_by_default="true" />
+  <inspection_tool class="SerializableInnerClassHasSerialVersionUIDField" enabled="true" level="WARNING" enabled_by_default="true">
+    <option name="ignoreAnonymousInnerClasses" value="false" />
+    <option name="superClassString" value="java.awt.Component" />
```
ggleyzer, line 50: I don't pretend to understand any of this, but why is "java.awt" is mentioned in the "xtc-" specific file?

gradle/checkstyle/xtc-idea-style.xml
ggleyzer, line 1: I thought it was a style for Ecstasy sources (".x" files). It appears, it's this a description of the "java code for Ecstasy internals" style. If so, can we please name them accordingly? Maybe "xvm" would be a more appropriate prefix, since the xvm project is written in Java

lagergren: Well it is for all sources supporters by the intellij settings, not just Java sources. I will certainly add an XTC code standard template as soon as I have finished my first beta of the language support in intellij plugin (currently does lexing and some syntax highlighting). As soon as this is up and not breaking, meaning IDEs have the concept of xtc as a first class language like Java or Kotlin, we will extend this mechanism in some way we are mutually happy with for xtc code. Until then, I'm on @cpurdy's mission to provide a pluggable single source of truth for Java conventions, which is part of the "looks familiar" adoption goal. Xtc support will need to be implemented on top of this, using the same framework. Here, at least I could pick something that required no extra weird additional settings from our side, and that Produces Java that can be ingested quickly by the majority of open source Developers who are used to the two main conventions for the language. With the plugins described, especially IDEA checkstyle, I find it very easy to correct unchecked code that is slightly different from the SUN/Oracle standard in realtime, and that is really valuble. With no extra config, the checkstyle IDEA plugin will make it easy to correct code that is slightly off from the standard when I write it. 

Anyway, this PR is invisble to its potential remaining users, javatools and javatools_utils. We can take our time porting those over, tweaking any automation that may not be in the ide configs, and also change parts of the code standard definition if there is a really good reason. 

Anyway, now that we have this, extending it to xtc code as soon as I have a rudimentary ide plugin working for language support, is easy and within well defined parameters.

lagergren: - What I'm saying is the exported XML has more config than Java. Already it's not Java specific. It's for all languages. For example, you can see that it also supports Kotlin/buildscripts. Luckily, in the Kotlin World, 99% of the audience just choose "this one button Click that ensures official Kotlin standard for kt and kts files. So that is there already. When I teach intellij that those .x files contain another language, it Will fit right in.

gradle/checkstyle/xtc-idea-inspections.xml
lagergren: I take no responsibility for what JetBrains spews out as "the stuff you are doing now, could you export it". 

The single source of truth for code format is checkstyle.xml and checkstyle.xml only. If we start converting the remaining Java to code standard, we will automatically find out what needs to be tweaked for intellij to reformat to unambigious Java code following the standard. I'm sure there are many things. But the checkstyle config is already the source of truth and without any tools, the build will pass when it is detected conformant. I'n sure we will incrementally work on updating those idea specific files together, to ensure that we can automatically comform to the standard, and even add hooks so that you can change a file, having it jump to whatever code standard you prefer, and then on save reformations to the orgwide one, just as I decribed my workflow to Cliff in the chat today.

gradle/checkstyle/xtc-idea-style.xml
lagergren: You will notice that the editorconfig, even though not fully supported by intellij is more than Java too. Hence, I argue that this is a correct start with correct config files, that can be kept the way they are, as they are not language specific (even though Java has the most rules). As I Said, we already have working Kotlin checkstyle baked into the same files (as standard Kotlin code style). There are also various activations of the standardizer for HTML, cvs, vue, Shell scripts etc etc.

lagergren: So to clarify - this is _not_ an XTC specific file when it comes to language. It's an XTC specific file because that is the code standard the XVM and other repositories we use will utilise to provide code formatting for all languages involved.

I started out calling it xvm-idea-inspections.xml, but since it's not just a config we want to use for the XVM project, I changed its name to XTC, to imply that it's for the entire XTC platform, not specifically for XTC language. 

I have no idea what this files contains either, as it's machine generated, and no one should ever poke into it. For me, it's enough to know it's generated from the code standard settings, and can be imported into IntelliJ to enforce it. We will probably incrementally add missing things, more checks, better code inspections for all our implementation languages during the dev process. This is done by changing some setting in the IntelliJ code style, and re-exporting the style xml. 

It doesn't really matter what the file is called. I'll be happy to just get this merged in its existing state, so that it's there, and we can then start playing with it together, including deciding that filenames or something aren't ideal, but it's not important at the moment.

PR #157, opened 2024-02-23
Change the log level to "info" for expected "installDist" warning

ggleyzer:
Same as [this](https://github.com/xtclang/xvm/pull/152). Build should have no extraneous messages under normal successful execution. In the future, use --info to see these messages.

### DIFF
--- a/build-logic/common-plugins/src/main/kotlin/XdkDistribution.kt
+++ b/build-logic/common-plugins/src/main/kotlin/XdkDistribution.kt
@@ -69,7 +69,7 @@ class XdkDistribution(project: Project): XdkProjectBuildLogic(project) {
             }
             return true
         }
-        logger.warn("$prefix 'distExe' is disabled for building distributions. Only 'tar.gz' and 'zip' are allowed.")
+        logger.info("$prefix 'distExe' is disabled for building distributions. Only 'tar.gz' and 'zip' are allowed.")
         return false
     }
 


lagergren:
I disagree. This is something documented to happen, or rather implemented and showing up as something that should happen, that doesn't happen. Hence you are given a warning. Either remove the distExe from the lifecycle or keep the warning. We should decide if we want the distExe task, which currently only runs on @cpurdy's unreproducible install build machine, and in the xdk-release container on all systems. 

It would be far better to treat warnings as reminders to actually solve the real problems behind them, so that they can be removed as no longer relevant, than hiding them. Here, the problem is what should go into our release distribution, and even more important, how we should build it. If we need a Windows installer, this is a very brittle way to create it.

ggleyzer:
I agree with that alternative. Let's remove the dependency on "distExe" task.

ggleyzer:
What is your suggestion? What would you like me to do in this regard?

ggleyzer:
Consider this: we have tons of "not yet implemented" or "too naively implemented" features in the compiler and run-time. But we don't issue all those warnings every time we compile something. Instead, we leave comments in the relevant code, create "To Do" lists, open issues, etc.
All I'm suggesting, that we act in the similar way with the build logic.

lagergren:
> What is your suggestion? What would you like me to do in this regard?

I would like someone to give me any feedback on what I have said about releases, and perhaps write some code to figure out in which direction we should move to automate them and publish them. 

To automate release publication the fastest way to do it would be to not include a Windows installer at all. So A simpler subproblem is that we should decide if we even want to keep maintaining a Windows binary exe installer at all, which is a large part of the xdk/build.gradle.kts, and a Frankenstein container at xdk-release. 

I've asked for feedback on both, and for someone to play around with them and help me figure out if they have any purpose. I have done some false starts here, and it would be great if someone else could sketch out this model better + the artifact publication, because I am far from an expert here. It would be extremely helpful if someone else also spent some cycles to e.g. hack out the beginning of a simple release flow, so we can figure out where the pieces fit. I think it's mostly a matter of implementing point 1 and 2.1-2.2 above, but it seems to need a floor to stand on first.

lagergren:
We still have the problem of three native binaries, and the three flavours not fitting to well with standard releases scenarios. @aalmiray gave us some feedback here earlier, but if we could have a single release zip/tar.gz that would run on all platforms, that would be ideal. If not, we have to deal with the platform matrix in some maintainable and automatable way.

ggleyzer:
I agree with all that, but I think my argument still stands. Developers don't need to constantly see warnings of things that are not yet implemented, there are different means for that

lagergren:
I suggest that PR does something to solve the release model questions instead of removing the warning. 

Either we get rid of distExe altogether, or we get rid of it, and add build logic for creating platform dependent zips, where the bin directory with the appropriate launcher gets included as part of the archive. Publish these. 

If I'm not the one who has to do this, it would be great, because most of my resources are currently spent answering questions about the build system, and for some reason it doesn't seems to work at all as knowledge transfer. I think the only way to minimise the bus factor if get run over and freeing my up enough to get deep into "proper" language support is if the rest of the team start working on some smaller missing features and bug fixes (e.g. the remaining dependency issues) in the build.

PR #161, opened 2024-02-25
Improve String.split() API

ggleyzer:
I think it makes sense to trim during the split

### DIFF
--- a/manualTests/src/main/x/TestSimple.x
+++ b/manualTests/src/main/x/TestSimple.x
@@ -3,7 +3,7 @@ module TestSimple {
 
     void run() {
         String s = "hello    there";
-        console.print(s.split(' ', False));  // old behavior
-        console.print(s.split(' '));         // new behavior
+        console.print(s.split(' '));       // old behavior
+        console.print(s.split(' ', True)); // new behavior
     }
 }
--- a/manualTests/src/main/x/TestSimple.x
+++ b/manualTests/src/main/x/TestSimple.x
@@ -2,11 +2,8 @@ module TestSimple {
     @Inject Console console;
 
     void run() {
-        String[] results = new String[];
-        String   next    = "012345";
-
-        results += next[1 ..< 4]; // this used to fail to compile
-        console.print(results);
-
+        String s = "hello    there";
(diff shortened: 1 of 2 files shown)


manualTests/src/main/x/TestSimple.x
```diff
@@ -2,11 +2,8 @@ module TestSimple {
...
     void run() {
-        String[] results = new String[];
-        String   next    = "012345";
-
-        results += next[1 ..< 4]; // this used to fail to compile
-        console.print(results);
-
+        String s = "hello    there";
+        console.print(s.split(' ', False));  // old behavior
+        console.print(s.split(' '));         // new behavior
```
lagergren, line 7: You could turn this useful test into a TCK test or a permanent unit test. @gridman and I need some test corpus for xunit raw input as well, and every little bit helps.

lagergren (approved):

manualTests/src/main/x/TestSimple.x
ggleyzer: This is not really a test (no invariant is checked), but rather a demonstration. But yes, I'll work on turning every applicable TestSimple submission into a real test when we have a framework

cpurdy (commented):
Gene - default for "trim" option should be False

manualTests/src/main/x/TestSimple.x
lagergren: Well so are most of the tests in manualTests, and just att the demonstrations as soon as you have them, because we want as much corpus code as possible. Even though JK hasn't figured out exactly what the tests will look like, it should be a trivial operation, so let's every little xtc snippet that is used in debugging to the manual tests repo asap imho.

ggleyzer: Definitely all the manual tests need to be converted as well. While most of the time it's indeed trivial to convert our "demos" to a proper invariant check, sometimes it's not. Take a look, for example, into services.x. I was thinking about it a bit and could not come up with any decent solution.

lagergren: Yeah, but in this case it's trivial, so slam it in there as TestSplit or something?

PR #162, opened 2024-02-26
Add better abstraction to distribution building code.

lagergren:
Added parameterizable distribution contents so we can modify it for several different Gradle distributions with different classifiers to get a more versatile release pipeline. Haven't actually done any yet, would want to discuss with @cpurdy first.

However, this should be possible to get in as a trivial abstraction raising change.

Also cleaned up several things related to the build, like dead code, unused configurations and so on.


lagergren:
This is rather trivial change. I moved to copyspec config for the distribution tasks to a common place so we can easily create new distros for new platforms with minimum hassle. I also took the Liberty to clean up some sucky Kotlin code I wrote in q2-2023 before I had fully understood Kotlin.

ggleyzer (approved):

Issue #155, opened 2024-02-23
Add a compiler warning for self-assigned properties


ggleyzer:
A quote from the post by JK:

> Today I spent a while trying to debug why one of my tests was failing.This was the issue, the code below is a constructor for one of my const classes:

>        construct (String[]? value = Null) {
>           this.values = values;
>        }
>       private String[]? values;

> As you can see, there is a typo so I am assigning the unassigned field values to itself. It should either be this.values = value; or change the parameter name to values.

As Markus pointed out, Java does have a "Variable is assigned to itself" warning. It would be nice to add a similar one to the Ecstasy compiler

ggleyzer:
Please review https://github.com/xtclang/xvm/pull/164

PR #164, opened 2024-02-26
Add a compiler check for self-assigned properties

ggleyzer:
A fix for issue https://github.com/xtclang/xvm/issues/155.

It's relatively simple. The only question in my mind: should it be a warning or an error?

### DIFF
--- a/javatools/src/main/java/org/xvm/compiler/Compiler.java
+++ b/javatools/src/main/java/org/xvm/compiler/Compiler.java
@@ -1413,6 +1413,10 @@ public class Compiler
      */
     public static final String MISSING_PARAM_RESOURCE             = "COMPILER-200";
     /**
+     * Property {0} is assigned to itself.
+     */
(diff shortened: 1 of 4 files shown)


lagergren (approved):
I think it should be an error. Finality issues and statements that probably haven't got a side effect but actually do are likely never correct. I think it's fair to break the build on this, because otherwise someone is going to disable all warnings if we allow that and just run. But the important thing to me is that this is at least detected and flaggen now. Let's merge it like this, and please merge a test program that violates it, and @gridman and I can give it official tck status when that works.

ggleyzer:
A "test program that violates this" is not a simple proposition. It would have to be a negative compilation test, which - while doable as a hack now - I'd prefer to be a part of a generic compiler testing framework

ggleyzer:
I agree with Markus. Besides, if at some point we realize this appears to be an overkill, it's easier to relax than to tighten the compiler checks.

Issue #163, opened 2024-02-26
Build output in .zip and .tar.gz does not match previous layout 


cpurdy:
**Describe the bug**

Current install layout (the child "xdk" directory should have been one level up with the other "xdk" directory):
```
xvm$ tree ./xdk/build/install/    
./xdk/build/install/
└── xdk
    ├── VERSION
    ├── bin
    │   ├── linux_launcher
    │   ├── macos_launcher
    │   └── windows_launcher.exe
    ├── javatools
    │   ├── javatools.jar
    │   ├── javatools_bridge.xtc
    │   └── javatools_turtle.xtc
    ├── lib
    │   ├── aggregate.xtc
    │   ├── collections.xtc
    │   ├── crypto.xtc
    │   ├── ecstasy.xtc
    │   ├── json.xtc
    │   ├── jsondb.xtc
    │   ├── net.xtc
    │   ├── oodb.xtc
    │   ├── web.xtc
    │   ├── webauth.xtc
    │   └── xenia.xtc
    └── xdk
        ├── README.md
        ├── bin
        │   ├── README.md
        │   ├── cfg_macos.sh
        │   └── cfg_windows.cmd
        ├── doc
        │   └── README.md
        ├── examples
        │   ├── FizzBuzz.x
        │   ├── HelloClient.x
        │   ├── HelloServer.x
        │   ├── HelloWorld.x
        │   ├── README.md
        │   └── https.p12
        ├── javatools
        │   └── README.md
        └── lib
            └── README.md
```

Current .tar.gz layout (needs same fix as above, plus no xdk-0.4.43 up front):
```
xvm$ tar -tf xdk/build/distributions/xdk-0.4.43.tar.gz 
xdk-0.4.43/
xdk-0.4.43/VERSION
xdk-0.4.43/xdk/
xdk-0.4.43/xdk/bin/
xdk-0.4.43/xdk/bin/README.md
xdk-0.4.43/xdk/bin/cfg_macos.sh
xdk-0.4.43/xdk/bin/cfg_windows.cmd
xdk-0.4.43/xdk/README.md
xdk-0.4.43/xdk/examples/
xdk-0.4.43/xdk/examples/HelloClient.x
xdk-0.4.43/xdk/examples/FizzBuzz.x
xdk-0.4.43/xdk/examples/README.md
xdk-0.4.43/xdk/examples/HelloServer.x
xdk-0.4.43/xdk/examples/https.p12
[...]

PR #165, opened 2024-02-27
Correct the installDist output as per the original design (#163)

cpurdy:


cpurdy (commented):
Sorry to review my own work, but this is blocking the build automation.

PR #166, opened 2024-02-27
FIx the distExe task

lagergren:
Debugged the distExe task, which was in a pathologic place of the build d life cycle, relying on an install expansion, which in turn relies on assembleDist. Now it should be on the same level as distZip and distTar. Also fixed distribution contents eager processResource output access that resolved during the evaluation phase instead of the run phase.

Please try it before you merge it, @cpurdy 

We now run fine in any environment with NSis, including the xdk-release container.


ggleyzer:
Let's try to keep the commit messages [short and descriptive](https://gist.github.com/tonibardina/9290fbc7d605b4f86919426e614fe692)

ggleyzer (approved):

PR #168, opened 2024-02-28
The debug flag in the build DSL, and its environment variable shortcuts, now supersede the Gradle debug flagging.

lagergren:
Either run from a task DSL inheriting "debug = true", or set the XTC_DEBUG (and optionally XTC_DEBUG_PORT) environment variables to get every launcher task to suspend and wait for the debugger to attach.  

TODO: I will come up with some nice command line variant to do this as well.  Just need to figure out what is easy to use and saves time.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,15 +9,6 @@ org.gradle.caching.debug=false
 # TODO: Experiment with enabling the configuration cache later
 org.gradle.configuration-cache=false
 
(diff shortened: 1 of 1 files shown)


ggleyzer (approved):

lagergren:
This, somewhat sloppily, got in as part of a different PR, which of course it shouldn't, since we want to be good Git citizens and lead by example with short, ephemeral and feature specific branches :)

PR #170, opened 2024-02-28
Tweaked SepratorWrap checkstyle rule

lagergren:
The Oracle/Sun codestyle represented by checkstyle.xml used the SeparatorWrap module, with default options, which means that for a chain of calls, e.g. StringBuilder append, separated with dots, the dot goes at the end of the line. It is more common to keep it at the start of the line, so modified that rule a little bit to allow us to have chained operators first. This is already enforced for && and || logic and similar constructrs, so it feels consistent. We can of course revisit later, and we need to get an autoformatter that works without eating all explicit line breaks, but that's a different problem. First order of business is to have a checkstyle.xml that represents a working code standard.

### DIFF
--- a/javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
+++ b/javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
@@ -212,10 +212,16 @@ public class BuildUnicodeTables {
         }
 
         final var sb = new StringBuilder();
-        sb.append(name).append(": [index] \"str\" (freq) \n--------------------");
+        sb.append(name)
+            .append(": [index] \"str\" (freq) \n--------------------");
         int index = 0;
         for (final var entry : map.entrySet()) {
-            sb.append("\n[").append(index).append("] \"").append(entry.getKey()).append("\" (").append(entry.getValue()).append("x)");
+            sb.append("\n[")
+                .append(index)
+                .append("] \"")
+                .append(entry.getKey())
+                .append("\" (")
+                .append(entry.getValue()).append("x)");
             entry.setValue(index++);
         }
 
@@ -317,7 +323,9 @@ public class BuildUnicodeTables {
         @Override
         public String toString() {
             final var sb = new StringBuilder();
-            sb.append("UCD description=").append(description).append(", repertoire=\n");
+            sb.append("UCD description=")
+                .append(description)
+                .append(", repertoire=\n");
 
             int c = 0;
             for (final var item : repertoire) {
--- a/javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
+++ b/javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
@@ -213,15 +213,15 @@ public class BuildUnicodeTables {
 
         final var sb = new StringBuilder();
         sb.append(name)
-            .append(": [index] \"str\" (freq) \n--------------------");
+                .append(": [index] \"str\" (freq) \n--------------------");
         int index = 0;
         for (final var entry : map.entrySet()) {
             sb.append("\n[")
-                .append(index)
-                .append("] \"")
-                .append(entry.getKey())
-                .append("\" (")
-                .append(entry.getValue()).append("x)");
+                    .append(index)
+                    .append("] \"")
+                    .append(entry.getKey())
+                    .append("\" (")
+                    .append(entry.getValue()).append("x)");
             entry.setValue(index++);
         }
 
@@ -324,8 +324,8 @@ public class BuildUnicodeTables {
         public String toString() {
             final var sb = new StringBuilder();
             sb.append("UCD description=")
-                .append(description)
-                .append(", repertoire=\n");
+                    .append(description)
+                    .append(", repertoire=\n");
 
             int c = 0;
             for (final var item : repertoire) {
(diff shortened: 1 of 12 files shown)


javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
```diff
@@ -212,10 +212,16 @@ void writeResult(final String name, final String[] array) throws IOException {
         }
 
         final var sb = new StringBuilder();
-        sb.append(name).append(": [index] \"str\" (freq) \n--------------------");
+        sb.append(name)
+            .append(": [index] \"str\" (freq) \n--------------------");
```
ggleyzer, line 216: This line seems to be misaligned. Visually, it looks much neater if all the ".append" sit right under each other

javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
```diff
@@ -212,10 +212,16 @@ void writeResult(final String name, final String[] array) throws IOException {
...
         int index = 0;
         for (final var entry : map.entrySet()) {
-            sb.append("\n[").append(index).append("] \"").append(entry.getKey()).append("\" (").append(entry.getValue()).append("x)");
+            sb.append("\n[")
+                .append(index)
```
ggleyzer, line 220: ditto

ggleyzer (changes requested):

javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
lagergren: It's not misaligned. The indentation and the continuation indentation are four (4) spaces. All code standards I have gone through, including the language and platform agnostic ".editorconfig" global flags for all languages maintain the concept of indentation with the special case continuation indentation. While it can be another indentation level (e.g. continuations are indented 2 indent steps == 8, for an indentation of 4, or one indent step == 4). We can never have a continuation indentation that is not a multiple of the indent depth for any line. I mostly see (indent space) + (indent space), or (indent space) + 2 * (indent space) (e.g Google/Palantir) for continuation indentations like this.

Maintaining a context sensitive indenter that would check the surrounding code and do "appropriate" alignment based on lex tokens and/or control flow "in the vicinity" is very likely a path to madness, and I'm not surprised that no one seems to have done that, as that is what it takes. 

Preserving while space between variable, equals sign and value in a sequence of declarations is fine, and another thing altogether. Checkstyle makes no demands for that (and I'm sure we can figure out how to get IntelliJ to just do what our checkstyle says, or at least find the button that doesn't gobble whitespace, but we DO pass the code style check no matter if we have that declaration list variable alignment whitespace or not).

Here, though, we you are talking about changing indentation continuation depth, which is not something we want to do. In Python, it will even be a syntax error. Even though we are not Python, the indentation levels are *never* different and *never* depend on context. It is considered first priority that control flow is indented correctly and associated with its nesting level, no matter where it is. Aligning the dot back would suddenly cause a two space indentation in a world that uses indentation steps of 4.

I am probably not alone in thinking it's harder to read, especially when I look at more lines than above at the same time. The picture my brain uses the indents to create, basically a control flow map, immediately red-flags that continuation indent for me and tells me "that one doesn't look quite right, does it?" if it's not placed at a multiple of the global indentation space (n * 4).

javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
lagergren: And again - this is continuation indentation. They MUST be larger than the indentation level of the beginning of the chain statement, and they MUST be a multiple of the global indentation level.

javatools_unicode/src/main/java/org/xvm/tool/BuildUnicodeTables.java
ggleyzer: As you can see, we already disagree on the desired look and what "looks right" for just one place in one file. Imagine what happens when we try to apply this to 300,000 lines of code that Cam and I wrote earlier.

I believe that things like that should be kept as an author of the source choose. If you look at the original submission by Cam, the code looked like this:

            sb.append("\n[")
              .append(index)
              .append("] \"")
              .append(entry.getKey())
              .append("\" (")
              .append(entry.getValue())
              .append("x)");


Therefore, I strongly believe it should remain to look like that

cpurdy: This type of thing doesn't occur _that_ often, but "sb" and "buf" (with the style of fluency seen above) are not uncommon. I do imagine that it's quite a complicated "rule" to explain (and/or codify). We generally use 8 space indent for trailing lines, so it would look like this:
```
    sb.append("\n[")
            .append(index)
            .append("] \"")
            .append(entry.getKey())
            .append("\" (")
            .append(entry.getValue())
            .append("x)");
```

The visual benefit of the former is fairly easy to defend, from my POV.

cliffclick: random semi-unrelated.  I'm tons and tons of short complex printing, similar to above, into a StringBuilder.
I got one with very short print names specifically for this kind of thing, plus some formatting variants:
`sb.p("\n[").p(index).p("] \"").p(entry.getKey()).p("\" (").print(entry.getValue()).p(x)");`
OR 
`sb.fmt("\n[%0] \"%1\" (%2x)",index,entry.getKey(),entry.getValue());`

lagergren:
Closing this. Will check in a minimally modified example to hopefully make the point that 90% of the stuff you are used to are not subject to any code standard changes. Especially regarding preserving explicit line breaks and almost all alignment spacing.

PR #173, opened 2024-02-29
Added feedback from yesterday's meeting, with all debug properties now being both ENV vars and first class DSL members.

lagergren:
You can configure debug, debugPort and debugSuspend (boolean, default to false, int default to 4711, and boolean default to true) in any launcher extension DSL, and override it in any launcher task.


ggleyzer (approved):
Looks good

PR #174, opened 2024-03-01
Add construction tests

ggleyzer:


lagergren (approved):

lagergren:
It seems to me that the tck would like to have a root build.gradle.kts file like the manualTests project. Again, here we have a lot of @Test annotation, but this time we know that we have the XDK build with everything in place, so this should be a pretty good project to test out XUnit to see if XTC can bootstrap run and test XTC. Right, @thegridman ?

This won't prevent the test from being run with a command line launcher from anywhere, but it makes it possible to e.g. easily run the TCK as an on_push test action before pull requests go into master, and it might be a good XUnit test bed.

ggleyzer:
Yes, that is the intent exactly

lagergren:
> Yes, that is the intent exactly

Perfect. Let me know if you or @thegridman want me to spend half an hour or so pair programming a skeleton for this, preserving all other semantics and ways to invoke the TCK suite.  Perhaps it may even be a good thing to do together on Wednesday, or at least record with commentary and screen share, so that we get some knowledge transfer going for the builds?

PR #177, opened 2024-03-04
Updated Windows launcher to respect XDK_HOME env var

cpurdy:
Rebuilt the WIndows native launcher with the updated functionality that uses XDK_HOME if it is set.


javatools_launcher/src/main/c/makefile
```diff
@@ -13,7 +13,7 @@ CMP_OPTS ?= -O0
 
 # extra source file for shared linux/unix code
 ifneq ($(OS_NAME),windows)
-    SHARED_IMPL = os_unux.c
+	SHARED_IMPL = os_unux.c
 endif
```
lagergren, line 18: Could you perhaps add a README.md that explains what you need to do to build the launchers, and where you do it? I assume on a native Windows machine.

Is it possible to cross compile the Windows EXE on Linux? I suspect that you could install "mingw" on a Linux box with "sudo apt-get install mingw-w64" and use the x86_64-w64-mingw32-g++ executable to build? 

Could you also please check out jpackage and see if we can use that to package the Windows release. It seems to support creating both .msi and .exe installers, but you have to run on a Native windows host (or on a Windows platform GitHub action, so it's still potentially very useful). If we can build a Windows installer for XDK that does what your existing NSI logic does, then we have filed down a huge bump, and it would make our release flows tremendously more easier and possible to automate.

lagergren (approved):

javatools_launcher/src/main/c/makefile
cpurdy: I wish I could 🤣

I use MinGW on Windows to build this using gcc, which is why (as we discussed a few days ago) this should REALLY BE CONTAINERIZED OR SOMETHING.

It's painful that such a simple application, built with plain C, and a simple `makefile`, is SO FREAKING HARD TO BUILD. But it's because Windows doesn't have its own open tool chain, and the support it gets from the OSS community is really weird and half-baked. I could convert this to Microsoft Enterprise Visual Edge C++ Enterprise Edition Studio Pro, or whatever it's called, and that might be the better (?) long term option for (temporary) supportability from Microsoft. Or I need to document the steps (mainly, the `windres` app) to be able to do this using the gcc toolchain (which I used). I spent hours last night trying to set up a new Windows machine to replicate the build for this -- and I failed to do so, and ultimately I had to use the older Windows machine that I originally built this on. (I had everything working, exception `windres`, which has 2 (!!) different versions installed as part of WSL, so this really shouldn't be so freaking difficult.)

cpurdy: The diff you highlighted in this sub-thread is a spaces to tab diff, because gnu `make` requires the recipe following a target to begin with a tab, so I changed all of the leading indents in the file to use tabs to avoid any other make weirdnesses. It becomes more clear if you look at the entirety of the file, with a viewer that shows tabs vs spaces (in IDEA choose "View -> Active Editor -> Show Whitespace").

PR #178, opened 2024-03-09
Cc xcodes explore

cliffclick:
First cut new backend.
Runs some selections of TCK tests and trivial demo programs.


cliffclick:
Reviewed

tck/.dir-locals.el
```diff
@@ -0,0 +1,10 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+(
+ (dired-mode . ((c-basic-offset . 4)))
+ (java-mode  . ((c-basic-offset . 4)))
+)
```
lagergren, line 6: Add .dir-locas.el files to .gitignore

lagergren:
This is a little bit behind master. Could you rebase so the checks can re-run? 

A suggestion is to squash as many of these 86 commits as possible to a single one, if there is a risk of merge conflicts. They will get in as one commit, when a squash and merge is done, but you should realise that the commit description will be 86 "*" bullet points containing the exact wordings of the 86 commits.

doc/examples/HelloWorld.x
```diff
@@ -1,6 +1,6 @@
 module HelloWorld {
+    @Inject Console console;
     void run() {
```
lagergren, line 3: Is this an intended change for HelloWorld.x? If not, I guess it's better to leave as many master files as possible untouched. I also think that it makes more sense to inject a single dependency as close the usage point as possible, but I'll leave that to the original XTC language designers to determine if that is preferred or not.

javatools_backend/.dir-locals.el
```diff
@@ -0,0 +1,6 @@
+;; List of lists
+(
```
lagergren, line 2: While now added to .gitignore, this will still cause the elisp file to be merged to master. Please delete it with git rm, commit, and put the original file back in the same location, and do a "git status" to ensure that this doesn't show up as under source control but is still in the file system.

Please do this for the other elisp files too.

javatools_backend/src/main/java/org/xvm/xec/ecstasy/UnsupportedOperation.java
```diff
@@ -0,0 +1,4 @@
+package org.xvm.xec.ecstasy;
```
lagergren, line 1: Why not use java.lang.UnsupportedOperationException ?

lib_ecstasy/src/main/x/ecstasy/collections/List.x
```diff
@@ -1468,4 +1468,4 @@ interface List<Element>
 
         return True, maxValue, maxCount;
     }
-}
```
lagergren, line 1471: Revert unnecessary whitespace change to master files to avoid creating resolutions (even trivial ones)

lib_ecstasy/src/main/x/ecstasy/text/Char.x
```diff
@@ -182,7 +186,11 @@ const Char(UInt32 codepoint)
     String add(String s) = new StringBuffer(1 + s.size).add(this).addAll(s).toString();
 
     @Op("-")
-    Char sub(UInt32 n) = new Char(codepoint - n);
+    Char sub(Int n) {
+        assert:bounds -(0x10FFFF-codepoint) <= n <= codepoint as
+                $"Subtracting {n} from character code-point {codepoint} would exceed the Unicode range";
```
lagergren, line 191: The XTC convention for the rest of the tree seems to prescribe space around binary operators.

manualTests/src/main/x/TestMisc.x
```diff
@@ -3,12 +3,12 @@ module TestMisc {
 
     void run() {
         console.print("hello world!");
-
+        
         testBools();
-        testChars();
+        //testChars();
+        testAsciiDigit();
```
lagergren, line 9: This is the standard master testmisc, and merging it will have the effect of whatever is in testChars() and the other connected out methods will not be run from master anymore. Is that what we want?

manualTests/src/main/x/TestMisc.x
```diff
@@ -89,70 +89,56 @@ module TestMisc {
...
-        Object o2 = Int:4;
-        console.print("o2=" + o2);
-
-        try {
-            console.print("i.as(String) should throw " + i.as(String));
-        } catch (Exception e) {
-            console.print("i.as(String) failed correctly: \"" + e.text + '"');
-        }
-    }
+    // Requires unicode character handling
```
lagergren, line 100: Since this test logic is part of master, should it be merged in commented out? This will reduce the test code coverage for existing backend.

tck/.dir-locals.el
```diff
@@ -0,0 +1,10 @@
...
+
+
```
lagergren, line 8: Please delete this file from source control (and replace it with a local copy after committing if you want) to prevent it from getting merged to master.

tck/src/main/x/tck/comparison/Compare.x
```diff
@@ -175,4 +169,29 @@ class Compare {
             return value1.key <=> value2.key;
         }
     }
+
+    @Test
+    void compareArraysFormal() {
+        //Int[] a1 = [1, 2];
+        //Int[] a2 = [2, 1];
```
lagergren, line 176: These methods have been moved from their original location, but their contents have also been commented out. Is this intentional, because it may reduce code coverage testing for existing backend?

xdk/build.gradle.kts
```diff
@@ -305,8 +305,8 @@ val installLocalDist by tasks.registering {
         // TODO: The launchers should just be application plugin scripts, this is kind of ridiculous.
         listOf("xcc", "xec", "xtc").forEach {
             val symLink = File(binDir, it)
-            logger.info("$prefix Copying launcher '$it' -> '${launcherExe.asFile}' (on Windows, this may require developer mode settings).")
-            Files.copy(launcherExe.asFile.toPath(), symLink.toPath())
+            logger.info("$prefix Creating symlink for launcher '$it' -> '${launcherExe.asFile}' (on Windows, this may require developer mode settings).")
```
lagergren, line 308: We are moving away from the symlinked installLocalDist and there are already other ways to create and run a local installation with a binary launcher. Revert this to master code to avoid unnecessary changes.

Makefile
```diff
@@ -0,0 +1,163 @@
...
+# Makefile expects to run at the project root, and not nested inside anywhere
+#
+#  cd $DESK/xvm; make
+
+# Use bash for recipes
+SHELL := /bin/bash
+
+# Keep partial builds but not partial recipes
+.NOTINTERMEDIATE:	
+
```
lagergren, line 12: This makefile may belong more in the javatools_backend directory than in the project root. We don't want to mistaken for a part of the global build system.

javatools_backend/src/main/java/org/xvm/notes.txt
```diff
@@ -0,0 +1,148 @@
...
+* org/xvm/XEC.java - entry point
+
+* org/xvm/xtc - Read an XTC file support classes; ClassPart, *Part, CPool,Version,VerTree
+* org/xvm/xtc/cons - "cons" things from XTC files
+* org/xvm/xtc/AST  - "AST"  things from XTC files
+
+* org/xvm/xtc/ClzBuilder, ModBuilder, Javac, XType, XValue,
+
+* org/xvm/xec - generated class strings act "as if" loaded from here
+* org/xvm/xec - runtime support classes (no ties with gen classes)
```
lagergren, line 14: Should this file be merged into master? Should it be turned into GitHub issues or something instead?

Makefile
```diff
@@ -0,0 +1,163 @@
...
+# Find a reasonable ctags.
+CTAGS = $(shell which ctags)
+# Hack for MacOS: /usr/bin/ctags is unfriendly, so look for ctags from brew
+ifeq ($(UNAME),Darwin)
+	CTAGS = $(shell brew list ctags 2> /dev/null | grep bin/ctags)
+endif
+
+# Fun Args to javac.
+JAVAC_ARGS = -source 21 -target 21 -XDignore.symbol.file -Xlint:all -Xlint:-this-escape -Xlint:-unchecked -Xlint:-preview -Xlint:-deprecation -Xlint:-serial -Xlint:-rawtypes
+
```
lagergren, line 41: Why are you even turning on lint, if you remove a huge chunk of the more common warnings? As much as people expect code standards not being too weird in 2024 when they take a look at open source projects, they also tend to react negatively to a mixture of Java 1.4 and Java 21 code. We should strive to use generics where generics are suggested by the Java language, we should not use raw types, we should avoid escaping "this" from constructors, and we should design stuff so that having to do unchecked casts are a rare exception. We should certainly warn about deprecations, because it's bad if something stops working without warning when e.g. Java is upgraded.

If we do not want to follow Java programming practices, and it's motivated to turn off a lint check, then that should be done by @SuppressWarnings at the location of the code smell, so that it can be identified and refactored later. For example, I have some issues with "this-escape" because javax Inject makes some stateless/final things hard, but I have flagged it as a suppressedwarning in the code so that both things like SonarQube and future greps can identify the sites that should be given some refactoring love in the future.

doc/examples/eightQueens.x
```diff
@@ -0,0 +1,133 @@
+/**
+ * A solver for the classic 8-queens problem.
+ *
+ * @see https://rosettacode.org/wiki/N-queens_problem
+ */
+module eightQueens {
+    void run() {
```
lagergren, line 7: *nitpick* I am not 100% sure, but given the general look of XTC code, there is probably a naming convention for XTC Modules similar to the one in Java? Pretty much all modules in our test and example code are named in CapitalCamelCase, and not lowerCamelCase. That means that this should be named EightQueens for consistency.

lagergren (changes requested):
Since you have git merged out to this branch, it's not longer guaranteed to be rebaseable as per our Git flow, and it is not to be merged until the merge out is either somehow resolved, or the changes can be cherry picked to a new branch. We do not merge things may cause disruption to the linear change history. 

(merge in question: https://github.com/xtclang/xvm/actions/runs/8256571722)

doc/examples/eightQueens.x
cliffclick: Code directly from @cpurdy , leaving it for him to rename.

cliffclick:
No merge was intentionally made, although tooling might have snuck one in on me.  The last rebase was a fluster cluck, with all sorts of madness conflicts (100 files conflicting with no visible differences, files conflicting with later commit changes, or just plain self-conflicting?).  I took my best effort to cleanup without losing any of my changes, and I borrowed heavily from SO to do so.

So, having this merge in place, how is it removed?  (and I don't even know how to tell a merge happened, and the link you posted did not mention any merge)

lagergren:
> No merge was intentionally made, although tooling might have snuck one in on me. The last rebase was a fluster cluck, with all sorts of madness conflicts (100 files conflicting with no visible differences, files conflicting with later commit changes, or just plain self-conflicting?). I took my best effort to cleanup without losing any of my changes, and I borrowed heavily from SO to do so.
> 
> So, having this merge in place, how is it removed? (and I don't even know how to tell a merge happened, and the link you posted did not mention any merge)

Which part of the large heading "Merge branch 'cc_xcodes_explore' of https://github.com/xtclang/xvm in… #372" indicates this is not a merge? 

So yes, I do not think that you have manually and deliberately added a commit message that starts with 'Merge branch' and just happens to conform to the exact template that GitHub uses for merge commits created by clicking the "update branch" button when something is flagged as out of date with master. 

Another way to tell would be to take a look at your git history, and with the "--graph" option to git log, that commit, the very same commit, totally indicates it's a merge. That is why it says "Merge:" and mentions two heads. 

*   commit b59ad769914631c5943aa27bef99e9dd16383827
|\  Merge: 388d0a22b c61be8386
| | Author: Cliff Click <cliffc@acm.org>
| | Date:   Tue Mar 12 15:29:48 2024 -0700
| | 
| |     Merge branch 'cc_xcodes_explore' of https://github.com/xtclang/xvm into cc_xcodes_explore
| | 

You can also see how the branch depth increases in the ASCII art thingy at the start of the lines. 

More alarmingly, this is not the only merge commit you have in this branch, and the presence of the older ones, for example the one from March 9, when you tried to do your rebase yesterday indicates that this is very likely EXACTLY why you have hade "100 files conflicting with no visible differences". If we merge this to master as is, everyone else would also be forced to live in worlds with "100 files conflicting with no visible differences."

As for your question "how is it removed", I don't know exactly how to unf*ck a git branch manually where there are historically committed and pushed merges, although I'm sure everything is probably possible with enough reflog manipulation, command line magic and things I have no idea how they work at more than a cargo cult level. In a rebase based GitFlow, it is significantly easier to manipulate history back to normal, which is why we don't use merge outs in ours.

As per my original comment, I would think that the best way out is to create a new branch and cherry pick your commits over.

lagergren:
Something like this seems to be what you are looking for, i.e. turning the merge commits into normal ones. Their presence also makes squashing over them impossible, which is another big problem in the current state. 

https://stackoverflow.com/questions/32840872/convert-merge-into-rebase-without-having-to-perform-the-merge-again

Pulling off a repair using something based on the methodology described in the StackOverflow link may require more work, and probably an understanding of how merge commits affect Git history as opposed to normal commits. I don't think you can copy and paste and change the depth from head to whatever works for your branch and that this magically just fixes it, but it indeed seems to be possible (maybe).

PR #179, opened 2024-03-11
Simplify generic Class<> usage

ggleyzer:


cpurdy (approved):
Looks good, if this is is the desired outcome. It does add an implicit rule, which needs to be doc'd (i.e. this isn't something we're just going to remember magically, and others won't understand it magically).

javatools/src/main/java/org/xvm/asm/ClassStructure.java
```diff
@@ -4255,6 +4255,28 @@ else if (cFormal > cActual)
...
+                            for (int i = cActual; i < cFormal; i++)
+                                {
+                                listActual.add(pool.ensureAccessTypeConstant(typePublic,
+                                    switch (i)
+                                        {
+                                        case 1  -> Access.PROTECTED;
+                                        case 2  -> Access.PRIVATE;
+                                        case 3  -> Access.STRUCT;
+                                        default -> throw new IllegalStateException();
+                                        }));
```
lagergren, line 4274: Any exception thrown just gets to the top as a launcher exception, and has no info in it. I strongly suggest that the IllegalStateException has a message object, or that we use our own CompilerException .

I also STRONGLY suggest that that exception is not a RuntimeException, because it is very hard to debug this code if you don't know it and runtime exceptions are thrown earlier in the flow. It further forces the user to declare throws clauses in methods, so that it can be seen this method propagates upwards and needs to be handled already. 

This has been one of my main issues when learning the fundamentals of the compiler and runner. If something is a runtime exception, it should really be asynchronous, random and not expected IMHO

javatools/src/main/java/org/xvm/asm/Component.java
```diff
@@ -3417,7 +3417,7 @@ protected void registerConstants(ConstantPool pool)
                     TypeConstant   type      = entry.getValue();
 
                     mapNew.put((StringConstant) pool.register(constName),
-                               (TypeConstant)   (type == null ? null : pool.register(type)));
+                               (TypeConstant) (type == null ? null : pool.register(type)));
                     }
                 m_mapParams = mapNew;
```
lagergren, line 3422: The assertion in this file (line 543) with the list assignment has side effects. Is this intentional? 

There are a few raw times in this changed code, and other linting issues. Perhaps we can start getting rid of those kinds of findings when we do bug fixes in their vicinity, for example unlinkSibling takes a raw map and so on.

javatools/src/main/java/org/xvm/asm/ClassStructure.java
ggleyzer: I would generally agree, but in this case this is basically an assert for a logically impossible scenario. The value of cFormal is known to be 4, the value of cActual is known to be greater than 0 and less than 4. The default case is just required by the switch expression.

javatools/src/main/java/org/xvm/asm/Component.java
ggleyzer: I don't know what assertion you're referring to. There is none on line 543 at Component.java

PR #184, opened 2024-03-13
Change the xtcCompile/compileXtc task default for the 'rebuild' flag and add an override

lagergren:
Changed the xtcCompile/compileXtc task default for the 'rebuild' flag to true. Stopped touching code as part of the rebuild config, because it is semantically incorrect. Cleaned up warnings. Added the property xtcDefaultRebuild, whose default is true, which will always add the --rebuild flag to the xcc process launched from the XtcCompileTask, as we know it is stale. This forces XTC to recompile XTC if any change has been made in Javatools, to ensure build soundness. To disable this, the property should be set to false, which will call xcc without --rebuild, potentially creating errors later, if the Javatools behavior changes.

Please test this with the xtcDefaultRebuild flag set to true and false and ensure it does what I think it does. Note that Gradle is smart enough to treat a very trivial whitespace change, for example, as non significant, so you need to add a small and trivial change, e.g. adding a line with a comment in Component.java, or something, that is more than just white space to produce a different hash value for the build cache. 

You can run with the old behavior with 

-PxtcDefaultRebuild=false on the Gradle command line, or by setting the corresponding environment variable "ORG_GRADLE_PROJECT_xtcDefaultRebuild=false", or by adding it on a per project basis, or any other variant of Gradle property assignment or its environment variable equivalents. 

I recommend that you also run with the -PxtcPluginOverrideVerboseLogging property set to true, or its equivalent, the env var ORG_GRADLE_PROJECT_xtcPluginOverrideVerboseLogging. 

That will show you a little bit of extra life cycle information, but nothing close to the info level, so that you can examine generated Java launcher command lines easily.

### DIFF
--- a/javatools_launcher/build.gradle.kts
+++ b/javatools_launcher/build.gradle.kts
@@ -19,6 +19,8 @@ val launcherExecutableDir = layout.projectDirectory.dir("src/main/resources/exe"
 
 val processResources by tasks.registering(Copy::class) {
     from(files(launcherExecutableDir))
+    // TODO: This may cause a warning in IDEA with non-delegated compilation, but it is safe.
+    //   "Cannot resolve resource filtering of . IDEA may fail to build project. Consider using delegated build (enabled by default)."
     exclude("**/README.md")
     eachFile {
         relativePath = RelativePath(true, name)
--- a/plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
+++ b/plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
@@ -296,7 +296,6 @@ public class XtcProjectDelegate extends ProjectDelegate<Void, Void> {
         final var processResourcesTask = tasks.register(processResourcesTaskName, Copy.class);
         final var classesTaskName = getClassesTaskName(sourceSet);
         final var classesTask = tasks.getByName(classesTaskName);
-        final var forceRebuild = resolveXtcCompileExtension(project).getForceRebuild();
 
         // In Java, the classes task would depend on process resources. In XTC, it depends on the compile task, and the compile
         // task needs to work with the output of process resources, so for XTC we attach the process resources task for this source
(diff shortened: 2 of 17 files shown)


javatools_launcher/build.gradle.kts
```diff
@@ -19,6 +19,8 @@ val launcherExecutableDir = layout.projectDirectory.dir("src/main/resources/exe"
 
 val processResources by tasks.registering(Copy::class) {
     from(files(launcherExecutableDir))
+    // TODO: This may cause a warning in IDEA with non-delegated compilation, but it is safe.
```
ggleyzer, line 22: I don't understand this "TODO". If that is safe, what is there to do? What is the action that is necessary?

plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
```diff
@@ -159,6 +159,9 @@ public Void apply(final Void args) {
 
         // This is all config phase. Warn if a project isn't versioned when the XTC plugin is applied, so that we
         // are sure no skew/version conflicts exist for inter-module dependencies and cross publication.
+        if ("test".equals(projectName)) {
```
ggleyzer, line 162: I don't think you meant this to be committed :)

plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
```diff
@@ -634,6 +634,7 @@ private static String requestToNotation(final ModuleVersionSelector request) {
 
     private void checkProjectIsVersioned() {
         if (UNSPECIFIED.equalsIgnoreCase(project.getVersion().toString())) {
+            new Throwable().printStackTrace();
```
ggleyzer, line 637: ditto

javatools_launcher/build.gradle.kts
lagergren: It is there to make users aware of that this is the reason behind the yellow scary looking alert thst can't be disabled, but is tracked as to be fixed by JetBrains and that it is not cause for concern. I hope that if someone notices this warning in IntelliJ and actually just clicks on the warning, so that intellij jumps to this build script, then just maybe I won't get the question "Have you seen this warning in IntelliJ? Why does it do that? Should it do that? Our intellij build is broken?" 5-10 times. This has already been the case for the "borrow the resource path from lib_ecstasy in javatools to make javatools is runnable as a Java app config in IntelliJ" warning and I keep explaining, but the question keeps being asked.

plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
lagergren: Nope sorry. Fixing

plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
lagergren: Fixing.

plugin/src/main/java/org/xtclang/plugin/tasks/XtcCompileTask.java
```diff
@@ -219,12 +213,18 @@ public void executeTask() {
             args.add("-r", processedResourcesDir.getAbsolutePath());
         }
 
+        final var shouldRebuild = getRebuild().get();
+        if (!shouldRebuild && overrideVerboseLogging) {
```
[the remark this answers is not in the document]
lagergren: Fixing. The outermost resolution is enough. I put this in because you can actually run with defaults and disable rebuild at more fine grained levels, and may not intend to, which actually was the case for the manualTests compile tasks for a bit.

javatools_launcher/build.gradle.kts
ggleyzer: I see. Maybe simply change "TODO" to "NOTE:"

ggleyzer (approved):

PR #185, opened 2024-03-13
Add back-end compiler

cliffclick:
Clean rebase vs master this time
Lost all history


ggleyzer (commented):
Why is this file here? It looks like a copy of `javatools/src/main/resources/errors.properties` file

ggleyzer:
My suggestion: revert the 'error.properties' and all changes in 'manualTests' and we are good to go

cpurdy (changes requested):
Two tests in Compare.x were removed. I'm pretty sure that's not what you meant to do. :)

It looks like the change you want to make in the TCK tests is to have a run() method in each one. I guess when JK is done with the test harness work, we'll just remove all of the run() methods, so this is fine for now.

javatools_backend/MANIFEST.MF
cpurdy, line 1: I'm assuming that this is necessary because there's no gradle build for the new work?

tck/src/main/x/tck/constructors/Basic.x
cpurdy, line 1: We don't want to lose the old tests here. Nothing wrong with adding your array based tests instead of list based, but let's not remove the old ones.

doc/examples/eightQueens.x
cpurdy, line 1: don't commit this file (I'll eventually add this from my examples branch when it's ready)

doc/examples/nQueens.x
cpurdy, line 1: don't commit this file (I'll eventually add this from my examples branch when it's ready)

cpurdy (commented):
Looking good overall but I left some notes after going through with Gene :)

lagergren:
And this would have been [pretty much your entire Gradle integration](https://gist.github.com/lagergren/7c45edb9a55b2523d77aa75ed0628f23
). All your stuff hooked up to rest of the XVM build, with knowledge about the XTC ecosystem and its artifacts. 

If you, for example, need a javatools.jar for this world, you refer to its unique single source of truth name, and no matter if it's built or not, on the Internet or not, or needs to rebuild, that is not anything you need to worry about.


```
 ~/src/xtc/xvm> ./gradlew javatools_backend:installDist                                                                                                                                                                                                                                                                                ✔ 

BUILD SUCCESSFUL in 1s
43 actionable tasks: 6 executed, 1 from cache, 36 up-to-date

 ~/src/xtc/xvm> find javatools_backend/build/install                                                                                                                                                                                                                                                                                   ✔ 
javatools_backend/build/install
javatools_backend/build/install/javatools-backend
javatools_backend/build/install/javatools-backend/bin
javatools_backend/build/install/javatools-backend/bin/javatools-backend
javatools_backend/build/install/javatools-backend/bin/javatools-backend.bat
javatools_backend/build/install/javatools-backend/lib
javatools_backend/build/install/javatools-backend/lib/javatools-backend-0.4.43.jar

 ~/src/xtc/xvm> ./javatools_backend/build/install/javatools-backend/bin/javatools-backend                                                                                                                                                                                                                                              ✔ 
Usage: xec (-L path)* [-M main] file.xtc args


```

lagergren:
> Why is this file here? It looks like a copy of `javatools/src/main/resources/errors.properties` file

The backend project compiles its own javatools.jar, apparently. 

> My suggestion: revert the 'error.properties' and all changes in 'manualTests' and we are good to go



> Why is this file here? It looks like a copy of `javatools/src/main/resources/errors.properties` file

My guess is that it's there because the Makefile tries to build its own javatools.jar. This also includes adding a MANIFEST.MF file, normally dynamically generated by the javatools:jar task, with dynamic version information from the build. However, this one is frozen forever in time at v0.4.3.  At the same time, other things that are in the ordinary javatools.jar, like the implicit.x definitions from lib_ecstasy and the javatools_utils (and the information extracted from *its* manifest) don't seem to be duplicated. At least not explicitly. 

It does, however, have the exact same file name, which makes me predict that it's only a matter of time before someone confuses is with the proper javatools.jar, tries to run something, and gets a build breakage. And that I will be the one who gets the question about why it does not work. 

At the same time the "Makefile" still seems to refer on the ephemeral xdk/build/xdk/install directory, which may or may not be there, depending on if an installDist task has been executed. If it's there, it will contain whatever version of the XDK build that was last executed, maybe a current one, and maybe something from weeks ago. If it's not there, it will not, as far as I can tell, not be built on demand.

cliffclick:
As for MANIFEST & errors.properties - I built a javatools.jar, then tried to use it.  Needed those files at least, so I cloned 'em for the nonce, then my internet went out.  They were always meant to be temp files until I figured out a proper way to gen those.  Don't care if you nuke 'em or not.

ggleyzer:
> As for MANIFEST & errors.properties - I built a javatools.jar, then tried to use it. Needed those files at least, so I cloned 'em for the nonce, then my internet went out. They were always meant to be temp files until I figured out a proper way to gen those. Don't care if you nuke 'em or not.

So, let's "nuke" - not merge them :)

ggleyzer (approved):
Please move the manualTests changes to a new sub dir and we are good to go

cpurdy (approved):

PR #188, opened 2024-03-15
Fix the compilation of short-hand property declarations

ggleyzer:
There are three different fixes in this change:

1. There was a incorrect call to `setParent` with a `null` argument, causing an AnnotatedTypeExpression to become "orphaned" and causing an NPE during the compilation.

2. For short-hand class declarations, we create a synthetic PropertyDeclarationStatement and used to simply ask it to create corresponding structures (properties). The problem with this approach was that the newly created PDS was immediately "orphaned" and therefore never validated properly. The change in TypeCompositionStatement.java adds it to the statement body and catches up to the current (Registered) stage, which does everything it used to do *plus* allowing the validation logic to be performed on it at during the validation phase.

3. The compiler fix revealed an incorrect class declaration in TxManager.x. The `Requirement` const used to be declared as `static`, but it refers to `DBObjectImpl`, which is a typedef that depends on the formal type parameter `Schema`, which is **not avaliable** outside of the TxManager context.

### DIFF
--- a/javatools/src/main/java/org/xvm/compiler/ast/PropertyDeclarationStatement.java
+++ b/javatools/src/main/java/org/xvm/compiler/ast/PropertyDeclarationStatement.java
@@ -888,6 +888,16 @@ public class PropertyDeclarationStatement
                 }
             }
 
+        // this is not quite correct; if there were annotations extracted from the type, we will
+        // see them doubled in the string; very rare inconvenience, not worth any effort atm
+        if (annotations != null)
+            {
+            for (AnnotationExpression annotation : annotations)
+                {
+                sb.append(annotation)
+                  .append(' ');
+                }
+            }
         sb.append(type)
                 .append(' ')
           .append(name.getValueText());


manualTests/src/main/x/TestSimple.x
```diff
@@ -1,34 +1,10 @@
-@TerminalApp("Simple command tool test")
+// NOTE: THE COMPILATION OF THIS MODULE SHOULD PRODUCE TWO COMPILER ERRORS
 module TestSimple {
-    package cli import cli.xtclang.org;
-
```
lagergren, line 4: In the interest of building a corpus with large amounts of XTC code, runnable or negative in that it should not run or compile, can we hang on to old versions of TestSimple somewhere? Even in another repository if you don't want them in the XDK. I read as much XTC code as I can find, so I feel it would be nice to not lose whatever thing was tested in older versions of TestSimple. If they need cleanup, I still think they could be kept somewhere in some corpus repo under a directory from which they are moved to positive or negative examples after we have cleaned them up. Having this incremental effort in place costs us very little, but we do collect more and more XTC code for a corpus that will likely be very useful.  Mostly for people trying to learn XTC, but also for things like training GitHub CoPilot.

lagergren (approved):

manualTests/src/main/x/TestSimple.x
ggleyzer: Well... they are preserved in the git history, aren't they?

ggleyzer: I use TestSimple as a "code" part of the issue description/explanation/reminder which is also not available in the repository

cpurdy:
Reviewed latest with Gene -- looks good!

PR #189, opened 2024-03-15
Refactor webauth API

ggleyzer:
Move HashInfo from webauth.User to Realm along with `createHashes` method


lib_web/src/main/x/web/security/Realm.x
cpurdy, line 1: Probably I'd normally s/creation/created, but it's fine.

cpurdy (approved):
Looks like as we discussed.

PR #190, opened 2024-03-20
Change the Console.readLine() API

ggleyzer:
The gist of the change is to add the "prompt" argument to the "readLine" API:
    
    String readLine(String prompt = "", Boolean suppressEcho = False)
   
First, IMHO it's not a bad consolidation of steps that basically **always** come together - printing a prompt, suppressing a new line, immediately followed by the readLine() call.

However, I would not consider changing it right now if I didn't step on the following anomaly.

While trying to integrate [jline](https://github.com/jline/jline3) support into the native [TerminalConsole](https://github.com/xtclang/xvm/blob/master/javatools_bridge/src/main/x/_native/TerminalConsole.x) I observed a very weird behavior. When I used [the readLine API that takes a prompt as an argument](https://github.com/jline/jline3/blob/19c031c0bccaeea6a4346bc82c04c7bd3ec18888/reader/src/main/java/org/jline/reader/LineReader.java#L586), everything behaved correctly. If I did it in two steps, printing the prompt first and calling prompt-less readLine() method next, the line editing (back space) could remove the prompt as it were a part of the entered string itself. 

To avoid that behavior we would need either report a bug with jline and wait for a resolution or change our "readLine" to take the prompt.

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/xTerminalConsole.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/xTerminalConsole.java
@@ -61,43 +61,6 @@ public class xTerminalConsole
         return pool().ensureEcstasyTypeConstant("io.Console");
         }
 
-    @Override
-    public int invokeNative1(Frame frame, MethodStructure method,
-                             ObjectHandle hTarget, ObjectHandle hArg, int iReturn)
-        {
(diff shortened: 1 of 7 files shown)


cpurdy (approved):
Implementation looks fine, as Gene and I discussed before he did the changes.

PR #192, opened 2024-03-20
Cc xcodes

cliffclick:
New backend runs the tck tuples tests.
Includes adding run methods to the tcks missing them, to allow testing via the new backend, and turning on a tck accidentally turned off.
Every touched file, also has trailing WS removed.


.dir-locals.el
```diff
@@ -0,0 +1,6 @@
+;; List of lists
+(
+ ;; Each list is a emacs mode, dot, list of vars.
```
lagergren, line 3: Remove these from the repo and add them to gitignore.

javatools_backend/src/main/java/org/xvm/xtc/ast/InvokeAST.java
```diff
@@ -4,11 +4,12 @@
...
 import org.xvm.xtc.cons.Const;
 import org.xvm.xtc.cons.MethodCon;
+import org.xvm.xec.ecstasy.AbstractRange;
 import org.xvm.util.S;
 import org.xvm.util.SB;
 
 public class InvokeAST extends AST {
-  String _meth;
+  String _meth, _slice_tmp;
   final boolean _async;
```
lagergren, line 13: Can we not have these whitespace diffs? It's a significant part of the pull request, so it's hard to find the actual "payload".  

More importantly, rebasing over a change with lots of whitespace diffs will increase the number of times where git has to conservatively fall back to manual resolution. You are actually the one who has suffered the most because of this. 

I strongly suggest that whatever home rolled solution you use for code formatting auto removes any line with just whitespace in it, replacing it with a line with length 0

tck/src/main/x/tck/tuples/Basic.x
```diff
@@ -7,11 +7,11 @@ class Basic {
...
         testEquality();
-        testVoidConv();
-        testIntConv();
+        //testVoidConv();
+        //testIntConv();
         testSlice();
     }
-    
+
     @Test
```
lagergren, line 15: Remove whitespace diff to consistent with the rest of the code base, including to Basic.x itself, which now has both len 0 whitespace lines as well as whitespace line with spaces.

javatools_backend/src/main/java/org/xvm/xtc/ast/InvokeAST.java
ggleyzer: The change is doing exactly that - removing non-zero length empty lines

ggleyzer (approved):
Please address Markus's comment regarding the .dir-locals.el file; the rest looks good

tck/src/main/x/tck/tuples/Basic.x
cpurdy: This is something that we should figure out together how to tackle, as it will save us all a lot of frustration over time (it's not just Marcus who will benefit :)

And good news: Gene says that Cliff has already addressed this problem, which is why this diff exists (and should be a one-time thing).

javatools_backend/src/main/java/org/xvm/xtc/ast/InvokeAST.java
cliffclick: And Marcus, can we please stop with the hugely judgemental comments.
They are insulting, and are not helping a healthy working environment.
As for "home grown", its the same tool that built all the tools your using in your daily life - its been in widespread use for 40 years.  
Per *YOUR REQUEST* I flipped the "strip trailing whitespace on save" flag, and as a natural consequence every file touched afterwards will 1-time include some WS diffs.

lagergren: Ah! I am sorry! I had it the other way around. The reason I picked this up as whitespaces being introduced rather than removed, was that I remember noticing them being added in your first merge request to master. I then asked if they could be removed before getting merged. I assumed that this was the case. Hence, I misread the diff going the other way while checking the review, and interpreted it that they were somehow back. Thinking that they had been reintroduced there, I pointed it out again, because as we know, white space diffs and git turn automatic rebases into manual ones at too many occasions. Standardising early can actually matter when it comes to avoiding future manual labor.

PR #193, opened 2024-03-21
Add jline support

ggleyzer:
It's quite useful for CLI applications to be able to edit the command lines and having a searchable command history. This draft PR is an experimentation how to achieve that using [jline library](https://github.com/jline/jline3)

The only question is how to deliver the jline.jar. Should it be merged with javatools.jar or delivered independently?

### DIFF
--- a/build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
+++ b/build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
@@ -97,7 +97,7 @@ fun Project.printResolvedConfigFiles(level: LogLevel = LIFECYCLE, configName: St
     val files = config.resolvedConfiguration.resolvedArtifacts.map { it.file }
     logger.log(level, "$prefix Configuration '$configName' has ${files.size} files:")
     files.forEach {
-        logger.log(level, "$prefix    file: '$it'")
+        logger.log(level, "$prefix file: '$it'")
     }
 }
 
@@ -105,16 +105,14 @@ fun Project.printTaskInputs(level: LogLevel = LIFECYCLE, taskName: String) {
     val task = tasks.getByName(taskName)
     val inputs = task.inputs.files
     logger.log(level, "$prefix Task '$taskName' has ${inputs.count()} inputs:")
-    inputs.forEach { logger.log(level, "$prefix     input: '$it' (type: ${it.javaClass.name})") }
-    //inputs.asFileTree.forEach { logger.log(level, "$prefix   input : '$it'") }
+    inputs.forEach { logger.log(level, "$prefix input: '$it' (type: ${it.javaClass.name})") }
 }
 
 fun Project.printTaskOutputs(level: LogLevel = LIFECYCLE, taskName: String) {
     val task = tasks.getByName(taskName)
     val outputs = task.outputs.files
     logger.log(level, "$prefix Task '$taskName' has ${outputs.count()} outputs:")
-    outputs.forEach { logger.log(level, "$prefix     output: '$it' (type: ${it.javaClass.name})") }
-    //outputs.asFileTree.forEach { logger.log(level, "$prefix   output: '$it'") }
+    outputs.forEach { logger.log(level, "$prefix output: '$it' (type: ${it.javaClass.name})") }
 }
 
 fun Project.printResolvedConfigFile(level: LogLevel = LIFECYCLE, configName: String) {
@@ -123,7 +121,7 @@ fun Project.printResolvedConfigFile(level: LogLevel = LIFECYCLE, configName: Str
     files?.resolvedConfiguration?.resolvedArtifacts?.map { it.file }?.also { f ->
         logger.log(level, "$prefix Configuration '$configName' has ${f.size} files:")
         f.forEach {
-            logger.log(level, "$prefix    Path: '${it.absolutePath}'")
+            logger.log(level, "$prefix Path: '${it.absolutePath}'")
         }
     }
 }
@@ -136,9 +134,9 @@ fun Project.printTaskDependencies(level: LogLevel = LIFECYCLE, taskName: String)
     logger.log(level, "$prefix $projectName.printTaskDependencies('$taskName'):")
 
     val parents = task.taskDependencies.getDependencies(task).toSet()
-    logger.log(level, "$prefix     Task '$projectName:$taskName' depends on ${parents.size} other tasks.")
+    logger.log(level, "$prefix Task '$projectName:$taskName' depends on ${parents.size} other tasks.")
     parents.forEach {
-        logger.log(level, "$prefix             Task '$projectName:$taskName' <- dependsOn: '${it.project.name}:${it.name}'")
+        logger.log(level, "$prefix Task '$projectName:$taskName' <- dependsOn: '${it.project.name}:${it.name}'")
     }
     val children = project.tasks.filter {
         var match = false // TODO: Better kotlin.
@@ -150,9 +148,9 @@ fun Project.printTaskDependencies(level: LogLevel = LIFECYCLE, taskName: String)
         }
         match
     }.toSet()
-    logger.log(level, "$prefix     Task '$projectName:$taskName' is a dependency of ${children.size} other tasks.")
+    logger.log(level, "$prefix Task '$projectName:$taskName' is a dependency of ${children.size} other tasks.")
     children.forEach {
-        logger.log(level, "$prefix         Task '$projectName:$taskName' -> isDependencyOf: '$projectName:${it.name}'")
+        logger.log(level, "$prefix Task '$projectName:$taskName' -> isDependencyOf: '$projectName:${it.name}'")
     }
 }
 
@@ -171,11 +169,11 @@ fun Project.printPublications(level: LogLevel = LIFECYCLE) {
     val count = publications.size
     logger.log(level, "$prefix Project '$projectName' has $count publications.")
     publications.forEachIndexed { i, it ->
-        logger.log(level, "$prefix     (${i + 1} / $count) Publication: '$projectName:${it.name}' (type: ${it::class}")
+        logger.log(level, "$prefix (${i + 1} / $count) Publication: '$projectName:${it.name}' (type: ${it::class}")
         if (it is MavenPublication) {
-            logger.log(level, "$prefix     Publication '${projectName}.${it.name}' has ${it.artifacts.size} artifacts.")
+            logger.log(level, "$prefix Publication '${projectName}.${it.name}' has ${it.artifacts.size} artifacts.")
             it.artifacts.forEachIndexed { j, artifact ->
-                logger.log(level, "$prefix         (${j + 1} / ${it.artifacts.size}) Artifact: '$artifact'")
+                logger.log(level, "$prefix (${j + 1} / ${it.artifacts.size}) Artifact: '$artifact'")
             }
         }
     }
--- a/build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
+++ b/build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
@@ -188,7 +188,7 @@ fun Project.printPublications(level: LogLevel = LIFECYCLE) {
  */
 class DebugBuild(project: Project) : XdkProjectBuildLogic(project) {
     companion object {
-        fun verifyJarFileContents(project: Project, required: List<String>, size: Int = -1) {
+        fun verifyJarFileContents(project: Project, required: List<String>, size: Int = -1): Int {
             val jar = project.tasks.getByName("jar").outputs.files.singleFile
             val contents = jarContents(jar)
 
@@ -206,6 +206,8 @@ class DebugBuild(project: Project) : XdkProjectBuildLogic(project) {
                     throw project.buildException("ERROR: Corrupted jar file; needs to contain entry matching '$it'")
                 }
             }
+
+            return contents.size
         }
 
         private fun jarContents(jarFile: File): Set<String> {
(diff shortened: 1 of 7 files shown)


ggleyzer:
In our conversation with Marcus, he argues that the best way to deliver the libraries separately. Markus wrote:

>  I think you should consider not having the jline.jar expanded in the javatools.jar ... If you look inside the jar, does it only contain the jline contents or does it contain all manners of transitive dependencies from jline and other parts of the world too?
And because the javatools.jar is exactly what Maven means with an "application". An "application" is something with a main entry point (the Launcher) in main class, with dependencies.  (in this case the only third party one is jline).  Normally you would do a two liner to create a useful application distribution. You would apply the build in application plugin in the plugin section. Then you would declare an implementation dependency in the dependencies section to the jline artifact.  Then you would run the install dist. Then you would end up with two start scripts, one for windows and one for unix in a bin dir, a lib dir with the javatools.jar and the jline.jar as separate files, and a resolution mechanism to run this as a single classpath through the init scripts.  This plugs directly into things like creating a binary exe distribution with jlink/jreleaser or creating releases, using a very standardized layout.
If you apply the application plugin, it automatically applies the distribution plugin. The default distribution which is e.g. what goes into the default distTar and the default distZip tasks will contain bin/javatools (a launcher script, which also is understood by Gradle, so that you don't need to use it explicitly if you use javatools elsewhere), lib/javatools.jar and lib/jline.jar.   I even think it adds whatever the java key for "here is a classpath to other jars dependent to me, the javatools.jar and my manifest" so that if you just run -jar javatools.jar, it will automatically figure out it has a sibling jar file that it needs as a dependency.
When you do shadow jars, you may end up with a HUGE amount of extra stuff that you don't need. And it becomes more bloated to more dependencies you add.  I would opt for using the existing mechanisms for applications and distributions, which includes manifest code that uses the "Class-Path:" attribute in the javatools.jar mainfest. I think it's automatically written by the aplpication plugin, and you don't even have to override anything or write any code.  In that case, the xdk distribution would contain the javatools.jar and the jlink.jar (possibly under javatools/lib to avoid cluttering up the xtc specific javatools directory). The javatools.jar would just refer to the relative Class-Path attribute in its manifest, with "Class-Path: ./lib/jline-verison.jar", or whatever it looks like, and anyone starting java -jar javatools.jar ... will automatically get the jline jar loaded.
The good thing about this if it works like I think it does, is that you need to write very little code for it.  You would enable the application plugin, add a jline dependency by alias in the libs.versions.toml file, and add an implementation(libs.jline) to the dependencies in the javatools project.
And likely a slight weak to the xdk distribution contents so it also brings the libs with it.   The XDK should actually be the main application class, which is what andres was talking about before.   Anyway - I recommend strongly that if we have dependencies, we distribute them outside the jar, use the application plugin to avoid having to manually do much , including adding the manifest entry, and that an unpacked xtc distribution has a libs directory under the javatools directory, where all the dependencies for our javatools.jar reside as separate jar files.

javatools/build.gradle.kts
```diff
@@ -98,6 +101,23 @@ val jar by tasks.existing(Jar::class) {
...
+ *       get the jline.jar into our distribution (installDist task) and point to it in the manifest
+ *       of the javatools.jar
+ */
+val shadowJar by tasks.existing(ShadowJar::class) {
+    dependsOn(jar)
+
+    archiveBaseName = "javatools"
+
+    from(tasks.jar.get().archiveFile)
+
```
lagergren, line 115: You know you are building multiple jars, right? The shadow jar gets a classifier "all" if you take a look at the build results. This is not the javatools.jar that ends up part of the installDist, but rather the one without jline as a dependency. I have taken a look inside the "all" jar, and it seems to contain only the jline classes without overly transitive dependencies, which is good, but I don't think it works as a dependency for the platform right now, if the distribution ships the standard one. This is partly why I wanted to avoid the shadowJar. I guess we have to make other modules aware of the twain jar artifacts somehow. The classifier is probably fine, maven wise, if you publish a maven component it will default to also building a "source" classifier zip file and a "javadoc" classifier zip file, but I'm not sure what the preferred way of shipping the javatools jar with the all classifier, and what needs to be changed in its dependencies. The xtcJavaTools configuration probably resolves the standard jar still, and other dependencies on it. The plugin looks for a javatools.jar on its classpath, and is probably bundle with the javatools.jar.

ggleyzer: Yes, I know that. To test, I currently make the shadow and copy the '-all' jar to the install dir manually:
`./gradlew :javatools:shadowJar;cp /Users/ggleyzer/Development/xvm/javatools/build/libs/javatools-0.4.43-all.jar /Users/ggleyzer/Development/xvm/xdk/build/install/xdk/javatools/javatools.jar`

javatools/build.gradle.kts
```diff
@@ -98,6 +101,23 @@ val jar by tasks.existing(Jar::class) {
...
+ * TODO: Marcus suggests that merging the jars is not a good way to go. Instead we simply need to
+ *       get the jline.jar into our distribution (installDist task) and point to it in the manifest
+ *       of the javatools.jar
+ */
+val shadowJar by tasks.existing(ShadowJar::class) {
+    dependsOn(jar)
+
+    archiveBaseName = "javatools"
+
+    from(tasks.jar.get().archiveFile)
```
lagergren, line 114: This forces the existing javatools fat jar, that master uses, to get executed archived, and put together just by configuring the build. Everything not in a doLast or doFirst block will run at configuration time, so this means that if gradle just starts looking at the build, this from statement forces it to build the non shadow jar. That is probably not what you want.  Dependent tasks should be resolved lazy and as late as possible. 

The tasks.jar is a task container, i.e. provider of a task. By calling get outside a doLast and doFirst block you force Gradle to compute that task as soon at is sees this statement by scanning the code. 

You are also asking for an output of the task, the archiveFile (javatools-0.4.43.jar) in the configuration phase after the get, so you can be sure just loading this file with gradle will force build javatools.

ggleyzer: I would really appreciate if you fixed it to do what you suggested earlier.

javatools/build.gradle.kts
```diff
@@ -98,6 +101,23 @@ val jar by tasks.existing(Jar::class) {
...
+
+    dependencies {
+        include(dependency(libs.jline.get()))
+    }
```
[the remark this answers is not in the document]
ggleyzer: Could you please do what's "correct" and "right"? I would really appreciate your help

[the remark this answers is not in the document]
lagergren: I have now done what I think is "correct" and "right". At the moment I remain with javatools.jar as a fat jar including all its dependencies, but I have sketched out a configuration that I can resolve with the application plugin for the xdk distribution in the xdk package, that will make it very easy to detach any dependencies and place them elsewhere, with a Class-Path manifest entry in the javatools.jar artifact. 

The split up solution should be completely invisible to the user, work the same way, and have the advantage that we can start taking stuff out of javatools, instead of watching it grow fatter and fatter. Javatools is already a "god class dependency" given that everything else in the build depends on it. If we compiled javatools.jar and javatools-utils separately, and tried to increase granularity even further, so that a single class change under javatools or javatools utils doesn't necessary require rebuilding the world, we are on the right path.  Fat jars and "god classes" that the entire rest of the world depend on, with no clear separation of its contents, usually mean that 80% of everything gets repackaged, affects the checksums, and throws away any dependency details known, replacing them with the "EVERYTHING" package. 

Homework: please read and understand the lazy map pattern and the distribution and application plugins in the gradle user guide. Given this transform, it should be clear that it's not a very complicated thing, and we have everything to gain from using Gradle built-in plugins, defaults, layouts and expected locations wherever we can.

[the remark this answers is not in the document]
ggleyzer: As a side note, specific to jline library, in the future we can trim the "merged in" content. If you look at the [documentation](https://github.com/jline/jline3), there are parts of the library that we definitely have no use for, e.g.: `jline-remote-ssh`, `jline-remote-telnet`, etc. This would require a bit more configuration work...

build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
```diff
@@ -97,24 +97,22 @@ fun Project.printResolvedConfigFiles(level: LogLevel = LIFECYCLE, configName: St
...
+        logger.log(level, "$prefix file: '$it'")
     }
 }
 
 fun Project.printTaskInputs(level: LogLevel = LIFECYCLE, taskName: String) {
     val task = tasks.getByName(taskName)
     val inputs = task.inputs.files
     logger.log(level, "$prefix Task '$taskName' has ${inputs.count()} inputs:")
-    inputs.forEach { logger.log(level, "$prefix     input: '$it' (type: ${it.javaClass.name})") }
-    //inputs.asFileTree.forEach { logger.log(level, "$prefix   input : '$it'") }
```
lagergren, line 109: Why did you remove this code? It's useful in that a directory doesn't tell you a lot. If it has files (which it will when resolved, hence printTaskInputs should only be called from the execution phase), it's quite useful to see what files it contains. However, it should probably be on a more fine grained log level, but I haven't implemented that.

javatools/build.gradle.kts
```diff
@@ -105,14 +120,15 @@ val assemble by tasks.existing {
 val sanityCheckJar by tasks.registering {
     group = VERIFICATION_GROUP
     description =
-        "If the properties are enabled, verify that the javatools.jar file is sane (contains expected packages and files), and optionally, that it has a certain number of entries."
+        "If the properties are enabled, verify that the javatools.jar file is sane " +
```
lagergren, line 123: Use a """ literal instead of plussing things together. This looks weird.

javatools/build.gradle.kts
```diff
@@ -75,25 +76,39 @@ val jar by tasks.existing(Jar::class) {
...
             "Name" to "/org/xvm/",
             "Specification-Title" to "xvm",
             "Specification-Version" to version,
             "Specification-Vendor" to "xtclang.org",
             "Implementation-Title" to "xvm-prototype",
             "Implementation-Version" to version,
-            "Implementation-Vendor" to "xtclang.org"
+            "Implementation-Vendor" to "xtclang.org",
+            "Sealed" to "true"
         )
```
lagergren, line 112: Where did the Xdk-Version go? You have just made it impossible for the plugin to determine which javatools it may have on the runtime classpath.

build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
ggleyzer: I didn't remove the code, only the commented out line, which seemed to be a trivial variation on the un-commented one. The diff here is misleading - please take a look at the file itself

javatools/build.gradle.kts
ggleyzer: I didn't understand the reason of that to be. Why was it called Xdk-Version? This is a javatools.jar, what does it have to do with the XDK?

ggleyzer: I'll revert, but it needs to be documented

build-logic/common-plugins/src/main/kotlin/DebugBuild.kt
```diff
@@ -97,24 +97,22 @@ fun Project.printResolvedConfigFiles(level: LogLevel = LIFECYCLE, configName: St
     val files = config.resolvedConfiguration.resolvedArtifacts.map { it.file }
     logger.log(level, "$prefix Configuration '$configName' has ${files.size} files:")
     files.forEach {
-        logger.log(level, "$prefix    file: '$it'")
+        logger.log(level, "$prefix file: '$it'")
     }
```
lagergren, line 101: This whitespace was deliberately there, since there is a lot of information, and it is significantly simpler to read the output where we have a lot of resolved artifacts, if we have a title line with the name of the config and the number of files, followed by each of these files on a separate line, but indented.

ggleyzer: I'm using the "--info" quite often now and the amount of printout is overwhelming. I can only analyze it with some additional tools (less, grep, etc.) Didn't realize that that particular extra space was important, looked like a random omission.

lagergren (approved):
Approved. If we start adding more dependencies to javatools that are third party jars/artifacts, we should continue with making them explicit in the shipment and referred to from javatools.jar by Class-Path/application plugin auto resolution.

javatools/build.gradle.kts
lagergren: No serious preference. For an output with constant changes or $ resolution of variables and function calls, I prefer to be "fully escaped" when doing a multi line printout. The + has been in the way before, for example when I reformat code, and it's added, then change the message or add stuff in it, and then auto reformat, and then I get another plus. Suddenly it's a combination of three word lines and thirty word lines that are plussed together. Here, I can auto format and get it adjusted without pluses, and IntelliJ tries to create equally long lines, lower than the maximum number of characters, and will word wrap nicely. That is definitely more readable. But The real question is if we need that huge description text. I did want to mention the possibility of checking entry count, though. And I do want to keep the verification mechanism, since it catches dependency bugs early when building a javatools. Accidentally missing something in javatools can trigger errors at runtime, or much later in the build, so that's why this semi-ugly pattern matcher is here in the first place. It actually started saving me serious time when I changed dependency configurations and ways to build the javatools.

javatools/build.gradle.kts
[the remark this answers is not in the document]
lagergren: This is typically done by the app vendor distributing the app as different packages, like jline-all, jline-api etc, but in this case they only seem to have a single master artifact on maven central.  Adding modifications to what that artifact contains, as a user, is probably a bad idea. While we do have absolute version control and know that it won't update to a new jline version unless we tell it to, any filter we implement manually will be safe until we change versions. But if we do that, for example because there is a security problem or a feature in the new version that we want, we would have to manually recheck our manual filter, which is potential maintenance cost added.  The artifacts should be considered "unmodifiable" when they are brought in from a repository.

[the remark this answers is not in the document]
ggleyzer: > This is typically done by the app vendor distributing the app as different packages, like jline-all, jline-api etc, but in this case they only seem to have a single master artifact on maven central.

This is incorrect; all those libraries do exist separately at the Maven Central:
https://mvnrepository.com/search?q=jline

PR #195, opened 2024-03-22
Add "Linker.collectInjectables" API

ggleyzer:
While working on the Platform functionality, I realize that we're missing support for an important step during application deployment. A deployed module(s) may require a number of injectable properties that would need to be setup by the deployment registration process. To facilitate collection of the necessary data, the UI (or CLI) tool may need to know what are the required injections and present that list to the deployer.

The suggested API would serve that purpose.


lagergren (approved):
Nice and readable change. 

Can this kind of thing be unit tested on the Java side?

ggleyzer:
I'm afraid the API is not quite right. I just realized that there is a chance for multiple injections of the same name but different type - and that would be perfectly legal. Need to change the method return type, maybe `Tuple<String, Type>[]`?

Cam, any suggestions?

PR #206, opened 2024-04-16
Fix the compiler inference inconsistencies

ggleyzer:
The compiler used to have a slightly different type inference logic between AssignmentStatement and  VariableDeclarationStatement. While the former always called "updateLValueFromRValueTypes()" at the end of the validation, the later did not. As a result, as Cliff first discovered, processing of 
   
    Int? n = 0;
    ... // use n 
    
was quite different from 

    Int? n;
    n = 0;
    ... // use n 
    
In the second scenario the compiler would "know" that "n" is not Null, while in the first one it did not, which is definitely wrong. The obvious solution was to add the following line at the end of VariableDeclarationStatement#updateLValueFromRValueTypes(). 

    getLValueExpression().updateLValueFromRValueTypes(ctx, branch, fCond, aTypes);

However, this simple fix uncovered a slew of bugs we have never seen before (the obvious reason we have not stepped onto any of those was that that `Int? n; n = 0;` style of variable initialization is quite unusual.

The good news is that the fix in this PR (on top of resolving the compilation inconsistency) made the compiler to uncover a couple of bugs that were not yet known (e.g. [this line was missing](https://github.com/xtclang/xvm/blob/dc66740a5969a5ca74251b2e44e0d8649baa7687/lib_jsondb/src/main/x/jsondb/Client.x#L1474), the Array API was a nothch off, and a couple of other small issues)

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/Register.java
+++ b/javatools/src/main/java/org/xvm/asm/Register.java
@@ -217,7 +217,20 @@ public class Register
     public Register narrowType(TypeConstant typeNarrowed)
         {
         // even when the types are the same, the shadow carries "not-in-place" flag
-        return new ShadowRegister(typeNarrowed, m_sName, f_nOrigIndex);
+        ShadowRegister regShadow = new ShadowRegister(typeNarrowed, m_sName, f_nOrigIndex);
+        TypeConstant   typeReg   = m_typeReg;
+        if (typeReg != null)
+            {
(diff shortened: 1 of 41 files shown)


ggleyzer:
Reviewed w/ Cam. A couple of small fixes were made and pushed

cpurdy (approved):
Reviewed everything this morning with Gene. A few potential future changes, if we decide to go a bit further (e.g. Dima/Cliff on StrInt examples), but for now this gets us to a much better place than we were, with no obvious / actively burning problems.

PR #207, opened 2024-04-18
Harden the List.sorted() contract

ggleyzer:
While working on something, I wrote a simple test similar to this:

        Int[] ints = [3, 1, 2];
        ints.sorted(inPlace=True);
        console.print(ints);

and was quite surprised to see a non-sorted result. In a minute, I realized that our List.sorted() API contract takes the value of `inPlace` argument as a suggestion and not a requirement. If the List (in this case an Array) is immutable it creates a new array, sorts and returns it. I think it's wrong and it should throw.

Proposed is the corresponding API doc and implementation change


cpurdy (approved):
As discussed.

PR #211, opened 2024-04-26
Tag commit and publish snapshot artifacts on pushes (to master).  All is ready for jreleaser releases. 

lagergren:


lagergren:
Draft state, but this should be a fairly basic code pipeline with lots of old code heading in various other directions deleted. 

This implements a build flow for maven artifacts, that corresponds to the xdk-latest channel for brew, whenever a change is pushed to master. If we push a snapshot version, a snapshot will be created and updated, and its label will be created or moved (if it exist) to the commit corresponding to that snapshot build. Packages will be updated in the GitHub Maven repository. 

If we push something that is not a snapshot version, it is currently a no-op, but the behavior should be: if the corresponding tag for the release already exists - fail - we cannot overwrite packages that aren't snapshots, ever. Otherwise, create the tag, tag the commit correponding to this package, and do exactly as above. 

Cleaned up a lot of code and concepts in distributions to use built-in mechanisms where possible. 


Final steps: 
1) Allow publications of new (non-existing) release package publications too. 
2) Do not just publish the non-snapshot release packages to GitHub, but also to Maven Central for the XDK and gradlePluginPortal() for the XDK. We already have credentials and they are in the GitHub workflow secrets. 
3) Hook up a JReleaser flow that creates GitHub releases, self extracting binaries, platform dependent launchers binaries, change logs, etc, that uses the release package as input.

lagergren:
This should work as intended right now.

Every time we commit to master, and we have a SNAPSHOT version in the VERSION file (the normal state of things, x.y.z-SNAPSHOT means "working on release x.y.z but not published yet).  Anyone depending on a snapshot version or "latest.integration", will get the bleeding edge stuff if declaring the xtclang.org GitHub Maven repository (along with read credentials), any time something is pushed to master. 

This corresponds to building manually from a VERSION, that is either a snapshot or not. If it's not a snapshot, we will fail if that release has been published before. If we are a snapshot, we will move any existing snapshot version tag to the last commit in master, and publish from that. It's trivial to also encode the SHA and not just the version, somewhere in the release .


There is also a 

./gradlew xdk:installWithLaunchersDist 

That creates an installation with the correct binary launchers for the build platform. This will be fed into jreleaser to create downloadable installers for releases, change logs, stuff like that. 

We also already support publishing (non snapshots) to gradlePluginPortal, and ALMOST mavenCentral (some credential issue).  Keys are locked down in the org secrets vault on GitHub. 
 

./gradlew publishRemote

build-logic/common-plugins/src/main/kotlin/org.xtclang.build.publish.gradle.kts
```diff
@@ -130,35 +93,46 @@ val listAllLocalPublications by tasks.registering {
...
+                timestamps.forEach { timestamp -> logger.lifecycle("$prefix             Created at: $timestamp") }
             }
-            versions.forEach { logger.lifecycle("$prefix        version: '$it'") }
         }
     }
 }
 
-val deleteAllRemotePublications by tasks.registering {
+/**
+ * Delete all packages. -PpackageName, -PpackageVersion wil lbe respected
```
ggleyzer, line 124: A typo "will"

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
```
ggleyzer, line 3: Do we **need** Gradle 8.7 for this? If we don't, I'd prefer to make this into a different change.

ggleyzer (approved):
I have not tested any of the new functionality, but aside of those two very minor things I commented on everything looks fine

gradle/wrapper/gradle-wrapper.properties
lagergren: I'll rerun the tests with Gradle 8.6 and check if it works. I did upgrade to 8.7 for a reason, there was something that had been fixed, but I can do that separately if the branch as it currently is passes the tests anyway.

PR #212, opened 2024-04-29
Progress tcks and new doc/examples

cliffclick:
Unions.  Elvis.  New tcks.  A host of minor changes, especially iterators.  Makefile/build greatly expanded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,12 @@
-# Are your paths correct?
 #
 # Makefile expects to run at the project root, and not nested inside anywhere
 #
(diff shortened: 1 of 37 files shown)


cpurdy (approved):
Reviewed with Cliff & Gene via Zoom. A few notes (for future PR(s) -- this one can go in as is!)

1. there's a Foo() that got left over unintentionally
2. there was some confusion on "Elvis" re: bast null check
3. we need some explicit operator precedence tests in the TCK in the future
4. we'll need to handle potential name collisions when creating / mapping names (longer term project)

PR #214, opened 2024-04-30
make for mac

cliffclick:
drop -e from echo
makedepends.sh is executable, uses bash not sh
makefile checks version


ggleyzer (approved):
Looks good

PR #215, opened 2024-05-09
Quick fix to avoid asking for the publishAllPublicationsToGitHubRepsitory task, if there are no GitHub credentials, and it's not set up.

lagergren:


ggleyzer (approved):
I trust Marcus it works )

lagergren:
> I trust Marcus it works )

"Trust, but verify!" /Ronald Reagan. 

Let's me at least wait until the checkin tests are green :)

PR #218, opened 2024-05-09
Dump Java system properties in Github workflow

lagergren:
This way, we may be able to better reproduce any issues that may crop up in the manualTests, due to assumptions about path assignments, for example with System.getPropert("java.io.tmpdir")


ggleyzer (approved):
I'm not sure if we need all that info all the time,  but I guess no harm in that...

lagergren:
> I'm not sure if we need all that info all the time, but I guess no harm in that...

Well, you have to go look for it and click "expand", and it's only on the GitHub checkin runner, so better safe than sorry. It doesn't affect anything else.

PR #222, opened 2024-05-09
Runs doc/examples/StringComparisons

cliffclick:
Back to running all the doc/examples, manualTest/.../TestMisc, and tcks.


ggleyzer (approved):
I haven't tested it, but looks right :)

PR #223, opened 2024-05-10
Fix a compiler bug

ggleyzer:
Cliff reported a compiler bug: failure to compile a nested ternary:

    conditional String test(Boolean b0, Boolean b1) {
        return b0
            ? (b1 ? False : (True,"def"))
            : (b1 ? (True,"abc") : (True, "xyz"));
    }

This change fixes it as well as simplifies the BAST production of ReturnStatement


cliffclick (approved):
looks good to me

cpurdy:
@cliffclick  A few small changes will show up in a minute from a zoom review I did with @ggleyzer .

PR #224, opened 2024-05-12
Allow adding extra WebServices to the HttpHandler

ggleyzer:
HttpHandler was designed for "declarative" endpoint discovery, driven by the reflection on WebServices and their methods declarations. As a result of that design all WebServices had to have default constructors and be a part of the compilation unit of the corresponding WebApp. This change removes those limitations by a allowing to:

    a) declare WebServices with non-default constructors 
    b) use WebServices that are external to the underlying WebApp


ggleyzer:
Reviewed w/ Cam

PR #225, opened 2024-05-12
Add various simplifications and more robust GitHub actions for the check-in test

lagergren:
Added various simplifications and more robust GitHub actions for the check-in test. Also added stale data cleaners, but disabled.

This needs to be merged to master for the workflow_dispatch manual activation selector to get activated. None of the cleanup jobs for old maven packages, caches or workflow results are enabled in the first merge. Nor are they run on a repeated schedule with cron. This is to make sure that they work as they should, and we will tune these and debug them in the next pull request after these are merged. 

There, I will add release publication + jreleaser support for release packaging.

### DIFF
--- a/.github/workflows/cleaners.yml
+++ b/.github/workflows/cleaners.yml
@@ -1,0 +1,115 @@
+
+#
+# This is a workflow that is meant to keep our temporary data in check on the server side:
+#
+# Things we want to make sure stay within bounds are:
+#
+#   * GitHub Action caches, especially for Java distribution and Gradle.
+#
+#   * Snapshot publications. GitHub handles SNAPSHOTS in a non standard way, so it's not possible to just
+#     enumerate all versions of a SNAPSHOT package, which of course is wrong, according to the Maven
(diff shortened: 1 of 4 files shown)


lagergren:
Please review. These are minor changes. The biggest one is that we have verified that the Gradle caches work between steps, branches and jobs, so nothing building close in time cannot overwrite the GitHub workflow caches for another commit.  

We also make sure that all installDist/manualTest verification for the test platforms is finished before trying to publish a snapshot artifact. This is to make sure that even if we run on the publishing platform (ubuntu-latest), and windows fails, we do not publish anything that may be a windows-breaker.  

This is all good prep work for the jreleaser configuration which will accept an uploaded build artifact blob consisting of the installWithLaunchersDist destination directories after uploadArtifact. 

We have also added various cleaners that are meant to run either manually dispatched, and on a cron schedule, so that we keep the number of published snapshots low, with only the ~10 latest ones for the same version. Furthermore we may want to delete old caches for old runs, as well as workflow results, when they cease to be relevant.

ggleyzer (approved):
Looks good.

PR #226, opened 2024-05-15
Add reverse proxy route calculation and TLS calculation support

cpurdy:
Note to Gene: This may temporarily break the Platform project


ggleyzer (approved):
Looks good (reviewed together)

PR #227, opened 2024-05-20
Cc tck5

cliffclick:
Elvis & AST rewrite overhaul

### DIFF
--- a/tck/src/main/x/tck/cond/Basic.x
+++ b/tck/src/main/x/tck/cond/Basic.x
@@ -1,0 +1,60 @@
+/**
+ * Basic conditional tests.
+ */
+class Basic {
+
+    void run() {
+        test1();
+        test2();
+        test3();
+        test4();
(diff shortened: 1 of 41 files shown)


tck/src/main/x/tck/cond/Basic.x
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    void test1() {
+        assert !test();
+        conditional String test() { return False; }
+    }
+
+    @Test
+    void test2() {
+        assert String s := test(), s=="abc";
+        conditional String test() { return (True,"abc"); }
```
ggleyzer, line 24: This all looks very neat, but just in case you'd like to simplify it a bit further, a shorthand method definition could be used, for example:

    conditional String test() = False; 
    
and 

    conditional String test() = (True,"abc");

ggleyzer (approved):
Looks good

PR #228, opened 2024-05-27
Add CertificateManager.extractKey() API

ggleyzer:
All the KeyStore API was intentionally design in such a way that if if you ask for a private key [of a key pair] or a symmetrical [secret] key, you would get an object that **could** be used as such a secret, but a developer would have no way to extract/expose the actual key material. Now, for the reverse proxy management project, we came to a requirement to get access/propagate that vulnerable data outside of the single machine boundaries.

The KeyStore [injectable] interface is meant to be read only, and all the keystore mutations are contained in the CertificateManager [injectable] interface. This change adds an API to the latter that allows the developer (at this time most likely just us as developers of the platform) to extract the raw key material from the keystore.

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
@@ -143,8 +143,8 @@ public class xRTCertificateManager
 
         runSilentCommand(
                 "keytool", "-delete",
-                "-alias", hName.getStringValue(),
-                "-keystore", hPath.getStringValue(),
+                "-alias",     hName.getStringValue(),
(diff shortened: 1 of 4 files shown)


cpurdy (approved):
I have two concerns that we need to eventually put to bed: 

- The code is very UNIX/Linux/macOS-centric in its assumptions, so it's unlikely to ever work on Windows (not that anyone uses Windows for servers); we don't have to support Windows, but it should be noted in the doc
-  Putting in args (or derivatives of args) as command line arguments is always very dangerous from a security POV, because it allows a caller (e.g. to revokeCertificate) to pass a name that actually contains commands to execute (the attack is called "command injection"); all arguments that could be provided from any potential unsafe entity must be sanitized (TODO)

PR #229, opened 2024-06-02
Upgraded Gradle to 8.8

lagergren:
This facilitates several useful things for us, among others: 

* Allow version catalog plugin aliases without a version
* Accessors for Settings extensions in Kotlin DSL
* New Gradle lifecycle callbacks
* More classes supported in the configuration cache, which means that it is easier to rewrite the plugin to use the config cache, which would save tremendous amounts of build time. Right now it's some manual work we haven't the time budget for to enable the configuration cache for the XDK build, but once enabled, I see no reason that we shouldn't get massive build speed improvements that would now pretty much any non caching system way out of the water. 
*


ggleyzer (approved):
Looks good

PR #230, opened 2024-06-04
Fix Json serialization for Tuples of enum values

ggleyzer:
Fix serialization of tuples containing enum values, such as:
    
    Tuple<Int, Boolean> tuple = (1, True)

Prior to this change json serialization would blow up at run time.

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/constants/ParameterizedTypeConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/ParameterizedTypeConstant.java
@@ -376,12 +376,9 @@ public class ParameterizedTypeConstant
                 }
             }
 
-        if (fDiff)
-            {
-            return getConstantPool().
-                    ensureParameterizedTypeConstant(constResolved, aconstResolved);
-            }
-        return this;
+        return fDiff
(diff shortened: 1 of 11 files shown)


javatools/src/main/java/org/xvm/asm/constants/ParameterizedTypeConstant.java
```diff
@@ -814,6 +808,34 @@ public ResolutionResult resolveContributedName(
         return result;
         }
 
+    @Override
+    public TypeConstant widenEnumValueTypes()
+        {
+        TypeConstant   constUnderlying = m_constType;
+        boolean        fDiff           = false;
```
cpurdy, line 815: `fDiff` will always be the same as `(aconstOriginal != aconstResolved)`

cpurdy (commented):
Reviewed.

cpurdy (approved):
Approved.

javatools/src/main/java/org/xvm/asm/constants/ParameterizedTypeConstant.java
ggleyzer: Changed accordingly

PR #235, opened 2024-06-18
Create a conversions and formats module 

ggleyzer:
There were converters that resided in web module that were quite useful outside of the web framework, e.g. crypto material generation.

This PR moves the basic codecs and formats functionality from "web" to "convert", while leaving some web-specific parts there


lib_web/src/main/x/web/Client.x
```diff
@@ -63,7 +61,7 @@ interface Client {
     }
 
     /**
-     * The registry for codecs.
+     * The registry for codecs an media types.
```
cpurdy, line 64: "and"

lib_web/src/main/x/web/Registry.x
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Register the passed [MediaType].
+     *
+     * @param mediaType  the `MediaType` to register
+     */
+    void registerMediaType(MediaType mediaType) {
+        for (String extension : mediaType.extensions) {
+            assert mediaTypeByExtension.putIfAbsent(extension, mediaType) as
+                $|Extension {extension.quoted()} has already been registered by
+                 |{mediaTypeByExtension.getOrNull(extension)}"
```
cpurdy, line 152: trailing double quote???

cpurdy (approved):
A few minor things, but otherwise as discussed.

PR #238, opened 2024-08-01
Add lifecycle support to ResourceProvider

ggleyzer:
Currently, the API for ResourceProvider which is used to control injections into a newly created Container doesn't have any lifecycle events. It becomes a problem when a custom ResourceProvider injects a lifecycle aware (Closeable) resource into a child container. When that container is "killed", the resources may leak, since no one properly "closes" them.

This change addresses the issue my making the ResourceProvider to extend Closeable and guaranteeing a RT call to close() when the container goes away.

Cameron suggested a bit more elaborate extension to the ResourceProvider API:

```
/**
 * At the point that the container (for which this ResourceProvider exists) is shut down or
 * killed, this method will be invoked and passed a list of all resources previously supplied
 * that implements the [Closeable] interface.
 *
 * The default implementation of this method invokes the `close()` method on each of those
 * resources.
 *
 * @param resources  each resource previously supplied by this provider that implements the
 *                   [Closeable] interface 
 */
void releaseResources(List<Closeable> resources) = resources.forEach(close);
```

That would make the implementor's job easier, but requires more work by the native code. I'm going forward with an easier a fix that puts onus on the ResourceProvider implementation to collect all necessary resources, but we may change it in the future.


cpurdy (approved):
Reviewed with Gene; looks good!

PR #239, opened 2024-08-13
Introduce native Container API to freeze time; make the console debugger to freeze all containers

ggleyzer:
When you used the Console Debugger for stepping through Ecstasy code, it used to be extremely hard to investigate any timer related logic, since stopping in the debugger didn't prevent the alarms to go off based on the wall clock.

This change allows the debugger to "freeze" time for all containers, dramatically simplifying such a debugging.


ggleyzer:
Reviewed w/ Cam

PR #241, opened 2024-09-12
Command line tool refactor to run tests

thegridman:
This PR is work in progress. 

This PR is for refactoring the command line too (Launcher) to add the `xtc` command so that we can run tests using `xtc test` and add flexibility to be able to easily add other sub-commands to `xtc`.

The code in the PR introduces a new "command" framework to add a hierarchical command structure, where commands can have sub-commands, which could have more sub-commands and so on. For example, `xtc` is a command that has a sub-command `test` so on the command line it is executed using `xtc test <flags...>`

The PR also introduces code to represent command line flags. The `FlagSet` represents a set of flags parsed from the command line. Each command has a set of valid flags. Each `Flag` in a `FlagSet` is types, for example a boolean flag, string flag, etc. Flags and parsing a `FlagSet` is posix compliant. Each flag must have a long name (which is used prefixed with `--` on the command line. Optionally a Flag can have a single character shortcut (used prefixed with a `-` on the command line). Single characters can be concatenated on the command line (e.g. `-abc` is the same as `-a -b -c`)

PR #243, opened 2024-09-23
Simple refactoring of xuint classes

ggleyzer:
This is a refactoring suggestion for JK


thegridman (approved):

plugin/src/main/java/org/xtclang/plugin/tasks/XtcTestTask.java
```diff
@@ -0,0 +1,112 @@
...
+ * from the source set in undefined order. That would be the exact equivalent of calling 'runAllXtc'
+ * (which does not need to be configured, and is a task added by the plugin).
+ */
+
+// TODO: Add modules {} segment to runtime DSL
+// TODO: Add a generic xtcPlugin or xtc extension, where we can set stuff like, e.g. log level for the plugin (which does not redirect)
+// TODO: Add WorkerExecutor and the Gradle Worker API to execute in parallel if there are no dependencies.
+//   Any task with zero defined outputs is not cacheable, which should be enough for all run tasks.
+// TODO: Make the module path/set pattern filterable for the module DSL.
+public abstract class XtcTestTask extends XtcBaseRunTask implements XtcRuntimeExtension {
```
lagergren, line 40: Looks good. Let me know if you need help figuring out any DSL modifications, it should be relatively staightforward by looking at what is there already, but there are some hidden traps. 

One thing that we want to do is to run the Gradle configuration cache, which will significantly speed up the pre-build phase of the gradle jobs - this requires that we do not store projects anywhere in the plugin, which is possible to do, but requires some rewrites.  Enabling the configuration cache will tell us what goes wrong and why, and I was hoping to get a volunteer to this fairly self contained task.

gradle/libs.versions.toml
```diff
@@ -57,17 +57,20 @@ xdk-oodb = { group = "org.xtclang", name = "lib-oodb", version.ref = "xdk" }
...
 javatools-utils = { group = "org.xtclang", name = "javatools-utils", version.ref = "xdk" }
 javatools-turtle = { group = "org.xtclang", name = "javatools-turtle", version.ref = "xdk" }
 javatools-bridge = { group = "org.xtclang", name = "javatools-bridge", version.ref = "xdk" }
 
+hamcrest = { module = "org.hamcrest:hamcrest", version = "2.2" }
 javax-activation = { module = "com.sun.activation:javax.activation", version = "1.2.0" }
 jakarta-xml-bind-api = { module = "jakarta.xml.bind:jakarta.xml.bind-api", version.ref = "jakarta" }
 jaxb-runtime = { module = "org.glassfish.jaxb:jaxb-runtime", version.ref = "jakarta" }
 jline = { module = "org.jline:jline", version.ref = "jline" }
+mockito = {module = "org.mockito:mockito-core", version = "5.11.0"}
```
lagergren, line 74: Hitting ctrl-alt-L in intelliJ adds trailing spaces for me, so that I don't have to think too much about indentation, but I'm not sure if it's a personal setting or propagated to our codebase. We should work on IDE-environment consitency together with language support. I did add support for the Oracle-Sun Java code standard with checkstyle, but this seems to be a thing no one wants to migrate too. Personally, as long as something is consistent, it doesn't matter how it looks, I can read it. The Coherence standard is the only one I have had trouble doing that with, and it only partly has to do with brackets on single lines makes longer methods and frequent page ups / page downs with a laptop size screen. :-).   We should likely implement a  plugin for an official XTC code standard too, and only have one.

thegridman:
I'm closing this PR as the changes have been merged into PR #241

PR #248, opened 2024-11-10
Remove unnecessary type declarations

ggleyzer:
I think it's makes the tests look better if we don't have to explicitly specify types for homogenous JsonObject constants


thegridman (approved):
That looks much cleaner

PR #250, opened 2024-11-12
Enabled explicit Gradle v8.9 for compatibility with the latest setup-gradle GitHub action. Bumped build Gradle to v8.9

lagergren:


lagergren:
Don't forget to do ./gradlew clean, since we upgrade the build Gradle too. This is because the setup-gradle GitHub action wants Gradle >= 8.9 for caching, and it felt unnecessary to have multiple Gradle version taking up our GitHub cache space allocation.

ggleyzer (approved):
Looks good even though some of it is black magic to me :)

lagergren:
> Looks good even though some of it is black magic to me :)

Very little black magic, I can promise you. It's just yaml file executing some scripts written by various script kiddies. For example https://github.com/gradle/actions/blob/main/setup-gradle/README.md

:)

PR #251, opened 2024-11-13
Make JsonObject/JsonArray API more usable

ggleyzer:


ggleyzer:
This is just a playground of ideas...

ggleyzer:
AFAIR, the reason you had to specify the constant type (`Map<String, Doc>:`) was the heterogeneous nature of the example. You have values that are Strings, intLiterals and Nulls, so the compiler cannot infer a proper union type without a hint. If all values were Strings it would not have been necessary

PR #252, opened 2024-11-26
Introduce webcli module

ggleyzer:
Add a facility to easily create REST based CLI application


ggleyzer:
As a part of this change, an array of service objects created by a VAR_S and VAR_SN ops has mutability =Constant

cpurdy (approved):
Note: Also re-ran Cliff's tests to make sure nothing broke

PR #254, opened 2024-12-13
WebSec project: new security module, plug into http & xenia

cpurdy:
Security updates: Create new security module, use that module in the http & xenia modules, and remove requirement for web sessions; refactor Realm API and DB-based realm implementation


ggleyzer (approved):
Reviewed a number of times together and tested with new Platform

PR #260, opened 2025-03-21
Enforce immutability of an immutable not const type

ggleyzer:
It's possible for a class or interface to be declared as immutable, for example:

    interface TextPosition
            extends immutable Hashable {
    }

If a class implements such an interface, the runtime must enforce a newly constructed object immutability before it becomes "visible".


cpurdy (approved):
Reviewed with Gene; approved.

PR #263, opened 2025-05-02
Ref and Tuple updates

cpurdy:
Updated BNF and Ecstasy compiler to support more natural Tuple syntax, and trailing commas in most list-oriented constructs. Updated core library interfaces, implementations, and documentation for Ref and Tuple.

### DIFF
--- a/javatools/src/main/java/org/xvm/compiler/ast/AssignmentStatement.java
+++ b/javatools/src/main/java/org/xvm/compiler/ast/AssignmentStatement.java
@@ -515,9 +515,8 @@ public class AssignmentStatement
         lvalueExpr = exprLeft;
         exprLeft.requireAssignable(ctxLValue, errs);
 
-        Expression     exprRight    = rvalue;
-        Expression     exprRightNew = null;
-        TypeConstant[] atypeRight   = null;     // used only to update the left type info
+        Expression     exprRight  = null;
+        TypeConstant[] atypeRight = null;     // used only to update the left type info
         switch (getCategory())
             {
             case Assign:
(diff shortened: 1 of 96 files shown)


javatools/src/main/java/org/xvm/compiler/ast/AssignmentStatement.java
```diff
@@ -650,9 +648,8 @@ else if (fConditional)
                     atypeReq[0] = pool().typeBoolean();
                     System.arraycopy(atypeLVals, 0, atypeReq, 1, cLVals);
 
-                    exprRightNew = exprRight.validateMulti(ctxRValue, atypeReq, errs);
-
-                    if (exprRightNew == null)
+                    exprRight = rvalue.validateMulti(ctxRValue, atypeReq, errs);
+                    if (exprRight == null)
                         {
                         return null;
```
ggleyzer, line 654: I'm not sure why we "return null" from here rather than "break". Returning here skips the "nodeLeft.resetLValueTypes(ctx);" call which happens on all other error paths

javatools/src/main/java/org/xvm/compiler/ast/ParenthesizedExpression.java
```diff
@@ -21,6 +29,138 @@ public ParenthesizedExpression(Expression expr, long lStartPos, long lEndPos)
...
+            TypeFit fitTuple = testTupleFit(ctx, typeRequired, true, null);
+            TypeFit fitValue = super.testFit(ctx, typeRequired, true, null);
+            if (fitTuple.betterThan(fitValue))
+                {
+                // this expression is actually a tuple literal, which means the delegated expression
+                // is the first and only element of the tuple
+                TypeConstant typeTuple = mightBeTuple(typeRequired);
+                assert typeTuple != null;
+                List<TypeConstant> listElements = typeTuple.getTupleParamTypes();
+                TypeConstant typeElementRequired = null;
```
ggleyzer, line 132: the value of `typeElementRequired` is computed, but never used

ggleyzer (commented):
Not sure why you decided to fix the archive; I wouldn't touch it - It's an archive after all :)

ggleyzer (approved):

javatools/src/main/java/org/xvm/compiler/ast/AssignmentStatement.java
cpurdy: True, but `TypeConstant[] atypeAll = exprRight.getTypes();` would be an automatic NPE ...

cpurdy:
@ggleyzer I pushed changes based on your feedback. Please review.

cpurdy:
Gene is traveling and texted me: "All looks good. There is no way for me to approve it again 😊"

PR #264, opened 2025-05-05
Fixed ops that needed "type" for comparisons and reformatted all Ops

cpurdy:
Updated source formatting for all Op classes

Reviewed OP_* design and closed TODOs related to missing type info


ggleyzer (approved):
Reviewed together; tested BAST verification. Looks good

PR #267, opened 2025-05-31
Java JIT project milestone: Container -1 (core/native type system) initializes.

cpurdy:
Added Linker, TypeSystem, Container, and Injector. Finished initial construction of the "Xvm" object, including building the native type system. Cleaned up Linker.


ggleyzer (approved):
Looks like a good start

PR #268, opened 2025-06-10
Fix an incorrect and-mask

mrcompator:
The code masked only a nibble instead of a byte which caused incorrect results.


ggleyzer (approved):
Good catch

PR #269, opened 2025-06-15
Introduce CooperativeCleanableReference

markandrewfalco:
This PR introduces `CooperativeCleanableReference` which is a `WeakReference` that runs a cleaner function after its referent becomes unreachable.


javatools_utils/src/main/java/org/xvm/util/CooperativeCloseableReference.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A {@link WeakReference} with a corresponding "cleaner" function which is run as part of {@link #close() closing} the
+ * reference. The reference is automatically closed after the referent becomes unreachable. The {@link #close} operation
+ * is idempotent and the cleaner will be invoked at most once even if {@link #close} is manually invoked.
+ *
+ * @param <V> the referent type
+ * @author falcom
```
ggleyzer: What is an expected use pattern? Would be nice to have it documented here

javatools_utils/src/main/java/org/xvm/util/CooperativeCloseableReference.java
```diff
@@ -0,0 +1,157 @@
...
+        KEEPALIVE.add(ref);
+        int cooperativity = CooperativeCloseableReference.cooperativity;
+        if (cooperativity != 0
+                && ThreadLocalRandom.current().nextInt(cooperativity) == 0
+                && COOPERATIVEITY.compareAndSet(cooperativity, 0))
+            {
+            try
+                {
+                cooperativity = clean()
+                        ? Integer.highestOneBit(Runtime.getRuntime().availableProcessors())
```
ggleyzer: I don't see any reason to compute this constant value every time

javatools_utils/src/main/java/org/xvm/util/CooperativeCloseableReference.java
```diff
@@ -0,0 +1,157 @@
...
+                    }
+
+                try
+                    {
+                    ref.close();
+                    }
+                catch (RuntimeException e)
+                    {
+                    LOGGER.log(Level.FINE, "ignoring exception during cooperative cleanup", e);
+                    Throwable parent = e.getCause();
```
ggleyzer: At the moment, we don't use Java logging (though maybe we should). I don't think this "fine-level" log message deserves the exception :)

markandrewfalco: I'd hate to eat the exception without logging, and rethrowing seems quite wrong as the thread which would receive it is almost certainly going to be unrelated to the code which caused it.

ggleyzer (approved):
Looks good

PR #271, opened 2025-06-25
Upgrade to Gradle 8.14.2; prepare to updating to JDK 24

lagergren:


ggleyzer:
I switched to this branch, then from the command line
- switched to java 24
- ./gradlew clean
- ./gradlew installDist -D org.xtclang.java.enablePreview=true --warning-mode none
Build completes with a number of warnings
********
- quit IDEA
- ran ./bin/purge-all-build-state.sh
- restart IDEA
- set JDK 24 with language level 24 Preview
- build project  fails with tons of errors that say:
   Kotlin: Plugin classpath entry points to a non-existent location: /Users/ggleyzer/.gradle/caches/8.9/generated-gradle-jars/gradle-api-8.9.jar

ggleyzer:
BTW, this branch is one commit behind

ggleyzer:
The setting that caused all those issues was the "Gradle VM", which needs to be in sync with the project VM. After resetting it, all seems fine. Can you please change the "warningAsError" default to "false"?

ggleyzer (approved):
Everything seems to be cleaner. Please change to 21
 org.xtclang.java.jdk=21
and merge. We will use 24 only in the jit branch for now

.github/workflows/xvm-verify-push.yml
```diff
@@ -80,14 +83,14 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: zulu
```
lagergren, line 85: You may want to create a separate change request just deleting xvm-verify-push.xml. ci.yml replaces it and does all it should do, now we are running a lot of unnecessary GitHub actions. Sorry bout that.

ggleyzer: I have already - it was a part of this PR https://github.com/xtclang/xvm/pull/272

PR #274, opened 2025-07-29
Fixed ci so that we use Java 21. Fixed foojay toolchain bootstrapping so gradlew always works, regardless of system JDK. Added multi platform Dockerfile for the XDK (including launcher build for e.g. Mac which is Linux/aarch64)

lagergren:
Fixed ci so that we use Java 21. Fixed foojay toolchain bootstrapping so gradlew always works, regardless of system JDK. Added multi platform Dockerfile for the XDK (including launcher build for e.g. Mac which is Linux/aarch64)

### DIFF
--- a/javatools/build.gradle.kts
+++ b/javatools/build.gradle.kts
@@ -62,13 +62,14 @@ sourceSets {
  * still being in the build output, as we have not cleaned.
  */
 val syncDependencies by tasks.registering(Sync::class) {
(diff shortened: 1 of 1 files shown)


lagergren:
What would be black magic, @ggleyzer ? I think it makes sense mostly. The issues are known, and the solutions aren't that hard to map against the issues?

PR #276, opened 2025-08-02
Optimised docker file and ci logic, fixed some issues with too long runs.

lagergren:


lagergren:
> assumption is that this one is smaller

It is smaller, but can be still be trimmed.  And the ci jobs should now go faster and treat docker artifacts as a separate thing to build pushed to master.

cpurdy (approved):
See: https://hub.docker.com/r/bellsoft/liberica-openjdk-alpine/

PR #277, opened 2025-08-03
Improve ci and docker builds.

lagergren:
Now better caching, and reuse of the build artifacts between jobs + we have a matrix for amd64/aarch64 containers because aarch64 linux is also free for GitHub actions these days, at least for public projects. 

The gradle tasks are slow and re-run. We also built the aarch64 container in qemu emulated mode which is terrible.


manualTests/src/main/x/runner.x
ggleyzer, line 1: This change makes absolutely no sense. The "loadAndRun" is a `Future` and all tasks run in parallel. Why would we want to reduce the parallelism?

manualTests/src/main/x/runner.x
```diff
@@ -32,6 +45,28 @@ module Runner {
         }
     }
 
+    Int getBatchSize() {
+        // Use conservative batch size that works well on both platforms
+        // TODO: Platform detection not working with @Inject("os.name")
```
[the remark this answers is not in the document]
ggleyzer: None of that makes any sense

manualTests/src/main/x/runner.x
```diff
@@ -32,6 +45,28 @@ module Runner {
...
+
+    void waitForBatchCompletion(Tuple<FutureVar, ConsoleBuffer>?[] batchResults) {
+        // Process all results in the batch as they complete
+        reportResults(batchResults, 0);
+        
+        // Wait for all futures in this batch to complete
+        for (Tuple<FutureVar, ConsoleBuffer>? result : batchResults) {
+            if (result != Null) {
+                try {
+                    result[0].get(); // Block until this future completes
```
ggleyzer: This is just plain horrible. Instead of using a non-blocking continuation paradigm, it blocks the fiber waiting for a result

ggleyzer (approved):
I rolled back runner.x changes. For the rest I rely on Marcus's knowledge and testing

PR #282, opened 2025-08-12
Fix intellij warning with resource path being moved to get the implicit.x/turtle.x bootstrap into the gradle builds.

lagergren:
(Also added an md (not word) version of the docx file for XTC so we can start working on describing the PROPER xtc grammar through AI by examining the doc and the lever)


ggleyzer (approved):
The rest looks good

PR #283, opened 2025-08-12
Added single source of truth for versions. All versions including API version, are now in a single source of truth: versions.properties in the root. 

lagergren:
(We will use this for releases as well, which is a manual process, that will build and publish a release on GitHub with a non SNAPSHOT version extension, and then bump the version to the next point-version + SNAPSHOT for further development, updating master, since what we will be working on after the release is the next release, and that is of course a SNAPSHOT release as long as it's unreleased)'

Also got around the issue with the hardcoded API versions that no dependency consumer knows anything about.

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/BuildInfo.java
+++ b/javatools/src/main/java/org/xvm/asm/BuildInfo.java
@@ -58,7 +58,7 @@ public final class BuildInfo
         }
 
     /**
-     * Get the API major version number.
+     * Get the XVM major version number.
      * Controls XTC file format compatibility.
      */
     public static int getXvmVersionMajor()
@@ -67,7 +67,7 @@ public final class BuildInfo
         }
 
     /**
-     * Get the API minor version number.
+     * Get the XVM minor version number.
      * Controls XTC file format compatibility.
      */
     public static int getXvmVersionMinor()


lagergren:
Here is a more structured response based on the results when I asked Claude Code to read this PR, study the code changes and explain the rationale.


  Thanks for the detailed questions! Let me clarify the motivation and address each concern:

  Big Picture Goal

  This PR creates a single source of truth for version management to enable automated release workflows. Currently,
   version information is scattered across multiple files, making release automation complex and error-prone.

  Release Workflow Context

  The unified system enables the planned release process where:
  1. Manual release builds use the non-SNAPSHOT version from version.properties
  2. Automated post-release scripts bump to next point-version + SNAPSHOT
  3. Only one file needs updating instead of hunting through multiple locations

  Addressing Specific Concerns

  "Runtime vs Compile-time": No architectural changes occurred. The API versions remain compile-time constants -
  they're just sourced from properties during static initialization instead of being hardcoded.
  BuildInfo.getApiVersionMajor() is resolved at build time, maintaining identical accessibility and scope.

  "Properties class verboten": Java Properties remains standard and non-deprecated for this use case. You might be
  thinking of deprecated XML properties methods. Alternative formats would require additional parsing dependencies
  - Properties integrates naturally with Maven/Gradle.

  "API version terminology": You're right about the XVM spec version vs API version distinction. If you believe
  version compatibility checking isn't needed for v0.x era, we can remove that part entirely. The core goal is
  centralized version management for build automation.

  "Moving version out of code": I converted the single-value VERSION file to multi-key version.properties to
  support comprehensive version metadata (XDK version, group ID, API versions). The version information remains in
  the same accessibility scope - just in a more structured format.

  The system consolidates what was scattered across VERSION, GROUP, and hardcoded constants, enabling the automated
   release workflow described in the PR while maintaining all existing behavior.

cpurdy (approved):

PR #287, opened 2025-08-18
Create an initial framework for the backend JIT compiler

ggleyzer:


build-logic/aggregator/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     `kotlin-dsl`
+    kotlin("jvm") version "2.1.20"
 }
```
lagergren: Get rid of the jvm plugin. It is unnecessary.

build-logic/common-plugins/common-plugins.gradle.kts
```diff
@@ -4,6 +4,7 @@
 
 plugins {
     `kotlin-dsl`
+    kotlin("jvm") version "2.1.20"
 }
```
lagergren: Get rid of the jvm plugin. It is unnecessary.

build-logic/common-plugins/src/main/kotlin/XdkDistribution.kt
```diff
@@ -124,7 +124,6 @@ class XdkDistribution(project: Project): XdkProjectBuildLogic(project) {
     companion object {
         const val DISTRIBUTION_TASK_GROUP = "distribution"
         const val JAVATOOLS_PREFIX_PATTERN = "**/javatools*"
-        const val JAVATOOLS_INSTALLATION_NAME : String = "javatools.jar"
         const val GPGKEY_FILENAME = "xtclang-gpgkey.asc"
```
lagergren: Master hasn't dropped JAVATOOLS_INSTALLATION_NAME, so put it back, even if it is not directly in use.

build-logic/settings-plugins/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     `kotlin-dsl`
+    kotlin("jvm") version "2.1.20"
 }
```
lagergren: Remove the jvm plugin.

gradle/libs.versions.toml
```diff
@@ -38,6 +39,7 @@ xdk-build-java = { id = "org.xtclang.build.java", version.ref = "xdk" }
 xdk-build-publish = { id = "org.xtclang.build.publish", version.ref = "xdk" }
 xdk-build-versioning = { id = "org.xtclang.build.xdk.versioning", version.ref = "xdk" }
 
+kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
 download = { id = "de.undercouch.download", version.ref = "download" }
```
lagergren: If no one refers to these kotlin and kotlin-jvm versions, remove them.

javatools_jitbridge/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+val xdkJavaToolsProvider by configurations.registering {
+    description = "Provider configuration for the javatools-jitbridge jar"
+// Note to ML: the build sill works without these
+//    isCanBeResolved = false
+//    isCanBeConsumed = true
+    outgoing.artifact(tasks.jar)
+    attributes {
+// Note to ML: the build sill works without this
+//        attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category.LIBRARY))
+        attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(XDK_ARTIFACT_NAME_JAVATOOLS_JAR))
```
lagergren: No it doesn't, it will build the XDK, but we are removing a consumable dependency and turning it to its default. If you are interested why you should leave these as is, you can ask claude. See: 

```
 ~/src/xvm  cpurdy/jit *12  claude                                                                                     ✔ 
╭───────────────────────────────────────────────────╮
│ ✻ Welcome to Claude Code!                         │
│                                                   │
│   /help for help, /status for your current setup  │
│                                                   │
│   cwd: /Users/marcus/src/xvm                      │
╰───────────────────────────────────────────────────╯

> While Cameron has commented out various lines in the xdkJavaToolsProcider as defined by 
  javatools_jitbridge/build.gradle.kts - is it a correct thing to do, especially considering how the distribution 
  build kts, the XdkDistribution and the plugin expects their javatools to be provided?

⏺ I'll examine the commented lines in the xdkJavaToolsProvider and analyze how it affects the distribution build,
  XdkDistribution, and plugin expectations.
  ⎿  Interrupted by user

> While Cameron has commented out various lines in the xdkJavaToolsProcider as defined by 
    javatools_jitbridge/build.gradle.kts - is it a correct thing to do, especially considering how the distribution 
    build kts, the XdkDistribution and the plugin expects their javatools to be provided?  I am talking about this 
  configuration in the new project javatools_jitbridge/build.gradle.kts // Provider configuration to make this jar 
  available to the XDK distribution
  val xdkJavaToolsProvider by configurations.registering {
      description = "Provider configuration for the javatools-jitbridge jar"
  // Note to ML: the build sill works without these
  //    isCanBeResolved = false
  //    isCanBeConsumed = true
      outgoing.artifact(tasks.jar)
      attributes {
  // Note to ML: the build sill works without this
  //        attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category.LIBRARY))
          attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(XDK_ARTIFACT_NAME_JAVATOOLS_JAR))
      }
  }

⏺ Update Todos
  ⎿  ☐ Examine the commented lines in javatools_jitbridge/build.gradle.kts
     ☐ Check how XdkDistribution expects javatools to be provided
     ☐ Analyze distribution build.gradle.kts dependencies
     ☐ Review plugin expectations for javatools provider configuration

⏺ Update Todos
  ⎿  ☐ Examine the commented lines in javatools_jitbridge/build.gradle.kts
     ☐ Check how XdkDistribution expects javatools to be provided
     ☐ Analyze distribution build.gradle.kts dependencies
     ☐ Review plugin expectations for javatools provider configuration

⏺ Read(javatools_jitbridge/build.gradle.kts)
  ⎿  Read 29 lines (ctrl+r to expand)

⏺ Search(pattern: "xdkJavaToolsProvider|javatools.*provider", glob: "**/*.gradle.kts")
  ⎿  Found 2 files (ctrl+r to expand)

⏺ Search(pattern: "XdkDistribution", glob: "**/*.gradle.kts")
  ⎿  Found 3 files (ctrl+r to expand)

⏺ Update Todos
  ⎿  ☒ Examine the commented lines in javatools_jitbridge/build.gradle.kts
     ☐ Check how XdkDistribution expects javatools to be provided
     ☐ Analyze distribution build.gradle.kts dependencies
     ☐ Review plugin expectations for javatools provider configuration

⏺ Read(xdk/build.gradle.kts)
  ⎿  Read 279 lines (ctrl+r to expand)

⏺ Read(javatools/build.gradle.kts)
  ⎿  Read 232 lines (ctrl+r to expand)

⏺ Update Todos
  ⎿  ☒ Examine the commented lines in javatools_jitbridge/build.gradle.kts
     ☒ Check how XdkDistribution expects javatools to be provided
     ☐ Analyze distribution build.gradle.kts dependencies
     ☐ Review plugin expectations for javatools provider configuration

⏺ Search(pattern: "isCanBeResolved|isCanBeConsumed", glob: "**/*.gradle.kts", output_mode: "content")
  ⎿  Found 76 lines (ctrl+r to expand)

⏺ Update Todos
  ⎿  ☒ Examine the commented lines in javatools_jitbridge/build.gradle.kts
     ☒ Check how XdkDistribution expects javatools to be provided
     ☒ Analyze distribution build.gradle.kts dependencies
     ☐ Review plugin expectations for javatools provider configuration

⏺ Read(plugin/build.gradle.kts)
  ⎿  Read 50 lines (ctrl+r to expand)

⏺ Update Todos
  ⎿  ☒ Examine the commented lines in javatools_jitbridge/build.gradle.kts
     ☒ Check how XdkDistribution expects javatools to be provided
     ☒ Analyze distribution build.gradle.kts dependencies
     ☒ Review plugin expectations for javatools provider configuration

⏺ No, commenting out those configuration properties is not correct.

  The commented lines in javatools_jitbridge/build.gradle.kts:16-17 are essential for proper Gradle configuration
  dependency resolution:

  //    isCanBeResolved = false  // Should NOT be commented
  //    isCanBeConsumed = true   // Should NOT be commented  

  Why these properties matter:

  1. Configuration Role Definition: These properties explicitly define the configuration's role:
    - isCanBeResolved = false: This is a provider configuration, not meant to resolve dependencies
    - isCanBeConsumed = true: This configuration provides artifacts to other projects
  2. Consistency with Pattern: Every other provider configuration in the codebase correctly sets these:
    - javatools/build.gradle.kts:19-20 ✅
    - javatools_launcher/build.gradle.kts:40-41 ✅
    - lib_ecstasy/build.gradle.kts:38-39 ✅
  3. Consumer Expectations: The plugin at plugin/build.gradle.kts:27-28 and XDK distribution at
  xdk/build.gradle.kts:46-47 both depend on these provider configurations being properly configured.

  The build may work without them due to Gradle's default fallback behavior, but this creates:
  - Ambiguous configuration semantics
  - Potential resolution issues in complex dependency scenarios
  - Inconsistency with established patterns

  Recommendation: Uncomment lines 16-17 in javatools_jitbridge/build.gradle.kts to maintain consistency and explicit
  configuration semantics.
```

plugin/gradle.properties
```diff
@@ -1,4 +1,4 @@
 # The plugin must be compiled to be compatible with the Gradle daemon JVM
```
lagergren: Just delete this file. The aggregated build will take the closest jdk defintion found, and this case it is at the top level in xdk.properties. This is redundant and it's probably wrong to keep it if it has a different version than the xdk.properties jdk version.

xdk/src/main/resources/xdk/bin/xcc.cfg
```diff
@@ -0,0 +1 @@
+opts = -ea --enable-preview
```
lagergren, line 1: Here is yet another place where we add flags that are a bit non-DRY, since the java build already sets these flags for the java launch. Of course if we keep these native binaries whose only job is basically to do an exec of java 24, then they need to be here as long as javatools uses previous flags. I am starting to dislike the huge amount of complexity some native binares contribute. They also force us to build different launchers and test different configs in the CI and build different releases, even though the launcher is really the only thing that is platform independent here.

lagergren: I would probably just ditch these mysterious files and hard code these two arguments in launcher.c  It is simpler, and this is really not more abstract, just more stuff to double check and keep track of IMHO

ggleyzer: This is not for the build, but the stand-alone compilation or running

build-logic/common-plugins/common-plugins.gradle.kts
```diff
@@ -4,6 +4,7 @@
 
 plugins {
```
lagergren: If you plan to go full Java 24, you need to update the Java version property in ci.yml from 21 to 24. And before you merge, you should manually try to dispatch the GitHub ci workflow so that you know it is safe. We don't want it to fail in master and crash all our artifacts.  Run:  │   gh workflow run ci.yml --ref cpurdy/jit --field always_build_docker_image=true                                       from your dev machine and study the results. Since this is not master, there is no risk that we overwrite master docker images, since they will be tagged differently, with the branch name and the commit, and not with xvm:latest, so this can't break anything. 


Note that the branch input should be set to the branch name or it will default to master.

ggleyzer: I would really like you do direct that process

xdk/src/main/resources/xdk/bin/xcc.cfg
cpurdy: The .cfg files are a temporary requirement until these features are incorporated into the JDK as official features. Changing the launcher source code to do a temporary work-around for this is the wrong choice, although it would obviously work. Instead, those .cfg files should be retained only as long as the JDK we are targeting requires them, and when no longer required, we can remove the .cfg files.

lagergren: It's not as much a workaround as something we need to put there until we use a jdk where the previous features are not preview features anymore. It is completely the same thing as having the enable preview flag set in the build logic java convention plugin for gradle. It is exactly as hacky or non hacky as that. Since we now have a build that breaks pre-24 and we need to use preview features to build, I don't think its a valid argument that we need three config files for the native launcher for preview overrides. Since we have to add enable preview to the java build, also a temporary thing, we can just save us complexity of another weird command line config (especially in triplicate) and hard code it.   I would argue that having a native program exec a system java process, is hella-brittle anyway, so the entire launcher mechanism is a hack until we have some other way to launch than invoking a jvm through system calls.  I would just put the flags in the launcher.c file, add a # TODO comment so it shows up as a problem to be addressed in intellij and then file a GitHub issue to track this. 

( Btw - we should really have done the with byte buddy instead, as I suggested, to avoid the entire java 24 class library horror. Besides being preview, the API is weak and immature.  It would save us a lot of a mess and is maintained, deployed everywhere, an and fully trusted by the industry.  Very likely there are already class file api migration paths for the things that can be supported, so I am in favour of that)

cpurdy (approved):
All tests are passing. Only remaining concerns are (i) unavoidable requirement for JDK 24 and (ii) some Java and Gradle warnings that we're still working on eliminating

PR #290, opened 2025-08-21
[DRAFT/POC] Very simple experiment to build all native binaries sans C compiler, for all platforms, as a manually dispatchable GitHub action using Graal native-builder.

lagergren:
Zero installation zero source C binary launchers as downloadable, usable local build artifacts workflow, built by the Graal native-builder for JDK 24. Completely self contained, and will not try to run "java" from the system path, making them our most expensive syntactic sugar that could just have been a Java command line or a docker run url:/xtc. 

Binaries contain exactly as much Java as XTC needs to run, using Compiler and Runner and javatools.jar as the starting point for the xtc and xec launchers. They require some system libs, but no Java installation, and they successfully work with our Java 24 code. 

Exit values in the OS turn into whatever Java's System.exit in our code was called with, all transparently. They are guaranteed to be self contained on the four native platforms in the build matrix for the GitHub action. There are probably glitches somewhere, but this seems to work remarkably well for most my basic use cases I've tried.  Also, since it's 2025 no one needs to run this on a local machine, or find windows boxes or whatever. No one even needs to install Graal, except for the GitHub action that exists already.  

> xec <something that results in a system exit 1>
> echo $? 
1

On all platforms. 

Manually triggering the build action will produce four GitHub artifacts that we can pull into our normal local gradle build. 

  - xvm-tools-linux-amd64.zip
  - xvm-tools-windows-amd64.zip
  - xvm-tools-macos-arm64.zip
  - xvm-tools-macos-amd64.zip

  Each contains the xcc and xec native executables for that platform. They are ~40-50 MB each. 

  Each job takes about 5-10 minutes to:
  1. Set up GraalVM 24
  2. Build the JAR
  3. Compile native executables
  4. Test they work (--version)
  5. Upload artifacts

You can run the build flow for your on your local machine with the gradlew javatools:nativeCompile and javatools:nativeRunnerCompile tasks (not part of any normal build), as long as you have graalvm for JDK 24 installed, which you can do with e.g. sdkman. A build that pulls these artefacts in would of course not have to run with any Graal JVM.


lagergren:
If you want to play with the GitHub build flow, and do manual dispatches to create Github action launcher zip artifacts, this must go into master first, or the build flow will not be visible.

PR #291, opened 2025-08-21
Updated Native Launchers, updated macos launchers, updated Windows la…

cpurdy:
…uncher, introduced launcher tests, updated Linux ARM 64 version of the native launcher, updated Linux AMD 64 version of the native launcher


cpurdy:
Built and tested on Windows, macOS, Debian (both ARM and x86/64).

javatools_launcher/src/main/c/os_windows.c
```diff
@@ -61,12 +61,12 @@ const char* escapePath(const char* path) {
     return result;
 }
```
lagergren, line 63: So you created a macos_launcher_amd64 for intel Macs - did you have to do anything to build an x86 build on mac, or did it just pick the right binaries for xdk:installWithLaunchersDist?  (I mean they are the same file on all Macs, but if amd64 x macOS is a combo we are expected to have launchers on, then you should be able to run xdk:installWithLaunchersDist -D/-Pos.arch= ... to see that it picks the right files. If not, you probably have to add a valid configuration combo in XdkDistribution.kt or something. It is still all good. 

Can we move the launcher binary source code and build logic to build them to some other repo, perhaps? Then all xvm builds will do is check if a launcher binary exists, wherever we choose to put it - likely on GitHub as an artifact - and just build it in.

manualTests/src/main/x/xrun.cmd
ggleyzer, line 1: I think you meant to call TestRun, not TestSimple

manualTests/src/main/x/xrun.sh
ggleyzer, line 1: Ditto

javatools_launcher/src/main/c/os_windows.c
cpurdy: The macos_launcher_amd64 and _arm64 are the same file, because we build a macos_launcher that includes both hardware platforms by default. This is a supported features of macOS, so that software can have a single install regardless of whether it's targeting ARM or x86, and yet still be "native" for each of those.

manualTests/src/main/x/xrun.sh
cpurdy: Weird. I think that I had made that change locally but somehow lost it :( ... let me double check.

ggleyzer (approved):

PR #296, opened 2025-08-28
Fix isA() implementation for immutable types

ggleyzer:
A number of DBObject derived types require `immutable Const` parameters, for example:

    `interface DBMap<Key extends immutable Const, Value extends immutable Const>`
    
However, the compiler would allow the following declaration:

    `mixin MyTable into DBMap<String, Int[]>`

instead of the correct one:

    `mixin MyTable into DBMap<String, immutable Int[]>`

This would result in a run-time exception.

This change fixes that compiler deficiency. An additional benefit is a compiler check for the actual value types for DBMap method calls, such as `put(key, value)`, forcing the value to be known an `immutable Const` at compile time

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/Component.java
+++ b/javatools/src/main/java/org/xvm/asm/Component.java
@@ -3276,11 +3276,12 @@ public abstract class Component
 
                         if (typeContrib.isAccessSpecified())
                             {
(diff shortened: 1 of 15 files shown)


ggleyzer:
Reviewed on-line with Cam

PR #299, opened 2025-09-02
POC of Canary based GCs

markandrewfalco:
Prototype canary based GC, written in the form of "tests", as well as diagrams of their basic architecture.


ggleyzer (approved):
Let's get it in

PR #301, opened 2025-09-04
Fixing java toolchain bug, still propagating extra java command line flags into the plugin build.

lagergren:


lagergren:
Git issues went away too - tidied it up a bit and will test the image builder pipeline

PR #335, opened 2025-11-05
Add programmatic javatools API access to plugin

lagergren:
## Summary

This PR enables the Gradle plugin to call XTC compiler and tooling methods programmatically, eliminating the need for reflection and subprocess spawning. The plugin can now directly invoke `Compiler.launch()`, `Runner.launch()`, and other javatools APIs as if they were part of the plugin's own codebase.

### 1. Compile-only dependency on javatools
```kotlin
// plugin/build.gradle.kts
dependencies {
    compileOnly(libs.javatools)
}
```
- Provides type information at compile time
- Not bundled with the plugin (kept lightweight)
- Resolved at runtime from XDK distribution

### 2. Runtime classloader management
- `XtcJavaToolsRuntime.ensureJavaToolsInClasspath()` loads javatools.jar dynamically
- Creates isolated classloader with proper parent delegation
- Handles both XDK builds and installed distributions

### 3. Direct method invocation
```java
// No reflection needed!
Compiler.launch(args);  // Direct API call
```
- Type-safe compilation
- Full IDE support (autocomplete, refactoring)
- Easier debugging and maintenance

### 4. Launcher flexibility with fork flag
- `fork=false`: In-process execution (fast, shared JVM)
- `fork=true`: Separate process (isolation, custom JVM args)
- `useNativeLauncher=true`: Native binary launcher

### Developer Experience
- **IDE Integration**: Full autocomplete and type checking for javatools APIs
- **Debugging**: Step through compiler code directly in debugger
- **Refactoring**: Safe renames and refactorings across plugin and javatools

### Performance
- **In-process execution**: No JVM startup overhead for quick compilations
- **Flexible forking**: Can still fork for isolation when needed
- **Native launcher**: Fast C++ implementation for production builds

### Maintainability
- **No reflection**: Compile-time type safety catches errors early
- **Clear API surface**: Explicit dependencies make architecture obvious
- **Easier testing**: Direct method calls simplify unit testing

## Architecture

```
┌─────────────────────────────────────────┐
│   Gradle Plugin (compileOnly)           │
│   - Knows javatools types at build time │
│   - No javatools.jar bundled            │
└────────────────┬────────────────────────┘
                 │
                 ▼
┌─────────────────────────────────────────┐
│   XtcJavaToolsRuntime                   │
│   - Loads javatools.jar at runtime      │
│   - Creates isolated classloader        │
└────────────────┬────────────────────────┘
                 │
                 ▼
┌─────────────────────────────────────────┐
│   JavaClasspathLauncher                 │
│   - fork=false: Compiler.launch(args)   │
│   - fork=true: ProcessBuilder           │
└─────────────────────────────────────────┘
```

## Testing

- ✅ All existing tests pass
- ✅ Configuration cache compatibility verified
- ✅ Both fork modes tested (in-process and separate process)
- ✅ Native launcher compatibility maintained

## Migration Path

This change is backward compatible. Users can:
- Continue using `fork=true` for existing behavior
- Switch to `fork=false` for improved performance
- Use `useNativeLauncher=true` for native binaries

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTViewToBit.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTViewToBit.java
@@ -105,18 +105,4 @@ public class xRTViewToBit
         throw new UnsupportedOperationException("RTViewToBitFrom" + typeElement.getValueString());
     }
 
-    /**
-     * Clear the static VIEWS cache to allow reentrancy.
-     * This is a workaround for the reentrancy bug where TypeConstant keys
-     * from different ConstantPools don't match due to object identity comparison.
-     * <p>
-     * TODO: We should keep track of reentrant state that may be cleared through something
-     * like a Reentrency interface instead, or something explicit like that. Or just
-     * handle static caches and similar global tables in ome specific place or module-
-     * <p<
-     * This should be called between Runner.launch() invocations in the same JVM.
-     */
-    public static void clearViewsCache() {
-        VIEWS = null;
-    }
 }


javatools/src/main/java/org/xvm/tool/Runner.java
```diff
@@ -322,7 +318,10 @@ protected void process() {
 
                 connector.invoke0(method, asArg);
 
-                System.exit(connector.join());
+                int exitCode = connector.join();
```
ggleyzer: The intent of this System.exit() was to communicate to the caller process a value returned by the Ecstasy method. Please see the conversation on ecstasy-questions channel on 8/20/25:
https://discord.com/channels/837372142576205885/1406009322307190834/1407755170787364995

As you can see, the System,exit() here is called upon a normal termination, not an exceptional one. Your change breaks that contract.

javatools/src/main/java/org/xvm/tool/Runner.java
ggleyzer, line 1: I'm not sure I understand the point of this refactoring. As it stands today, outside of the Runner.process() method that uses it for **normal** termination, the only places we call System.exit() are "main{}" methods, that are not supposed to be called programmatically. Therefore, as long as you do `Compiler.launch(...)` instead of `Compiler.main(...)`, no `System.exit()` will occur

ggleyzer (dismissed):

javatools/src/main/java/org/xvm/tool/Runner.java
lagergren: I am not finished yet. I will mark it for review when it satisfactorily does what I was intending it to do.

javatools/src/main/java/org/xvm/tool/Runner.java
lagergren: But the whole point of using launch instead of main is that the plugin is guaranteed that no one will System.exit on it when i calls launch.   If you want someone to give you exit codes then you must use main() or some other way of handling system exit.   Since we have no in-process launcher in any configuration in master, although there is one (the buildThread launcher stuff) - that has been broken ever since this came in.

ggleyzer (changes requested):
I looked again at the `xRTViewToBit` code I realized that I was too hasty with my re-entrancy assessment. The `initNative()` actually reloads the `VIEW` map, which happens during the `NativeContainer` initialization by`loadNativeTemplates`. Therefore, the `clearViewsCache` method is unnecessary. 

Moreover, we could simply remove the "static" from `VIEWS` declaration.

javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -1395,6 +1395,7 @@ public void validateModuleOutput(File file, boolean fMulti) {
         if (dir != null && !dir.exists()) {
             log(Severity.INFO, "Creating directory " + dir);
             // ignore any errors here; errors would end up being reported further down
+            //noinspection ResultOfMethodCallIgnored
```
ggleyzer, line 1398: What is this added comment about? What does it say?

lagergren: It removes yellow.  You can easily find out yourself, for example by this:  

```
> What does the //noinspection line in @Launcher.java mean ?   if (dir != null && !dir.exists()) {
            log(Severity.INFO, "Creating directory " + dir);
            // ignore any errors here; errors would end up being reported further down
            //noinspection ResultOfMethodCallIgnored 

⏺ The //noinspection ResultOfMethodCallIgnored comment is an IntelliJ IDEA inspection suppression directive.

  What it does:

  It tells IntelliJ IDEA (and other JetBrains IDEs) to suppress the "Result of method call ignored" warning for the next line of code.

  Why it's needed here:

  The code following this comment likely calls a method whose return value should normally be checked but is intentionally ignored. For example:

  dir.mkdirs();  // Returns boolean but we're ignoring it

  The mkdirs() method returns a boolean indicating success/failure, but the comment explains the reasoning: "errors would end up being reported further down" - meaning if directory
  creation fails, a subsequent operation will catch and report the error anyway, so there's no need to check the return value here.

  Alternative syntax:

  You might also see:
  - @SuppressWarnings("ResultOfMethodCallIgnored") - Java annotation (broader scope)
  - //noinspection - IntelliJ-specific comment (line-level suppression)

  The //noinspection approach is preferred when you want to suppress a warning for just one specific line while keeping the inspection active for the rest of the file.

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

lagergren:
I removed the clear method.

javatools/src/main/java/org/xvm/tool/Launcher.java
ggleyzer: I hate to be picky, bun not everyone is using IntelliJ and more commonly, not everyone has the same IntelliJ inspections as you do.

PR #337, opened 2025-11-07
Improve the Launcher API

ggleyzer:
Some time ago Michael and Matt (@Compator) made a small feature request to return a result of an Ecstasy module execution via "xec" command back to the external caller via the `System.exit()` API. It was implemented [by this commit](https://github.com/xtclang/xvm/commit/ec73bacb8850bf551971327f6784b94d8dc2edf4).

Marcus pointed out that the implementation was incorrect, preventing a programmatic use of the `Runner` API.

This change corrects the implementation, making sure that **the only time** the `System.exit()` is used when a tool is invoked as an external Java command - via the `public static main(String[])` method.

### DIFF
--- a/javatools/src/main/java/org/xvm/tool/Compiler.java
+++ b/javatools/src/main/java/org/xvm/tool/Compiler.java
@@ -124,7 +124,7 @@ public class Compiler
      */
     public static int launch(String[] asArg) throws LauncherException {
         return new Compiler(asArg).run();
-}
+    }
 
     /**
      * Compiler constructor.
--- a/javatools/src/main/java/org/xvm/tool/Ecstasy.java
+++ b/javatools/src/main/java/org/xvm/tool/Ecstasy.java
@@ -18,14 +18,27 @@ public class Ecstasy
      */
     public static void main(String[] asArg) {
         try {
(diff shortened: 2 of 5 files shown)


lagergren:
> Two issues:
> 
> 1. (Minor) TestSimple changes were unnecessary, because the manualtests directory already contains TestRun.x and two supporting shell scripts xrun.cmd and xrun.sh for Windows and therestoftheuniverse.
> 2. (Major) Launcher is the JAR's entry point, so having different behavior for Launcher vs (eg) Compiler is a certifiably bad thing. Instead, why can't we just add a command e.g. "test" or "ci-test" (whatever you want) to the Launcher to test the success of the Launcher build? Alternatively, we should have explicit non-main() entry points for each command, including Launcher, that are identical in functionality **other than calling exit()**.

1) There is no way I will introduce a dependency to another composite project's build files in. order. If it happens to be some identical lines of code in a resource directory, that is fine. The plugin functionality is self contained. The plugin is a separate project in a composite build, and the plugin definitely knows zero about manual test. This is the correct way to do it. 

2) I am certainly going to use the launcher as entry point. I have not requested this test for review yet.

cpurdy (approved):
Looks fine.

PR #338, opened 2025-11-08
Fix constructing a JSON ObjectInputStream using a NestedParser.

thegridman:


lib_json/src/main/x/json/ObjectInputStream.x
```diff
@@ -33,6 +30,23 @@ class ObjectInputStream(Schema schema, Parser parser)
         construct ObjectInputStream(schema, new Parser(lexer.ensureMarkable()));
     }
 
+    /**
+     * Construct an ObjectInputStream from a [Parser].
+     *
+     * If the `Parser` is a `NestedParser` the `ObjectInputStream` will only
```
ggleyzer: Across our code base the editor code style is set to have "hard wrap" at 100 columns. Could you please set up your IDEA code style accordingly. With those settings, this doc would look like this:

     * If the `Parser` is a `NestedParser` the `ObjectInputStream` will only read the tokens for
     * the JSON Doc currently pointed to by the `Parser`.

ggleyzer (approved):
Looks good

PR #344, opened 2025-12-05
Add XUnit test framework

thegridman:
### The `lib_xunit_engine` module
- This module discovers test methods annotated with `@Test` and executes them.
- To run tests application developers do not need to import the `lib_xunit_engine` module, this module is used internally by the `xtc test` CLI command.
- If the module being tested also imports the XUnit API (lib_xunit) then the engine also supports processing all of the XUnit API extensions.

### The `lib_xunit` module 
- This module contains the XUnit API that application developers can import to write more functionally rich tests than just `@Test` annotated methods. For example before and after test methods, parameterised tests, etc.

### The `xtc` CLI command

The `xtc` CLI command has been enhanced to support the following:
- `xtc build` equivalent of `xcc`
- `xtc run` equivalent of `xec`
- `xtc test` executes test in a module

### Gradle

The `testXtc` task in Gradle can be used to execute tests.

### Other Changes
- Support for injectables (the `-I` parameter on the `xec` command) has been enhanced to support both `String` and `String[]` types as well as conversion between `String` and `Destringable` types. There is no support for `Destringable[]` but this could be added.
- If an @Inject annotated element is nullable an exception is no longer thrown if the injection is missing.

### DIFF
--- a/lib_xunit/src/main/x/xunit/extensions.x
+++ b/lib_xunit/src/main/x/xunit/extensions.x
@@ -26,4 +26,37 @@ package extensions {
          */
         ExtensionProvider[] getExtensionProviders() = [];
     }
+
+    /**
+     * Returns the directory to for any files specific for the current test.
+     *
+     * @param root  the root directory to place the test directory under
+     *
+     * @return the directory to for any files specific for the current test
+     */
+    static Directory testDirectoryFor(Directory         root,
+                                      Class?            testClass,
+                                      MethodOrFunction? test = Null) {
+        Directory dir = root;
+        if (testClass.is(Class)) {
+            String path = testClass.path;
+            Int    size = path.size;
+            if (path[size - 1] == ':') {
+                // this is a Module
+                dir = dir.dirFor(testClass.name);
+            } else {
+                assert Int colon := path.indexOf(':');
+                String packages   = path[colon >..< path.size];
+                for (String name : packages.split('.')) {
+                    dir = dir.dirFor(name);
+                }
+                if (test.is(Test)) {
+                    dir = dir.dirFor(test.name);
+                }
+            }
+        }
+        dir.ensure();
+        return dir;
+    }
+
 }
--- a/lib_xunit_engine/src/main/x/xunit_engine/console/ConsoleExecutionListener.x
+++ b/lib_xunit_engine/src/main/x/xunit_engine/console/ConsoleExecutionListener.x
@@ -1,13 +1,22 @@
-import xunit_engine.Model;
-import xunit_engine.UniqueId;
-
+/**
+ * An XUnit ExecutionListener that prints test execution results to the console.
+ */
 service ConsoleExecutionListener
         implements ExecutionListener {
 
+    /**
+     * The console to print results to.
+     */
     @Inject Console console;
 
+    /**
+     * A map of `UniqueId`s to `Result`s.
+     */
     private Map<UniqueId, Result> results = new HashMap();
 
+    /**
+     * Whether all tests have succeeded.
+     */
     Boolean success = True;
 
     /**
--- a/lib_xunit_engine/src/main/x/xunit_engine/executor/ContainerExecutionLifecycle.x
+++ b/lib_xunit_engine/src/main/x/xunit_engine/executor/ContainerExecutionLifecycle.x
@@ -30,10 +30,17 @@ const ContainerExecutionLifecycle<ModelType extends ContainerModel>(ModelType mo
     	        extensions.add(extension, ep);
             }
         }
-	    return super(context.asBuilder(this.model)
+
+        EngineExecutionContext.Builder builder = context.asBuilder(this.model)
                 .withTestClass(model.testClass)
-                .withTestMethod(Null)
-                .build(), extensions);
+                .withTestMethod(Null);
+
+//        UniqueId id = model.uniqueId;
+//        if (id.type == Module) {
+//            builder.withModule(id.value);
+//        }
+
+	    return super(builder.build(), extensions);
 	    }
 
     @Override
--- a/lib_xunit/README.md
+++ b/lib_xunit/README.md
@@ -1,0 +1,220 @@
+# XUnit - The Ecstasy Testing Framework
+
+XUnit is a testing framework that allows developers to write unit tests in their Ecsatsy projects.
+XUnit is based on the popular Java JUnit test framework.
+
+## Writing Tests
+
+A test in XUnit is a method or function annotated with `@Test`.
+The `@Test` annotation is part of the Ecstasy core library, so you don't need to import anything 
+to write simple tests.
+
+Test methods and functions can be written inside modules, packages and classes.
+
+```ecstasy
+module MyApplication {
+
+    @Test
+    void testRange() {
+        Range<Int> range = 10..100;
+        assert range.covers(50..60);
+    }
+}
+```
+
+To write more feature-rich tests using the features of XUnit, the XUnit module should be 
+imported into the module to be tested.
+
+```ecstasy
+module MyApplication {
+
+    package xunit import xunit.xtclang.org;
+
+}
+```
+
+Test methods cannot be placed inside none-static inner classes or inside special classes such as 
+enums.
+
+### Basic Test Annotations
+
+The XUnit framework contains a number of annotations that can be used to customize the behavior of 
+tests. These annotations are part of the `xunit.annotations` package.
+
+- `@AfterAll` is used to annotate a function that will be executed after all tests in the 
+functions containing class and any subclasses have executed. If the function is in a module or 
+package, then the function will execute after all tests in the entire module or package have 
+executed. The annotated function will execute regardless of whether any tests have failed or not.
+
+- `@AfterEach` is used to annotate a method or function that will be executed after each test in the 
+containing class. If the method or function is in a module or package and is static, then the 
+method or function will execute after each test in the entire module or package has executed. 
+The annotated function will execute regardless of whether any tests have failed or not.
+
+- `@BeforeAll` is used to annotate a function that will be executed before all tests in the 
+containing class and any subclasses have executed. If the function is in a module or package, 
+then the function will execute after all tests in the entire module or package have executed. 
+If the annotated function throws an exception, no further "before" processing will be executed 
+and no tests will be executed for the containing class or any subclasses. Any "after all" 
+processing will still be executed.
+
+- `@BeforeEach` is used to annotate a method or function that will be executed before each test in 
+the containing class is executed. If the method or function is in a module or package and is 
+static, then the method or function will execute before each test in the entire module or 
+package is executed. If the annotated method or function throws an exception, no further "before" 
+processing will be executed and no tests will be executed for the containing class or any 
+subclasses. Any "after" processing will still be executed.
+
+- `@Disabled` is used to annotate a test method or function that should be ignored by the test 
+runner. The annotation takes a single string argument that can be used to provide a reason for 
+ignoring the test, which will be displayed in the test output.
+
+- `@DisplayName` is used to annotate a test method or function to provide a custom name for the 
+test that will be displayed in any test output.
+
+### Test Fixture Lifecycle
+
+By default, XUnit will create a new instance of a test class before executing each test in the 
+class. So each test method in a class will be executed in a separate instance of the class. It 
+is important to be aware of this behavior when writing tests that rely on any shared state. 
+Ideally, relying on non-static shared state should be avoided in tests.
+
+For tests that really do require shared state in the test class instance, the `@TestFixture` 
+annotation can be used on the test class to specify the lifecycle of the test class instance.
+The `@TestFixture` annotation takes a single parameter, which is an enum value that specifies the 
+lifecycle of the test class instance.
+
+- `@TestFixture(EveryTest)` this is the default behaviour where a new instance of the class is 
+  created for every test method.
+- `@TestFixture(Singleton)` a single instance of the class is created and all test methods in 
+  the class will use that instance.
+
+Ecstasy modules and packages are singletons, so any tests in a module or package will use the same 
+instance of the module or package.  
+
+
+### Assertions
+
+Ecstasy's `assert` keyword already offers excellent assertion support so there is no real need to use 
+any alternative assertion library.
+
+#### Asserting Exceptions
+
+One place where XUnit offers more flexibility is when asserting that some code should throw an 
+exception. The `xunit.assertions.assertThrows` function can be used to assert that a block of code 
+throws a specific type of exception. The function returns the exception that was thrown, to 
+allow further assertions to be performed on it.
+
+There are two forms of the `assertThrows` function:
+
+- `void assertThrows(Type, Function)` which just asserts that the function throws an exception of 
+the specified type.
+- `<E extends Exception> E assertThrows(Function)` which asserts the function throws an 
+  excpetion of type `E` and returns the thrown exception. 
+
+For example, if an application has a const that should always be a positive number between 1 and 
+100, we can test that the const throws an exception when the number is outside the valid range:
+
+```ecstasy
+module MyApplication {
+
+    const NumberHolder(Int number) {
+        assert() {
+            assert:arg number > 0 && number < 100 as "Number must be between 1 and 99";
+        }
+    }
+
+    @Test
+    void shouldNotCreateHolderWithNegativeNumber() {
+        assertThrows(IllegalArgument, () -> new NumberHolder(-1));
+    }
+}
+```
+
+If we wanted to assert the message of the exception, we could use the second form of the 
+`assertThrows` function:
+
+```ecstasy
+    @Test
+    void shouldNotCreateHolderWithNegativeNumber() {
+        IllegalArgument thrown = assertThrows(() -> new NumberHolder(-1));
+        assert thrown.message == "Number must be between 1 and 99";
+    }
+```
+
+## Running Tests
+
+XUnit tests can be run from the command line using the `xtc test` command.
+
+The `xtc test` command take a number of options that can be used to control the behavior of the 
+test runner. The options are the same as when running an Ecstasy module, with the additions of 
+some test-specific options.
+
+```
+  -d, --deduce                     Automatically deduce locations when possible
+  -h, --help                       Display this help message
+  -L <path>                        Module path (can be specified multiple times, or use : separator)
+  -v, --verbose                    Enable verbose logging and messages
+  --version                        Display the Ecstasy runtime version
+  --no-recompile                   Disable automatic compilation
+  -o <file>                        If compilation is necessary, the file or directory to write compiler output to
+  -I, --inject <name=value>        Specifies name/value pairs for injection; format is 'name=value'
+  -c, --test-class <class>         the fully qualified name of a class to execute tests in
+  -g, --test-group <group>         only execute tests with the specified @Test annotation group
+  -p, --test-package <package>     the name of a package to execute tests in
+  -t, --test-method <method>       the fully qualified name of a test method to execute
+```
+
+For example, if we have a module named `my-app.xtc` that contains test, the tests can be executed 
+using the following command:
+
+```shell
+xtc test my-app.xtc
+```
+
+If the module file is not in the current directory, then just like with running the application, 
+we can specify a module path.
+
+For example if all the modules are under a directory named `libs/`:
+
+```shell
+xtc test -L libs/ my-app.xtc
+```
+
+### Running Specific Tests
+
+By default, XUnit will run all tests in a module. If we want to run only specific tests, we can 
+use one of the command line options that allow us to specify a test or tests.
+
+The `--test-class` option allows us to specify the fully qualified name of a class to execute 
+tests in. For example if an application had a test class named `Foo`, in a package named 
+`a.b.c`, we could run the tests only in the `Foo` class using the following command:
+
+```shell
+xtc test --test-class a.b.c.Foo my-app.xtc
+```
+
+If we wanted to run tests in class `a.b.c.Foo` and `d.e.f.Bar`, then we can use the `--test-class` 
+option twice:
+
+```shell
+xtc test --test-class a.b.c.Foo --test-class d.e.f.Bar my-app.xtc
+```
+
+The `--test-method` option allows us to specify the fully qualified name of a test method to 
+execute. For example, if we wanted to run only the `testFoo` method in the `Foo` class, we could 
+use the following command:
+
+```shell
+xtc test --test-method a.b.c.Foo.testFoo my-app.xtc 
+```
+
+The `--test-package` option allows us to specify the fully qualified name of a package to execute 
+all tests in.  
+
+For example, if we wanted to run all the tests in the package `a.b.c`, we could use the 
+following command:
+
+```shell
+xtc test --test-package a.b.c my-app.xtc 
+```
--- a/lib_xunit/src/main/x/xunit/UniqueId.x
+++ b/lib_xunit/src/main/x/xunit/UniqueId.x
@@ -1,5 +1,5 @@
 /**
- * A unique identifier for a test model.
+ * A unique identifier for a test container or actual test.
  *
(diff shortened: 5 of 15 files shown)


javatools/src/main/java/org/xvm/runtime/MainContainer.java
```diff
@@ -38,21 +40,66 @@ public MainContainer(Runtime runtime, NativeContainer containerNative, ModuleCon
...
-            if (sValue != null) {
-                if (type.isNullable()) {
-                    type = type.removeNullable();
+            Object oValue = m_mapInjections.get(sName);
+            if (oValue != null) {
+                ConstantPool pool             = frame.poolContext();
+                TypeConstant typeDestringable = pool.ensureEcstasyTypeConstant("text.Destringable");
+                TypeConstant typeString       = pool.typeString();
+                TypeConstant typeStrings      = pool.ensureArrayType(typeString);
+                TypeConstant requiredType     = type;
```
ggleyzer: Trivial: we no longer require the "Hungarian" notation, but it since all variables above are, it reads strange. Maybe `typeRequired` would be more consistent?

javatools/src/main/java/org/xvm/runtime/MainContainer.java
```diff
@@ -38,21 +40,66 @@ public MainContainer(Runtime runtime, NativeContainer containerNative, ModuleCon
...
+                        return toDestringable(frame, requiredType, sValue);
+                    }
+                } else if (oValue instanceof String[] asValue && asValue.length > 0) {
+                    if (requiredType.equals(typeStrings)) {
+                        // require String[] and value is String[]
+                        return xString.makeArrayHandle(asValue);
+                    } else if (requiredType.equals(typeString)) {
+                        // require String and value is String[] so return the last element
+                        // from the array
+                        return xString.makeHandle(asValue[asValue.length - 1]);
```
ggleyzer: That seem a bit random. Why last and not first? I would say it's an error

javatools/src/main/java/org/xvm/runtime/MainContainer.java
```diff
@@ -38,21 +40,66 @@ public MainContainer(Runtime runtime, NativeContainer containerNative, ModuleCon
...
             }
         }
 
+        // No matching injectable found, so if the required type is nullable return null, otherwise
+        // return an exception
         ObjectHandle hResource = f_parent.getInjectable(frame, sName, type, hOpts);
-        return hResource == null
-                ? new DeferredCallHandle(xException.makeHandle(frame, "Invalid resource: " + sName))
-                : maskInjection(frame, hResource, type);
+        if  (hResource == null) {
```
ggleyzer: I think this would read better as a single `return` with multi-ternary

javatools/src/main/java/org/xvm/runtime/MainContainer.java
```diff
@@ -38,21 +40,66 @@ public MainContainer(Runtime runtime, NativeContainer containerNative, ModuleCon
...
+                    : new DeferredCallHandle(xException.makeHandle(frame, "Invalid resource: " + sName));
+        }
+        return maskInjection(frame, hResource, type);
+    }
+
+    private ObjectHandle toDestringable(Frame frame, TypeConstant type, String sValue) {
+        ConstantPool    pool     = frame.poolContext();
+        TypeComposition clz      = type.ensureClass(frame);
+        ClassTemplate   template = clz.getTemplate();
+        MethodStructure ctor = template.getStructure()
```
ggleyzer: Why cut in two lines? Unless you wanted to align the equal signs and it didn't fit, but now it's neither

javatools/src/main/java/org/xvm/runtime/MainContainer.java
```diff
@@ -1,10 +1,12 @@
 package org.xvm.runtime;
 
 
+import java.util.Collection;
```
ggleyzer: Unused import

lib_ecstasy/src/main/x/ecstasy/annotations/Test.x
```diff
@@ -32,15 +32,19 @@
...
  *
- * * [expectedException] - if this is non-Null, it indicates that the unit test must throw the
- *   specified type of exception, otherwise the test will be considered a failure. This option is
- *   useful for a test that is expected to always fail with an exception.
+ * * [order] - this assigns an ordering to execution of the annotated elements. Best practice is
+ *   that unit tests should be agnostic of ordering, but other annotated elements, such as XUnit
+ *   before test or after test methods, test extensions, etc. may require ordering to be specified.
+ *   Annotated elements are ordered based the priority value, where a lower value has greater
+ *   priority than a higher value. For example, `Int.MaxValue` is the lowest priority (i.e. natural
+ *   Int order).
```
ggleyzer: This is a strange priority convention; usually the bugger the number, the higher the priority.

Also, why did you remove the `expectedException` option?

lib_ecstasy/src/main/x/ecstasy/mgmt/BasicResourceProvider.x
```diff
@@ -1,3 +1,5 @@
+import ecstasy.annotations.Inject.Options;
```
ggleyzer: You added this import to use on line 63, but there other places that fully qualify it (lines 46 and 53). We should either change them all or qualify them all

lib_ecstasy/src/main/x/ecstasy/mgmt/BasicResourceProvider.x
```diff
@@ -57,11 +59,18 @@ service BasicResourceProvider
             // if the type is Nullable, no need to complain; just return Null, otherwise
             // return a deferred exception (thrown only if the container actually asks for the
             // resource at run time)
-            return Nullable.as(Type).isA(type)
-                ? Null.as(Supplier)
-                : (Inject.Options opts) ->
-                    throw new Exception($|Unsupported resource: type="{type}", name="{name}"
-                                       );
+            @Inject Injector injector;
```
ggleyzer: I'm very uneasy about this change. The idea was that the BRP itself is an `Injector` from the standpoint of the child container and serves as a "barrier", providing just the necessary minimum to the child container. You seem to open the door for any injections, fully exposing all parent's resources to the child container, which is very questionable.

lib_xunit/src/main/x/xunit/annotations/AfterAll.x
```diff
@@ -0,0 +1,62 @@
...
+ *   a single `@AfterAll` annotated function in a test class.
+ *
+ * @param order  applies an ordering to the execution of `AfterAll` annotated functions.
+ */
+annotation AfterAll(Int order = Int.MaxValue)
+        extends AbstractAll(order)
+        implements ExtensionProviderProvider {
+
+    /**
+     * Return the `AfterAll` annotated function as an `ExtensionProvider`.
```
ggleyzer: I don't see any reason to have two basically identical doc lines; I think @return would suffice and that is our usual pattern elsewhere

lib_xunit/src/main/x/xunit/annotations/BeforeAll.x
```diff
@@ -0,0 +1,59 @@
...
+ *   single `@BeforeAll` annotated function in a test class.
+ *
+ * @param order  applies an ordering to the execution of `BeforeAll` annotated functions.
+ */
+annotation BeforeAll(Int order = Int.MaxValue)
+        extends AbstractAll(order)
+        implements ExtensionProviderProvider {
+
+    /**
+     * Return the `BeforeAll` annotated function as an `ExtensionProvider`.
```
ggleyzer: Ditto - duplicate doc

lib_xunit/src/main/x/xunit/annotations/BeforeEach.x
```diff
@@ -0,0 +1,63 @@
...
+ * * Priority of annotated method execution is determined by tge `priority` property. Where methods
+ *   or functions with a lower priority are executed first. Methods or functions with the default
+ *   `Int.MaxValue` priority will be executed in order of super class annotated methods or functions
+ *   first.
+ *
+ * * Whilst priority can be used to affect ordering, it is clearer and recommended to only have a
+ *   single `@BeforeEach` annotated method in a test class.
+ *
+ * @param order  applies an ordering to the execution of `BeforeEach` annotated methods or
+ *               functions.
```
ggleyzer: Trivial - we don't put period at the end of `@param` or `@return`

lib_xunit/src/main/x/xunit/extensions/Extension.x
```diff
@@ -0,0 +1,175 @@
...
+            return clz.order;
+        }
+        return Int.MaxValue;
+    }
+
+    @RO Boolean requiresTarget.get() = True;
+
+    // ---- Orderable ------------------------------------------------------------------------------
+
+    // Highest lowest comes first
```
ggleyzer: What does that mean? :)

lib_xunit/src/main/x/xunit/extensions.x
```diff
@@ -0,0 +1,29 @@
...
+ */
+package extensions {
+
+    /**
+     * An class that determines whether a test should be skipped.
+     */
+    interface TestExecutionPredicate
+            extends Const {
+        /**
+         * Returns whether a test should be skipped.
```
ggleyzer: Maybe "Check whether a test..." would read better? Or "Evaluate ..."?

lib_xunit/src/main/x/xunit.x
```diff
@@ -0,0 +1,42 @@
...
+     * This is typically used in tests to indicate that a the preconditions for running the test
+     * cannot be met so the test is marked as skipped rather than failed.
+     */
+    const PreconditionFailed(String? text = Null, Exception? cause = Null)
+            extends Exception(text, cause);
+
+    /**
+     * A `Method` or a `Function`.
+     */
+    typedef Method<Object, Tuple<>, Tuple<>> | Function<Tuple, Tuple> | Function<<>, <Object>>
```
ggleyzer, line 30: Why did you need to parameterize it? Wouldn't simply `Method|Function` work?

lib_xunit_engine/src/main/x/xunit_engine/console/ConsoleExecutionListener.x
```diff
@@ -0,0 +1,82 @@
...
+
+	    if (result.status == Skipped) {
+	        if (ex.is(Exception)) {
+	            onSkipped(model, ex.message);
+	        } else {
+	            onSkipped(model, "");
+	        }
+	    } else {
+	        if (model.isContainer) {
+	            console.print($"Finished:  {model.displayName} {result.status} in {result.duration} (passed={result.succeeded} failed={result.failures} skipped={result.skipped} errors={result.errors})");
```
ggleyzer: There is no need to have such a long string:

	            console.print($|Finished: {model.displayName} {result.status} in {result.duration}\
	                           | (passed={result.succeeded} failed={result.failures}\ 
	                           | skipped={result.skipped} errors={result.errors})\
	                           );

lib_xunit_engine/src/main/x/xunit_engine/executor/BaseExecutionLifecycle.x
```diff
@@ -0,0 +1,46 @@
...
+
+    // ----- ExecutionLifecycle methods ------------------------------------------------------------
+
+	@Override
+    Iterable<Model> getChildren(EngineExecutionContext context) = model.children;
+
+    // ----- helper methods ------------------------------------------------------------------------
+
+	/**
+	 * Returns the fixture to execute tests against, creating a new instance
```
ggleyzer: No need to split into two lines

lib_xunit_engine/src/main/x/xunit_engine/executor/BaseExecutionLifecycle.x
```diff
@@ -0,0 +1,46 @@
...
+	 */
+	Object? ensureFixture(EngineExecutionContext context, ExtensionRegistry extensions, Class clz) {
+	    if (Object fixture := clz.isSingleton()) {
+	        // ToDo: call pre/post fixture constructor extensions
+	        return fixture;
+        }
+
+	    // ToDo: check extensions for any test fixture factory and use it if found
+
+	    MethodOrFunction? constructor = model.constructor;
```
ggleyzer: This could be simplified as:

	    if (MethodOrFunction constructor ?= model.constructor.is(MethodOrFunction)) {

lib_xunit_engine/src/main/x/xunit_engine/executor/ContainerExecutionLifecycle.x
```diff
@@ -0,0 +1,84 @@
...
+		return SkipResult.NotSkipped;
+	    }
+
+    @Override
+	EngineExecutionContext before(ExceptionCollector     collector,
+	                              EngineExecutionContext context,
+	                              ExtensionRegistry      extensions) {
+
+	    Class testClass = model.testClass;
+	    if (testClass.is(TestFixture)) {
```
ggleyzer: Ditto - can be simplified

lib_xunit_engine/src/main/x/xunit_engine/executor/ContainerExecutionLifecycle.x
```diff
@@ -0,0 +1,84 @@
...
+        }
+		return context;
+    }
+
+    @Override
+	void after(ExceptionCollector     collector,
+               EngineExecutionContext context,
+               ExtensionRegistry      extensions) {
+
+        for (AfterAllCallback after : extensions.get(AfterAllCallback, False, False)) {
```
ggleyzer: we usually name the default args, e.g. `fromParent=False, parentFirst=False`

javatools/src/main/java/org/xvm/runtime/MainContainer.java
thegridman: Initially I did use the first value. I changed it to the last because these injections come from the command line. If there is an injection key maps to a String[] that means it was entered multiple times on the command line, e.g. for and injection key "foo" on the command like there could be `-I foo=One -I foo=Two` so I coded it so that in this case the last value on the command line wins and is the one used. 

Typically I'd expect a developer to know whether the injection key was a String or String array and use the correct type at the @Inject point in the code.

But now that injections supports String and String[] there could be a case where an @Inject is supposed to just be a String but for whatever reason it appears twice on the command line. I think previously this would cause an error, but now it does not because the injection code does not know what the injectable will actually be used for. 

I'm happy to discuss this more and work out the best approach

lib_ecstasy/src/main/x/ecstasy/annotations/Test.x
thegridman: Having expected exception in the Test annotation is a poor way to assert exceptions. All it can say is that an exception of a specific type was thrown, but there is no way for a test to do any further assertions on that exception. A test might expect an exception of a given type but it was still not the expected exception.

thegridman: I have sorted out the ordering, it was a little confused and inconsistent in other places too.

lib_ecstasy/src/main/x/ecstasy/mgmt/BasicResourceProvider.x
thegridman: I have reverted the changes to this class as everything seems to work without them.

javatools/src/main/java/org/xvm/tool/Ecstasy.java
```diff
@@ -57,20 +58,53 @@ public Ecstasy(String[] asArg, Console console) {
...
+                    exitCode = new Runner(argsWithout("run"), null).run();
+                    break;
+                case "test":
+                    exitCode = new TestRunner(argsWithout("test"), null).run();
+                    break;
+                case "help":
+                    displayHelp();
+                    exitCode = 0;
+                default:
+                    out("ERROR: Unrecognised XTC sub-command: " + sCmd);
```
lagergren, line 92: In my PR, when I got rid of command lines, I made the compiler/runner/ other tool implementations work with configuration/options that are a java class that can be created with a builder pattern. The main command line args can be created from the options class (or args -> options). The complexity and shared responsibility of the Ecstasy class the way it was implemented broke the options, since there were no clear "joint" responsibilities in the code yet. It should be simple to extend this into using the Options, for one of us, or if it's OK that it goes away for a bit and comes back - are there any new code paths that use The Ecstasy or the "xtc" launcher in this branch? Let me know and I will tell you how it should port.

javatools/src/main/java/org/xvm/tool/Ecstasy.java
```diff
@@ -57,20 +58,53 @@ public Ecstasy(String[] asArg, Console console) {
...
-        // TODO JK
-        return 0;
+        return exitCode;
+    }
+
+    private List<String> argsWithout(String sArg) {
+        List<String> listArgs = new ArrayList<>(m_listArgs);
+        listArgs.remove(sArg);
+        return listArgs;
     }
```
lagergren, line 108: var listArgs = new ArrayList<String>(m_listArgs)?   

Java no longer needs the Ivan Ivanovich Ivanov pattern, and I find it helpful to use it whenever I declare something in java with an obvious type, that can be trivially derived by looking at the code, or when it is already explicit once on the assignment line, like here, on the RHS.

javatools/src/main/java/org/xvm/tool/Ecstasy.java
```diff
@@ -114,6 +158,7 @@ public Options() {
         /**
          * @return the list of files in the module path (empty list if none specified)
          */
+        @SuppressWarnings("unchecked")
         public List<File> getModulePath() {
             return (List<File>) values().getOrDefault("L", Collections.emptyList());
         }
```
lagergren, line 164: This will just go into the options soon. A modulePath is a common mechanism to all launchers.

javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -120,8 +121,15 @@ public static int launch(String[] asArg) throws LauncherException {
...
-        m_console = console == null ? DefaultConsole : console;
+        this(asArgs == null ? List.of() : List.of(asArgs), console);
+    }
+
+    /**
+     * @param listArgs  the Launcher's command-line arguments
+     */
+    public Launcher(List<String> listArgs, Console console) {
+        m_listArgs = listArgs == null ? List.of() : listArgs;
+        m_console  = console == null ? DefaultConsole : console;
```
lagergren, line 132: Again, I seem to have replaced this with a list based framework in my branch already. We have a large problem in Ecstasy when working with the code that we have ternary identities - things can be null, things can be empty lists or arrays, and things can be non empty. The first two things usually mean hte same thing. I would assert that we never put anything in here than lists. Null is the same as an emptyList - List.of() (which is also a singleton) . This ternary "null and empty may mean slightly different things" and the usage of arrays rather than Java lists has caused me HUGE amounts of trouble when trying to get a programmatic API to work from shortest possible Java, so I highly suggest that we do not use nulls. 

In the places where we DO have to use nulls, like parent directories, Java Optional is usually a lot better, so that we can use lambda programming and chained operators instead of adding null checks or empty checks everywhere.

javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -1675,16 +1711,17 @@ public String desc() {
      * Represents a single available command line option.
      */
     public static class Option {
-        public Option(String sPosix, String sLinux, Form form, boolean fMulti, String sDesc) {
+        public Option(String sPosix, String sLinux, Form form, boolean fMulti, String sDesc, String sInjectionKey) {
             assert sPosix != null || sLinux != null;
             assert (sPosix == null || !sPosix.isEmpty()) && (sLinux == null || !sLinux.isEmpty());
             assert form != null;
```
lagergren, line 1718: It seems you have implemented the same Options pattern , I think we should sync and figure out how to unify it!  I did haul of most of the complicated flag parsing loops to Apache CLI which is really slim

javatools/src/main/java/org/xvm/tool/Runner.java
```diff
@@ -448,8 +474,27 @@ public String[] getMethodArgs() {
...
+            // Add any options that have an injection key defined
+            options().values()
+                    .stream()
+                    .filter(Option::isInjection)
+                    .forEach(o -> {
+                        Object value = values.get(o.posixName());
+                        if (value instanceof String sValue) {
+                            injections.put(o.injectionKey(), sValue);
+                        } else if (value instanceof Collection<?> col && !col.isEmpty()) {
+                            String[] asValue = col.stream().map(String::valueOf).toArray(String[]::new);
```
lagergren, line 492: Is there a reason that we have to represent this as arrays in the injections table? Lists tend to be more Java native than the array reflection these days (also more compact, for the general case)

plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
```diff
@@ -345,6 +362,24 @@ private void createDefaultRunTask(final Project project) {
...
+     */
+    private void createDefaultTestTask(final Project project) {
+        final var testTaskName = getTestTaskName();
+        // The test task depends on all compile tasks, for all source sets these days.
+        final var compileTaskNames = getSourceSets(project).stream().map(sourceSet -> sourceSet.getCompileTaskName(XTC_LANGUAGE_NAME));
+        final var testTask = tasks.register(testTaskName, XtcTestTask.class, project);
+        testTask.configure(task -> {
+            task.setGroup(APPLICATION_GROUP);
+            task.setDescription("Test an XTC program with a configuration supplying the module path(s).");
+            logger.info("[plugin] Configured, dependency to tasks: {} -> {}", XDK_EXTRACT_TASK_NAME, compileTaskNames);
```
lagergren, line 378: The plugin changes look great.  I have made some things a little bit easier, but I think they can be used just as they are written. Nicely done!

plugin/src/main/java/org/xtclang/plugin/internal/DefaultXtcTestExtension.java
```diff
@@ -0,0 +1,133 @@
...
+    public void setModuleNames(final String... moduleNames) {
+        setModuleNames(Arrays.asList(moduleNames));
+    }
+
+    @Override
+    public void setModules(final List<XtcTestModule> modules) {
+        this.modules.empty();
+        modules.forEach(this::addModule);
+    }
+
```
lagergren, line 113: Since this is pretty much the same interface as the launcher tasks (run and compile), if might be worth it to ask claude-code how to unify them. I think that you would be able to use only one class with these public dsl accessor definitions - and I think Gradle understands that and makes public properties available in subclasses. This is typically something that I relish that I no longer have to figure out manually - and if it works, I will have compilable code automatically.

javatools/src/main/java/org/xvm/tool/Runner.java
thegridman: Inside the main container in Ecstasy injectables are either String or String[] If I used a List here I would just end up converting it to a String array inside the main container anyway.

javatools/src/main/java/org/xvm/runtime/MainContainer.java
```diff
@@ -38,23 +39,64 @@ public MainContainer(Runtime runtime, NativeContainer containerNative, ModuleCon
...
         // check the custom injections first
         if (m_mapInjections != null) {
-            String sValue = m_mapInjections.get(sName);
-            if (sValue != null) {
-                if (type.isNullable()) {
-                    type = type.removeNullable();
+            Object oValue = m_mapInjections.get(sName);
+            if (oValue != null) {
+                ConstantPool pool             = frame.poolContext();
+                TypeConstant typeDestringable = pool.ensureEcstasyTypeConstant("text.Destringable");
```
lagergren: This code nests deeper and deeper, but pretty much everything in this rather large method is inside an if, or an if in an if - It is a nitpick, but I find guards/early returns a lot easier to read.

lagergren:
I guess we can close this one now, @thegridman ? You have probably gotten all this already from the other merge of our joint branch. The conflicts now shown are probably mostly doing the identical changes under my branch, including adding improvements, but you may want to go over them here to make sure you didn't miss anything if that is a concern. 

After that please close this PR.

PR #345, opened 2025-12-05
Add Medium size model support to JSON DB

thegridman:
This PR adds Medium model support to JSON DB.

JSON DB Small model support stores all data in memory. For stores that can hold a lot of data (specifically `JsonMapStore`) the Medium model stores key in memory and values off heap.

- A JSON DB starts with en Empty model and then transitions to Small when data is first added.
- When the small size limits (based on bytes used or files used) is exceeded the model transitions to Medium.
- The model will grow when committing a transaction. At the point of committing if a model has transitioned then all data of the commit is stored off heap.
- After a store has moved to the Medium model, the periodic DB maintenance will move any remaining values off heap.
- DB maintenance could cause a model to transition back to Small if it deletes old files.
- When a model has transitioned from Medium back to Small values are moved back to the heap as they are read by the application

### DIFF
--- a/manualTests/src/main/x/jsondb/README.md
+++ b/manualTests/src/main/x/jsondb/README.md
@@ -1,0 +1,31 @@
+# JSON DB Tests
+
+This module contains the tests for the `jsondb.xtclang.org` module. These tests are run using the 
+XUnit framework.
+
+## Running the Tests
+
+To run the full test suite use the `xtc test` command from the `manualTests` directory.
+
+```shell
+xtc test -L build/xtc/main/lib -o build/xtc/main/lib src/main/x/jsondb/jsondb_test.x
+```
+
+### Running Tests in One Class
+
+The `xtc test` command allows you to run a single test class using the `--test-class` or `-c` 
+options. The argument to `--test-class` or `-c` is the fully qualified name of the test class.
+
+For example, to run the tests in the `storage.JsonMapStoreTest` class use the following command:
+
+```shell
+xtc test -L build/xtc/main/lib -o build/xtc/main/lib \
+    --test-class storage.JsonMapStoreTest
+    src/main/x/jsondb/jsondb_test.x
+```
+
+## Test Output
+
+The tests create test databases in the `build/test-output` directory. 
+There will be a subdirectory for each test class and within that will be a subdirectory for each test.
+


lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
```diff
@@ -41,8 +41,76 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
         this.jsonSchema   = catalog.jsonSchema;
         this.keyMapping   = keyMapping;
         this.valueMapping = valueMapping;
+
+        // Model default sizes can be overridden by injectable configuration values
+        @Inject(ConfigSmallModelMaxBytes)
```
ggleyzer: I think an in-lined version reads better:

        @Inject(ConfigSmallModelMaxBytes)   Int? smallModelBytesMax;
        @Inject(ConfigSmallModelMaxFiles)   Int? smallModelFilesMax;
        @Inject(ConfigMediumModelMaxBytes)  Int? mediumModelBytesMax;
        @Inject(ConfigMediumModelMaxFiles)  Int? mediumModelFilesMax;

lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
```diff
@@ -159,6 +244,78 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
...
+    /**
+     * Check a value is valid to use for the smallModelBytesMax property.
+     *
+     * @param value                the value to validate
+     * @param mediumModelBytesMax  the medium model max bytes value to use for validation
+     *
+     * @throws IllegalArgument if the value is invalid
+     */
+    static void validateSmallModelBytesMax(Int value, Int mediumModelBytesMax) {
+        assert:arg value > 0 as $"Invalid smallModelBytesMax {value} JSON DB DBMap small model maximum bytes must be greater than zero. ";
```
ggleyzer: Please use multi-line literals, e.g.:

        assert:arg value > 0 as 
            $|Invalid smallModelBytesMax {value} JSON DB DBMap small model maximum bytes must be \ 
             |greater than zero.
             ;
        assert:arg value < mediumModelBytesMax as 
            $|Invalid smallModelBytesMax {value} JSON DB DBMap small model maximum bytes must be \ 
             |less than medium model maximum bytes {mediumModelBytesMax}.
             ;

lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
```diff
@@ -159,6 +244,78 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
...
+    /**
+     * Check a value is valid to use for the smallModelFilesMax property.
+     *
+     * @param value                the value to validate
+     * @param mediumModelFilesMax  the medium model max files value to use for validation
+     *
+     * @throws IllegalArgument if the value is invalid
+     */
+    static void validateSmallModelFilesMax(Int value, Int mediumModelFilesMax) {
+        assert:arg value > 0 as $"Invalid smallModelFilesMax {value} JSON DB DBMap small model maximum files must be greater than zero. ";
```
ggleyzer: ditto

lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
```diff
@@ -699,6 +883,28 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
         for (Int writeId : writeIds) {
             inFlight.remove(writeId);
         }
+
+        // After a commit the model could have become larger.
+        StorageModel newModel  = checkModelSize();
+        StorageModel prevModel = model;
+        if (newModel != prevModel && newModel != Small) {
+            // The model size has changed and is not Small (there is nothing to do going from Empty to Small)
+            lastModelTransitionTx = lastCommit;
```
ggleyzer: You assign `lastModelTransitionTx`, but never read. What was the intent?

lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
```diff
@@ -699,6 +883,28 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
         for (Int writeId : writeIds) {
             inFlight.remove(writeId);
         }
+
+        // After a commit the model could have become larger.
+        StorageModel newModel  = checkModelSize();
```
ggleyzer, line 897: I wonder if the `commit` is the right time. Maybe the `doMaintenance` is? It would also prevent a potential chatter

lib_jsondb/src/main/x/jsondb/storage/ObjectStore.x
```diff
@@ -146,6 +146,11 @@ service ObjectStore(Catalog catalog, DboInfo info)
      */
     public/protected StorageModel model = Empty;
 
+    /**
+     * A simple enum used to indicate a value that is stored on disc.
+     */
+    enum DiscStorage {OnDisc}
```
ggleyzer: This seems to be no longer used

manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
```diff
@@ -0,0 +1,345 @@
...
+import jsondb.storage.JsonMapStore;
+import jsondb.storage.JsonMapStore.History;
+import jsondb.storage.JsonMapStore.MapValue;
+import jsondb.storage.JsonMapStore.Marker;
+import jsondb.storage.JsonMapStore.Marker.OffHeap;
+import jsondb.Client.DBObjectImpl;
+import jsondb.Client.DBMapImpl;
+import jsondb.TxManager;
+
+import test_db.*;
```
lagergren, line 14: You can probably move the json_db tests that are self contained and self tests for the json_db module into that, even though they are part of the XDK. The dependency substitution mechanism should resolve things fine in a src/test/x source set dir, and work with xunit - I think.  I did some analysis of how I actually wrote the xdk self-bootstrap and the Gradle substitution mechanism that substitutes semver artifact descriptions with their projects, if they exist in the same build, seems to be working fine. There SHOULD be no bootstrap issues, but I can't promise 100%.

lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
```diff
@@ -757,6 +971,25 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
...
+                        entry.value = OffHeap;
+                    }
+                }
+            } else if (modelAtCleanup <= Medium && modelNow == Large) {
+                // the model has grown to Large, mark all history keys and values as OffHeap
+                // ToDo we need to add logic for transition to Large after we add Large model support
+            }
+            modelAtCleanup = modelNow;
+            model          = modelNow;
+        }
```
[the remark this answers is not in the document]
thegridman: We're ok with the inefficiency at the moment.

PR #347, opened 2025-12-09
Programmatic APIs for Compiler/Runner and Plugin Execution Strategy Modernization

lagergren:
## Summary

This PR introduces programmatic APIs for the XTC Compiler, Runner, and Disassembler tools, enabling direct invocation without command-line parsing. This is the foundation for Language Server Protocol (LSP) support and IDE integrations, while also modernizing the Gradle plugin's execution model.

### Key Changes

**Programmatic Launcher APIs (`LauncherOptions`)**
- New `CompilerOptions`, `RunnerOptions`, and `DisassemblerOptions` classes with fluent Builder APIs
- Options can be constructed programmatically OR parsed from command-line arguments
- Immutable options objects with clear separation between configuration and execution
- JSON serialization support for options (useful for language server communication)
- Proper error handling via `LauncherException` with severity levels

**Plugin Execution Strategy Refactoring**
- Replaced monolithic `JavaClasspathLauncher` with clean strategy pattern:
  - `DirectStrategy`: In-process execution via ServiceLoader (fastest, shares JVM with Gradle)
  - `AttachedStrategy`: Forked JVM with inherited I/O (default - isolated with visible output)
  - `DetachedStrategy`: Background process with file redirects (for long-running servers)
- New `ExecutionMode` enum for clear mode selection
- Simplified `XtcLauncherTask` using `LauncherOptionsBuilder` to construct options
- Removed legacy `CommandLine`, `LauncherContext`, `SimpleExecResult` classes

**XDK Integration Tests**
- New comprehensive test suite (`XdkIntegrationTest`) that validates:
  - Builder API compilation and execution
  - Command-line parsing API
  - Error handling for invalid arguments and missing dependencies
  - Module path resolution with real XDK libraries
- Tests run against the fully built XDK distribution

**Code Modernization**
- Replaced `Collections.unmodifiableMap()` with `Map.copyOf()`
- Replaced stream collectors with `Set.copyOf()` where appropriate
- Refactored visitor pattern in `ModuleInfo` to functional Stream-based approach
- Removed mutable accumulator fields and synchronized blocks
- Eliminated star imports across changed files

### Why Programmatic APIs?

The traditional command-line-only approach has several limitations:
1. **IDE Integration**: IDEs and language servers need to invoke the compiler without forking processes
2. **Error Recovery**: Programmatic APIs can return structured errors instead of exit codes
3. **Performance**: In-process execution avoids JVM startup overhead for rapid feedback
4. **Testability**: Direct API calls are easier to test than subprocess invocations

### Plugin Execution Model

The plugin defaults to `ATTACHED` mode (forked process) for several reasons:
- **Isolation**: Compiler crashes don't affect the Gradle daemon
- **Incremental Builds**: Clean process state for each invocation
- **Error Handling**: Exit codes and output capture work reliably
- **Robustness**: Memory leaks or resource exhaustion are contained

However, the new `DIRECT` mode enables the programmatic API for scenarios where performance matters more than isolation (e.g., language server, IDE plugins).

### Exception Handling Note

The `checkErrors()` and similar methods in the launcher may throw exceptions based on error severity thresholds. This behavior is intentional for command-line use but requires care in programmatic contexts. Future work will provide more granular control over error reporting vs. exception throwing.

## Test Plan

- [x] All existing tests pass
- [x] New `XdkIntegrationTest` validates programmatic APIs against real XDK
- [x] Plugin tests verify execution strategies work correctly
- [x] Manual testing of `./gradlew xdk:test` with full XDK build


javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -1,162 +1,284 @@
...
+     * @param asArg  command line arguments (first arg is command name: xcc, xec, or xtc)
      */
-    public static void main(String[] asArg) {
+    static void main(final String[] asArg) {
         try {
-            // use System.exit() to communicate the result of execution back to the caller
             System.exit(launch(asArg));
-        } catch (LauncherException e) {
-            System.exit(e.error ? 1 : 0);
+        } catch (final LauncherException e) {
```
ggleyzer, line 159: Here and in many other places you added "final" to the declaration. What is the purpose of that? We have a hundred of "catch" lines across our code line and never done that. I think the consistency of the code is valuable for a reader. A deviation of the common style usually says: "pay attention, there is a reason this code looks different". I don't see any reason for that deviation here, am I mistaken? If so, we may need to change all those places

javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -1,162 +1,284 @@
...
+     * Executes a launcher command and returns an exit code.
+     * Use this when calling from a daemon or other long-running process.
+     *
+     * @param cmd command name (xcc or xec)
+     * @param args command line arguments (without the command name)
+     * @param console console for output (must not be null)
+     * @param errListener optional ErrorListener to receive errors, or null
+     * @return exit code (0 for success, non-zero for error)
+     */
+    public static int launch(final String cmd, final String[] args, final Console console, final ErrorListener errListener) {
```
ggleyzer, line 200: Ditto here. What is the point of declaring the arguments "final"?

javatools/src/main/java/org/xvm/tool/ResourceDir.java
```diff
@@ -24,15 +25,15 @@ public class ResourceDir {
      *                     location to use as the entire resource path
      */
     public ResourceDir(File resourceLoc) {
-        this(new File[] {resourceLoc});
+        this(List.of(resourceLoc));
```
ggleyzer, line 31: And you didn't declare "resourceLoc" as "final" neither here nor anywhere below. This "final" thing looks completely arbitrary

javatools/src/main/java/org/xvm/tool/Launcher.java
lagergren: I more or less do it as a central nervous system reflex. Everything SHOULD be final unless purposefully been granted a state. This is the code standard I normally use for Java, and the plugin is written that way. I would argue that parameters, exceptions and variable should as be final as can be. Any non final state is a non-normal case. It has usually been enforced in my code standards, like the javatools stuff in the JDK for subprojects, and IntelliJ thinks it's "yellow" to allow potential mutating state.  So I would be in favour of changing it everywhere. It has gotten easier in Java with the "var" keyword, where something is trivially typed elsewhere on the same assignment line to create code that has the final keyword that doesn't add too much verbosity.   

In Kotlin and Scala it's "var" or "val", the rule of thumb is pretty much that you do not use var unless you mean "mutable". I like that order of precedence.   I also tend to declare my methods with final parameters, since I think it's a code smell and may potentially disrupt the functional by-default contract to overwrite them, which in turn can lead to bugs and hacky solutions. This has also been enforced by all Java code bases I've worked with the last decade. 
Nothing else than final parameters can be checked in, and that seems to be a pretty common practice. 

For catches especially - we don't want anyone to reassign e = .. in the catch block, which is (horribly) allowed by default in Java. No one usually does that, so the final here is just the same for me as final parameters to methods, final fields that ideally are written to ONLY during construction time, and so on. It is a small vaccination effort against stateful code where no stateful code should be.  In Java it is usually just as simple to take the functional approach for most code and no good reason to reuse state. This is not universally true, but I find it helps prevent incremental statefulness, which has given me problems more than once.

javatools/src/main/java/org/xvm/tool/Launcher.java
lagergren: Most code standards do. It is a knee jerk effect and an intellij autocomplete.  To me, every field, variable and parameter that cannot be reassigned (especially fields and parameters), hardens the codebase a little bit against non functional approaches, quick fixes and and potential future bugs.  There was the accumulator thing under a synchronised block writing timestamps to a field in the ModuleInfo.java, for example, which caused my reentrancy analysis / builder based approach some grief. With mandated finals, you could not even write that - you would have to implement it with the (functional) streaming approach that has been the recommended way to do things in Java 8. I think it has been worth it to me over the years, when it comes to extensibility and preventing errors. I also haven't had the choice, so I am just not used to non-final defaults in Java for the last decade.

(For example - The ModuleInfo class IMHO should really have either all-final fields instantiated at construction time, and / or implement lazy computation with one of the more preferred ways than if (this.field == null) this.field = ... - e.g. lazily with factories using lambda syntax, which will do the same thing, but without having to use non final fields.  Once an instance is fully immutable, it hashes trivially, comparesTo trivially etc, so code gets better from it.  And there is really little reason to do otherwise since you can write the same kind of lazy one liners in Java 8+ without resorting to the "1.0" if field == null pattern.)

ggleyzer: final fields is a completely separate issue and we do do it wherever possible. My comment has nothing to do with that. We have a source code base of hundreds of files and thousands of method. In my opinion, changing the project style in one or two classes creates unnecessary entropy and adds basically nothing of value.

javatools/src/main/java/org/xvm/tool/ResourceDir.java
lagergren: If you want I can click the "auto-final", if you are fine with that- Be happy to. I just changed as little code as possible outside the plugin and Options, but of course I didn't want to change everything everywhere so the changes drowned in that. Sorry just tried to minimise the changes and I was barely aware of the finals - They just automatically get there when I tab / auto complete everything!

ggleyzer: Yes; let's keep the number of stylistic changes to a minimum and try to maintain the code style consistency. Thank you!

javatools/src/main/java/org/xvm/tool/Launcher.java
lagergren: I understand your concerns about inconsistency, but I disagree. IntelliJ has several notifications about the code and potential improvements, and given final parameters, in this case, it has been able to do static analysis of things that haven't been seen before and deduce further simplifications. Especially coming into the code base and trying to find ones way around, to me it gives me a huge advantage if know the names are single assigned on creation when I step through and understand how the code works. With a raised level of IDEA assitance, there is a noticeable difference in efficiency for both me and the machine. The parameter finality has concrete quality and development time improvements, which is not just my subjective opinion, enough though of course I have been conditioned to it as well.

javatools/src/main/java/org/xvm/tool/ResourceDir.java
lagergren: Removing unncessary arrays, when the ground type is lists is an improvement. I did change the constructor that takes the resource list, because it creates potential tech depth to pass 1) mutable arrays, and there is no way to make them immutable, and sooner or later someone is going mutate them, where not intended, and it will take hours to debug 2) The list were the input form, and it lends it self to readable Java in an order of magniture more ways than arrays do, so in general it is preferable. There are no longer any performance differences to speak of, so I just wanted to avoid messy conversions back and forth and remove unnecessary formatting logic that was on my control flow path.

thegridman (approved):

PR #352, opened 2025-12-15
Programmatic APIs for Compiler/Runner and Plugin Execution Strategy Modernization (without finals in existing files)

lagergren:
## Summary

This PR introduces programmatic APIs for the XTC Compiler, Runner, and Disassembler tools, enabling direct invocation without command-line parsing. This is the foundation for Language Server Protocol (LSP) support and IDE integrations, while also modernizing the Gradle plugin's execution model.

### Key Changes

**Programmatic Launcher APIs (`LauncherOptions`)**
- New `CompilerOptions`, `RunnerOptions`, and `DisassemblerOptions` classes with fluent Builder APIs
- Options can be constructed programmatically OR parsed from command-line arguments
- Immutable options objects with clear separation between configuration and execution
- JSON serialization support for options (useful for language server communication)
- Proper error handling via `LauncherException` with severity levels

**Plugin Execution Strategy Refactoring**
- Replaced monolithic `JavaClasspathLauncher` with clean strategy pattern:
  - `DirectStrategy`: In-process execution via ServiceLoader (fastest, shares JVM with Gradle)
  - `AttachedStrategy`: Forked JVM with inherited I/O (default - isolated with visible output)
  - `DetachedStrategy`: Background process with file redirects (for long-running servers)
- New `ExecutionMode` enum for clear mode selection
- Simplified `XtcLauncherTask` using `LauncherOptionsBuilder` to construct options
- Removed legacy `CommandLine`, `LauncherContext`, `SimpleExecResult` classes

**XDK Integration Tests**
- New comprehensive test suite (`XdkIntegrationTest`) that validates:
  - Builder API compilation and execution
  - Command-line parsing API
  - Error handling for invalid arguments and missing dependencies
  - Module path resolution with real XDK libraries
- Tests run against the fully built XDK distribution

**Code Modernization**
- Replaced `Collections.unmodifiableMap()` with `Map.copyOf()`
- Replaced stream collectors with `Set.copyOf()` where appropriate
- Refactored visitor pattern in `ModuleInfo` to functional Stream-based approach
- Removed mutable accumulator fields and synchronized blocks
- Eliminated star imports across changed files

### Why Programmatic APIs?

The traditional command-line-only approach has several limitations:
1. **IDE Integration**: IDEs and language servers need to invoke the compiler without forking processes
2. **Error Recovery**: Programmatic APIs can return structured errors instead of exit codes
3. **Performance**: In-process execution avoids JVM startup overhead for rapid feedback
4. **Testability**: Direct API calls are easier to test than subprocess invocations

### Plugin Execution Model

The plugin defaults to `ATTACHED` mode (forked process) for several reasons:
- **Isolation**: Compiler crashes don't affect the Gradle daemon
- **Incremental Builds**: Clean process state for each invocation
- **Error Handling**: Exit codes and output capture work reliably
- **Robustness**: Memory leaks or resource exhaustion are contained

However, the new `DIRECT` mode enables the programmatic API for scenarios where performance matters more than isolation (e.g., language server, IDE plugins).

### Exception Handling Note

The `checkErrors()` and similar methods in the launcher may throw exceptions based on error severity thresholds. This behavior is intentional for command-line use but requires care in programmatic contexts. Future work will provide more granular control over error reporting vs. exception throwing.

## Test Plan

- [x] All existing tests pass
- [x] New `XdkIntegrationTest` validates programmatic APIs against real XDK
- [x] Plugin tests verify execution strategies work correctly
- [x] Manual testing of `./gradlew xdk:test` with full XDK build

---

## Note on `final` Parameters

This branch is identical to `lagergren/options-phase-1` except that `final` qualifiers have been removed from method parameters, catch clause exception variables, and for-each loop variables **in files that existed in master**. New files (e.g., `LauncherOptions.java`, `Console.java`, `TestRunner.java`, and all new plugin classes) retain their `final` qualifiers.

This change was made to reduce friction in code review. However, this is somewhat uncomfortable because the `final` parameters were introduced intentionally as part of moving toward clearer API contracts.

### Why `final` Parameters Matter

The original intent was to establish explicit contracts for the XVM programmatic APIs. When code is consumed externally (by the Gradle plugin, future language servers, IDE integrations), callers need to trust that:

1. **Parameters won't be silently substituted** — If a method accepts `null` and internally does `if (errs == null) errs = BLACKHOLE`, the caller has no idea their error listener was discarded
2. **Contracts are explicit** — `@NotNull final ErrorListener errs` says exactly what the method needs; the caller explicitly passes `BLACKHOLE` if they want to discard errors
3. **Errors surface at API boundaries** — Not deep in call chains where null propagates silently

The XVM codebase currently has 625 `return null` statements with varying semantics, and null propagates through layers with ambiguous meaning. Many comments say things like "an error must've been reported" — wishful thinking that often isn't true.

### The Broader Issue

The concern isn't just `final` — it's that without clear API boundaries, every layer ends up doing its own input sanitization. This leads to:
- Errors being swallowed or lost in call chains
- `null` meaning different things (error, not-found, not-applicable, caller-doesn't-care)
- Bugs surfacing far from their origin

Modern languages (Kotlin, Rust, Swift) default to immutable parameters and non-null types precisely because the industry learned these lessons.

### Future Direction

It is strongly suggested that we incrementally move the XVM APIs toward explicit contracts — starting with the `org.xvm.tool` package that external consumers use. This PR without finals is a compromise for reviewability, but the finals should be restored as part of a broader effort to establish clear API boundaries.

### DIFF
--- a/lib_xunit_engine/src/main/x/xunit_engine/TestResourceProvider.x
+++ b/lib_xunit_engine/src/main/x/xunit_engine/TestResourceProvider.x
@@ -25,7 +25,10 @@ service TestResourceProvider(Directory curDir)
     /**
      * The `FileStore` to use to access files.
      */
-    @Lazy FileStore store.calc() = new DirectoryFileStore(curDir);
+    @Lazy FileStore store.calc() {
+        @Inject FileStore storage;
+        return storage;
+    }
 
     /**
      * The build output directory.


javatools/src/main/java/org/xvm/javajit/JitConnector.java
```diff
@@ -77,9 +75,9 @@ public void invoke0Impl(MethodStructure methodStructure, String... asArg) {
 
         TypeSystemLoader loader = ts.loader;
         try {
-            Class  clz    = loader.loadClass(typeName);
-            Ctx    ctx    = Ctx.get();
-            Object module = clz.getDeclaredConstructor(Ctx.class).newInstance(ctx);
+            var clz    = loader.loadClass(typeName);
```
ggleyzer, line 79: I really don't see any reason for this change. Does it read better? I don't think so.

javatools/src/main/java/org/xvm/javajit/JitConnector.java
```diff
@@ -93,7 +91,7 @@ public void invoke0Impl(MethodStructure methodStructure, String... asArg) {
                 this.result = lr;
             }
         } catch (ClassNotFoundException | NoClassDefFoundError e) {
-            e.printStackTrace();
+            e.printStackTrace(System.err);
```
ggleyzer, line 95: This is unnecessary change (it's definitely equivalent)

javatools/src/main/java/org/xvm/javajit/JitConnector.java
```diff
@@ -110,7 +108,7 @@ public void invoke0Impl(MethodStructure methodStructure, String... asArg) {
                         cause.getClass().getField("exception").get(cause));
                 } catch (Throwable ignore) {}
             } else {
-                e.printStackTrace();
+                e.printStackTrace(System.err);
```
ggleyzer, line 112: ditto

javatools/src/main/java/org/xvm/tool/Compiler.java
```diff
@@ -98,221 +102,203 @@
...
     }
 
-    /**
-     * Helper method for external launchers.
+    protected Strictness strictLevel = Strictness.Normal;
 
-     * @param asArg  command line arguments
-     *
-     * @return the result of the {@link #process()} call
+    protected ModuleRepository repoLib;
```
ggleyzer: What is a reason to move fields declarations from the section at the end of the file that says "---- fields ----" under a section in the top of the file that says "----- constants -----"?

I'm fine with fields declared at the top or at the bottom, but randomly changing this in one file and under an incorrect section header seems very strange. Besides, it makes the work of reviewer dramatically harder. Are there any mo-trivial changes here? I cannot say without spending a lot of time unnecessraly

ggleyzer (changes requested):
It's extremely hard to review such a humongous change where 80% of the changes are trivial reformatting or refactoring. I would suggest splitting the change in two parts. One that has nothing but reformatting/refactoring. And the other with non-trivial changes/bug fixes/enhanced implementations. That would make the review process much more efficient IMHO

javatools/src/main/java/org/xvm/tool/Runner.java
```diff
@@ -1,231 +1,167 @@
...
+            checkErrors("module identification");
 
-            fileBin   = info.getBinaryFile();
+            fileBin   = requireNonNull(info).getBinaryFile();
             binExists = fileBin != null && fileBin.exists();
 
             boolean fCompile = false;
             if (!binExists) {
-                String qualName = info.getQualifiedModuleName();
+                var qualName = info.getQualifiedModuleName();
```
ggleyzer, line 126: I strongly believe that there is no reason to replace this simple line:
`String qualName = info.getQualifiedModuleName();`
with this:
`var qualName = info.getQualifiedModuleName();` 
What does this accomplish?

javatools/src/main/java/org/xvm/tool/Launcher.java
```diff
@@ -1,162 +1,370 @@
...
-        Options opts = options();
+        final T opts = options();
 
-        boolean fHelp = opts.parse(m_asArgs);
-        if (Runtime.version().version().getFirst() < 21) {
-            log(Severity.INFO, "The suggested minimum JVM version is 21; this JVM version ("
-                    + Runtime.version() + ") appears to be older");
+        final Runtime.Version jdkVersion = Runtime.version();
+        if (jdkVersion.version().getFirst() < JDK_VERSION_MIN) {
+            log(INFO, "The suggested minimum JVM version is 21; this JVM version ({}) appears to be older", Runtime.version());
```
ggleyzer: There is already "jdkVersion" var

javatools/src/main/java/org/xvm/javajit/JitConnector.java
lagergren: It certainly reads better than a fat raw type warning on Class clz. If you do not want to use generics, in this case a wildcard, at least var can help us out. So yes, I would certainly say it reads better. Clazz<?> would be the alternative.

javatools/src/main/java/org/xvm/javajit/JitConnector.java
lagergren: It is best practice and picked up as an error warning. Any posix program would expect System.err as the stack trace stream, or it will have the unwanted behavior of you trying to pipe out errors from a run, but still seeing stack traces with no other explanation.  It is also a warning to leave the printStackTrace without a stream, because there is no control over where it goes. You can change it to System.out, but that would violate standard output / standard error conventions.  Also it is a big yellow warning from IntelliJ out of the box to skip the stream.
'

javatools/src/main/java/org/xvm/javajit/JitConnector.java
lagergren: It is not unnecessary. Why would we print stack traces to stdout, and why would we call a console output function without clearly specifying its output?

javatools/src/main/java/org/xvm/tool/Compiler.java
lagergren: The fields are typically at the top in every known Java code standard, and I don't even think this was a manual move. I could put it back, but i find that there is significantly more work involved trying to figure out field types and page down to the end of the file. If it is important to you, I can put them back at the bottom, but I think that is harder to read. Especially when it comes to things like static final constants and similar things that I would normally expect to find near the constructor.

lagergren: There is also no IDE support that an understand and use sections like "---- constants -----" so there is also no "shortcut" if I want to find them.

Oracle code standard mandates fields at the top of the field in decreasing order of visibility, static fields before private fields, final fields before mutable fields in each accessibility group. Is there a good reason to do something different?

javatools/src/main/java/org/xvm/tool/Runner.java
lagergren: The reason that is now a var, is that there were other sibling variables in this logic, for example the set of possibles, and since the code was rather complicated to puzzle out, and there were collection types around, that simplified to var, given their easily readable surrounding context, leaving a single variable getter that was used together with the others with a String return type didn't feel like it fit with the rest of the logic. I would argue that anything called getQualifiedModuleName() carries 100% of type type information to the reader of the code here. As a single change, I would likely have left it alone, but I felt this was a benefit for readability.

javatools/src/main/java/org/xvm/tool/Launcher.java
lagergren: There is, but it is certainly not set to 21. 

Also - any version information about JDKs lives in gradle.properties, which requires a bootstrapped plugin framework, that I would very much prefer to not have to bring in into the outermost most generic entry point for the Launchers. I would suggest this is better than the previously hard coded 21.

javatools/src/main/java/org/xvm/javajit/JitConnector.java
ggleyzer: I completely disagree. I strongly believe that one of the main purposes of the code is to make it as friction-less as possible for a reader to quickly understand. Having "var ctx" instead of "Context ctx" does nothing but obscures

javatools/src/main/java/org/xvm/javajit/JitConnector.java
ggleyzer: The documentation for `printStackTrace()` clearly states "Prints this throwable and its backtrace to the standard error stream." What "unwanted behavior" could we be possibly talking about here?

javatools/src/main/java/org/xvm/javajit/JitConnector.java
ggleyzer: It is absolutely unnecessary. What "stdout" are you talking about?
> Prints this throwable and its backtrace to the standard error stream.

javatools/src/main/java/org/xvm/tool/Compiler.java
ggleyzer: You added the "--- constants --- " section. And you are modifying just one file to the "Oracle code standard" leaving 1300 others unchanged. 

If we decide to change all our code to the Oracle code standard, we'll do it all at once, not as a piece meal

javatools/src/main/java/org/xvm/tool/Runner.java
ggleyzer: In my view there is no benefit at all. So now when someone gets to fix something in this file do you think they are entitled to change it back if they "feel this was a benefit for readability." What a mess that would become. I think we should never change some one else's code based on our own subjective "feelings"

javatools/src/main/java/org/xvm/tool/Launcher.java
ggleyzer: What I simply meant was to use jdkVersion var instead of calling Runtime.version() again

javatools/src/main/java/org/xvm/javajit/JitConnector.java
lagergren: IntelliJ's inspection for printStackTrace() does indeed go away when you explicitly specify the stream. That particular inspection is on by default, and it checks several other "anitpatterns" with logging. The full warning is "replace all printStackTrace with more robust error logging", which is a valid point, but does not explicitly bring up anything else.  What it means is that things like this should in the future be moved to something like logger.exception("..", e) which may or may not display a stack trace depending on the warning level. 

Adding an explicit stream parameter removes the warning

So the IntelliJ warning is about intentionality — when you write printStackTrace(System.err), you're signaling "yes, I deliberately want this to go to stderr" rather than just calling the no-arg method without thinking about it. 

* The explicit form silences the IntelliJ warning
* It documents intent — someone reading the code knows the author consciously chose stderr
* It's consistent with the principle of being explicit rather than relying on defaults

You are only focusing purely on the runtime behavior for this ("it already goes to stderr"), while ignoring the code quality aspect. Yes, the behavior is identical, but the explicit form:

* Passes static analysis without warnings
* Makes the code's intent clearer
* Shows the author considered where the output should go

*Note: It's the same reason we might write equals(false) vs !value in some contexts, or explicitly specify charset in String.getBytes(StandardCharsets.UTF_8) rather than relying on platform default — being explicit has value beyond just the runtime behavior*

So it should either be tagged with the appropriate suppression, so that we get rid of the warning, and then still have some kind of TODO intent visible in the code for better logging in the future, or we could put an explicit stream argument here, even if it is the default stream, signalling intent. This is not just something that intellij recommends, it is a very common lint/rewrite operation for most java code bases and is done automatically. 

Don't you get the "use better logging" in intellij? If you don't you must have explicitly disabled it since it is part of the Java code inspections that come with it in a clean install.   And if you do get it - doesn't it make sense to handle it in some way (for example, by using a specific parameter, even though it is a default).

Finally - Looking at the full XVM code base, there are a lot of printStackTrace, which of course should go into some future error handling logging system, and it's a 50/50% change that there is an explicit argument for these calls. We should AT LEAST be consistent. 


```
./javatools_utils/src/test/java/org/xvm/util/SetTest.java:        e.printStackTrace(System.out);
Binary file ./javatools/build/classes/java/main/org/xvm/runtime/ServiceContext.class matches
Binary file ./javatools/build/classes/java/main/org/xvm/runtime/template/_native/web/xRTServer$RequestHandler.class matches
Binary file ./javatools/build/classes/java/main/org/xvm/runtime/Container.class matches
Binary file ./javatools/build/classes/java/main/org/xvm/javajit/JitConnector.class matches
Binary file ./javatools/build/classes/java/main/org/xvm/tool/Runner.class matches
Binary file ./javatools/build/classes/java/main/org/xvm/tool/Compiler.class matches
./javatools/src/main/java/org/xvm/runtime/ServiceContext.java:            e.printStackTrace(System.err);
./javatools/src/main/java/org/xvm/runtime/ServiceContext.java:                    e.printStackTrace(System.err);
./javatools/src/main/java/org/xvm/runtime/template/_native/web/xRTServer.java:            t.printStackTrace();
./javatools/src/main/java/org/xvm/runtime/template/_native/web/xRTServer.java:                e.printStackTrace();
./javatools/src/main/java/org/xvm/runtime/Container.java:                e.printStackTrace(System.err);
./javatools/src/main/java/org/xvm/javajit/JitConnector.java:            e.printStackTrace();
./javatools/src/main/java/org/xvm/javajit/JitConnector.java:                e.printStackTrace();
./javatools/src/main/java/org/xvm/tool/Runner.java:            e.printStackTrace();
./javatools/src/main/java/org/xvm/tool/Compiler.java:                    e.printStackTrace(System.err);

```

javatools/src/main/java/org/xvm/tool/LauncherOptions.java
```diff
@@ -0,0 +1,1387 @@
...
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.OptionGroup;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.commons.cli.help.HelpFormatter;
+import org.apache.commons.cli.help.TextHelpAppendable;
+
+import com.google.gson.GsonBuilder;
```
thegridman, line 26: Why do we need gson?

From what I can see it is only used in this class to save and restore options using json. Is this an actual requirement or used anywhere? Given previous comments about introducing third-party runtime dependencies it probably shouldn’t be here unless it is an actual requirement.

lagergren: Gson was added so that I can talk and marshall compilation state back and forth from the LSP server if it is done over the wire.  Starting out at the LSP skeleton mandates using existing wire protocol interfaces that have some kind of serialisation, even if running locally.  Gson was added so we have that mechanism.  The LSP and XVM cannot immediately call each other programatically in the LSP protocol implementations I looked at, but now we have a mechanisms to communicate state and config for the options so that we can launch our own on the other side with those options, support a compiler daemon and different things like that. If you don't think it belongs in the PR I can remove it.

thegridman: OK, I thought it might be something related to future LSP use but wanted to check

lagergren: Gson is another of those "top ten maven artifacts", and is small and widely used.  The very few dependencies that we have now are all popular, small, widely used, widely understood, industry standards, and even plugged into dependabot on the CI side to immediately yank if any security violations are discovered. 

However, I am not sure yet, this might move over to use protobuffer or some other wire protocol, I haven't had time to concentrate on how pluggability works in other LSPs, but if you think it is too early to have json support for the configurations, I can remove it and put it back later when I use it. 

I did also experiment with bridging XTC launching from the plugin through json, since that looked very reusable for the LSP, which would have divorced me from the need to use toCommandLine on a programatically populated options, but since that is still used for command line launchers, it doesn't matter overly much.

javatools/src/main/java/org/xvm/javajit/JitConnector.java
ggleyzer: I don't take IntelliJ inspections as a word from God that I need to follow unquestionably. If I think it's meaningless (like in this case) I do disable it. I understand it may annoy you to see a little squiggly some times, but I don't think it's a strong enough reason to change some one else's code. For example, IntelliJ often complains about some English that Cam wrote. I would never take its suggestions over the native speaker :-)

As to the consistency - I'm with you 100%. We need to work toward it; but not in a middle of already out of hands PR. Let's do a change dedicated to those

javatools/src/main/java/org/xvm/tool/Compiler.java
```diff
@@ -323,66 +295,61 @@ && isExplicitCompiledFile(outputLoc.getName())) {
      * @param repoLib  the library repository being used for compilation
      */
     protected void injectNativeTurtle(ModuleRepository repoLib) {
-        ModuleRepository repoBuild    = extractBuildRepo(repoLib);
-        ModuleStructure  moduleTurtle = repoBuild.loadModule(Constants.TURTLE_MODULE);
+        var repoBuild    = extractBuildRepo(repoLib);
```
thegridman: The variable declarations in this method seem to have been changed to `var` for no particular reason. I certainly do not want to start a multi-week debate on the use of `var` like we have with finals. But this sort of change in a PR just adds noise for the reviewers to have to look at.  I realise this particular class has been heavily changed and on the whole this method has not added too much to the review work load, but I'd prefer not to have changes to exiting code for no apparent benefit.

lagergren: The reason I put the var in there was that I wanted to get rid of this Java 1.0 style cast - I knew that if Javac would accept var, and removing the cast, was allowed to compile, then the compiler can prove that this never becomes a runtime error, which is a nice feeling with type safety, and it did.  Then, in order to make the code more readable and not just have a single var sticking out to prove that the cast was not necessary, I made the other ones vars to.  I have restored the original code, but this can actually be argued to be an improvement - proving that that loadModule is safe and that there if it passes the compile, it will never fail in runtime.   

The "correct" way to do that would be to use generics for the Component return type, but apparently we don't do that.  Anyway. Reverted.

thegridman:
After spending a couple of day going over the changes, I am going to approve this PR. I know we might have a couple of areas still being discussed but I do not think any are blockers for merging this PR.

PR #359, opened 2025-12-19
Json DB TxManager log file maintenance

thegridman:
This PR adds logic to the Json DB TxManager to archive old transaction log files.

### DIFF
--- a/lib_jsondb/src/main/x/jsondb/TxManager.x
+++ b/lib_jsondb/src/main/x/jsondb/TxManager.x
@@ -125,8 +125,17 @@ service TxManager<Schema extends RootSchema>(Catalog<Schema> catalog)
         this.statusFile = store.fileFor(root.path + "sys" + "txmgr.json");
 
         // Configuration from injectables
-        this.maxLogSize       = configureMaxLogSize();
-        this.cleanupThreshold = configureCleanupThreshold();
+        @Inject(ConfigMaxLogSize)       Int?      maxLogSize;
+        @Inject(ConfigCleanupThreshold) Int?      cleanupThreshold;
+        @Inject(ConfigMaxLogArchiveAge) Duration? maxLogArchiveAge;
+
+        this.maxLogSize       = maxLogSize       ?: DefaultMaxLogSize;
+        this.cleanupThreshold = cleanupThreshold ?: DefaultCleanupThreshold;
+        this.maxLogArchiveAge = maxLogArchiveAge ?: DefaultMaxLogArchiveAge;
+
+        validateMaxLogSize(this.maxLogSize);
+        validateCleanupThreshold(this.cleanupThreshold);
+        validateMaxLogArchiveAge(this.maxLogArchiveAge);
 
         // build the quick lookup information for the optional transactional "modifiers"
         // (validators, rectifiers, and distributors)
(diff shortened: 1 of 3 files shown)


lib_jsondb/src/main/x/jsondb/TxManager.x
```diff
@@ -3136,4 +3220,67 @@ service TxManager<Schema extends RootSchema>(Catalog<Schema> catalog)
...
+                        ;
+            return cleanupThreshold;
+        }
+        return DefaultCleanupThreshold;
+    }
+
+    // ----- testing -------------------------------------------------------------------------------
+
+    /**
+     * @return a copy of the current LogFileInfo array.
```
ggleyzer: We usually don't put a period at the end of @return line

lib_jsondb/src/main/x/jsondb/TxManager.x
```diff
@@ -166,6 +170,21 @@ service TxManager<Schema extends RootSchema>(Catalog<Schema> catalog)
 
     // ----- properties ----------------------------------------------------------------------------
 
+    /**
+     * The injection name prefix for all JSON DB TxManager configuration.
+     */
+    static String ConfigTxManagerPrefix = jsondb.ConfigPrefix + ".txManager";
```
ggleyzer: I wonder if `$"{jsondb.ConfigPrefix}.txManager"` reads a bit better

lib_jsondb/src/main/x/jsondb/TxManager.x
```diff
@@ -3136,4 +3220,67 @@ service TxManager<Schema extends RootSchema>(Catalog<Schema> catalog)
...
+        return DefaultMaxLogSize;
+    }
+
+    /**
+     * Determines the value to use for the cleanup threshold.
+     *
+     * If the value has been configured using an injectable the injected value is validated to be
+     * greater than zero.
+     *
+     * @return the configured cleanup threshold. or the default value if not configured
```
ggleyzer: Random '.' in the middle of @return line

lib_jsondb/src/main/x/jsondb/TxManager.x
```diff
@@ -3043,6 +3089,44 @@ service TxManager<Schema extends RootSchema>(Catalog<Schema> catalog)
...
+                if (info.name != LogFileName && info.txIds.isBelow(oldestTx)) {
+                    archived = True;
+                    File infoFile = sysDir.fileFor(info.name);
+                    if (infoFile.exists) {
+                        File archiveFile = archiveDir.fileFor(infoFile.name);
+                        if (archiveFile.exists) {
+                            // this should never happen
+                            archiveFile.delete();
+                        }
+                        infoFile.store.move(infoFile.path, archiveFile.path);
```
lagergren: This is pretty branchy and nested. In Java there are several ways to tighten down something like this and make it more stream style / temporary variable removal. Is there any way in ecstasy to prettify and shorten this flow and still make it more readable?

(Also there are probably things that I would have done here like guard ifs, early returns, avoid uncessary elses, but that is a personal preference, and your existing code is quite readable, just curious if there are way to reduce this in xtc and maintain readability mainly - nitpick)

PR #360, opened 2025-12-19
XUnit DB adding extensions to XUnit for writing database tests.

thegridman:
This PR adds extensions to XUnit to allow database testing.

### DIFF
--- a/javatools/src/main/java/org/xvm/tool/LauncherOptions.java
+++ b/javatools/src/main/java/org/xvm/tool/LauncherOptions.java
@@ -45,6 +45,9 @@ public abstract class LauncherOptions {
 
     private static final String PSEP = File.pathSeparator;
 
+    public static final String OPTION_XUNIT_OUT = "xunit-out";
+    public static final String ARG_XUNIT_OUT = "--" + OPTION_XUNIT_OUT;
+
     private static final Options COMMON_OPTIONS = new Options()
         .addOption(builder("d").longOpt("deduce").desc("Automatically deduce locations when possible").get())
         .addOption(builder("h").longOpt("help").desc("Display this help message").get())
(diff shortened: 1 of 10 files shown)


lib_xunit_db/src/main/x/xunit_db/DatabaseTest.x
```diff
@@ -0,0 +1,36 @@
...
+ * The test database configurations will be used by any tests in the annotated fixture and its
+ * subclasses unless overridden by another annotated fixture.
+ *
+ * @param scope        the scope of the database.
+ * @param templateDir  the path to a directory of files to copy to initialize the test database
+ * @param configs      an optional function that returns a database configurations for a given
+ *                     schema type
+ */
+annotation DatabaseTest(DbConfig.Scope  scope       = Shared,
+                        Directory?      templateDir = Null,
```
ggleyzer: Why is it call templateDir and not dataDir? Is this not how the user indicates the location of the data files?

lib_xunit_db/src/main/x/xunit_db.x
```diff
@@ -0,0 +1,15 @@
+/**
+ * The XUnit DB test framework provides functionality to test Ecstasy database applications.
+ *
+ * The XUnit DB module is a `TestEngineExtender` so that it is automatically loaded by the XUnit
+ * test framework when it is imported into a module.
+ */
```
ggleyzer, line 14: I think it would be useful to have a couple of paragraphs to explain how to use it in not the most trivial scenarios. For example, can I run a test against a "live" database?

lib_xunit_db/src/main/x/xunit_db/DbConfigProvider.x
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A provider of test database configurations.
+ */
+interface DbConfigProvider {
+    /**
+     * Return the `DbConfig` for the specified schema type.
+     *
+     * @return True iff this database is configured for the specified schema type.
+     * @return the `DbConfig` to use for the specified schema type
```
ggleyzer: we usually would say 
    * @return (conditional) the `DbConfig` to use for the specified schema type

PR #373, opened 2026-01-20
Auto-discover test modules and add transitive dependency resolution

lagergren:
## Summary
- Enable XTC plugin to automatically discover and run test modules from `src/test/x` without requiring explicit configuration
- Add transitive dependency resolution for `xtcModule` dependencies - projects now only need to declare direct imports
- Simplify XDK module build files by leveraging transitivity (significant reduction in boilerplate)
- Add support for running tests during XVM bootstrap build (with some limitations for fundamental modules)

### Plugin Improvements
- Module artifacts now depend on `compileXtc` instead of `build` task (fixes circular dependencies)
- Producer configurations extend consumer configurations, enabling transitive dependency resolution
- Separate consumer/producer attribute handling so test configs can consume main outputs
- `testXtc` task properly depends on all compile tasks

### XDK Dependency Simplifications
Leveraging the new transitivity, many modules have significantly fewer explicit dependencies:
| Module | Before | After | Transitives via |
|--------|--------|-------|-----------------|
| lib_webauth | 12 deps | 3 deps | web |
| lib_webcli | 10 deps | 3 deps | web |
| lib_xenia | 9 deps | 2 deps | web |
| lib_web | 8 deps | 4 deps | sec |
| lib_xunit_db | 7 deps | 3 deps | jsondb |
| lib_xml | 5 deps | 2 deps | convert |

### Test Source Set Support in Bootstrap
- Example test added to `lib_json` demonstrating the pattern
- Tests require `xtcModuleTest(libs.javatools.bridge)` and `xtcModuleTest(libs.xdk.xunit.engine)` for runtime
- Works for modules later in dependency chain; may have issues for fundamental modules

## Test plan
- [x] Verified in xtc-app-template with HelloWorldTest module
- [x] Tests are auto-discovered from `build/xtc/test/lib/*.xtc`
- [x] `./gradlew build` compiles main, compiles test (with correct dependency), and runs xunit tests
- [x] All XDK modules compile successfully with simplified dependencies
- [x] lib_json test passes during XVM bootstrap build


plugin/src/main/java/org/xtclang/plugin/XtcTestExtension.java
```diff
@@ -21,19 +22,4 @@ public interface XtcTestExtension extends XtcRuntimeExtension {
      * @return property controlling build failure on test failure
      */
```
lagergren, line 23: These operations already exists on the source set level through the SourceSet implementation, so if you want to include or exclude tests, that is the best place to do it right now.  If we also need runtime filtering, of course I can put that back. (Or finer granularity than a source code file of tests). 

We should modify xunit to also support negative tests - there are already provisions for negative/positive tests , and that would probably go here, though, but likely work on the source set include/exclude selection of files.

lagergren: Never mind. With an xtc unit test framework like xunit, we can just expect negative tests to fail (and they are hence positive tests)=. Tired.

lib_json/src/test/x/JsonTest.x
ggleyzer, line 1: I'm not sure it makes sense to add such a trivial test to a core module. Unless, of course, we plan to expand it later.

thegridman: Yes, I'd agree. There are a number of existing json tests in `manualTests/src/main/x/json`

lagergren: Well it is a POC style, but I wanted to add something to show that we can put JSON tests in the XDK libraries directly. I would like to have it there so we have something that verifies that these work on the way through the XDK build, and if it doesn't test anything useful, let's do another PR where we change that to doing something useful. I feel it should be there, or the XVM bootstrap testing will immediately bit rot and stop working.

lagergren: The whole idea is to prove that we do not need to put tests in the main source set in a manualTest project for the XDK. I think we should bring this in, and if necessary, convert the manual tests that explicitly test the lib_json project into its test source set incrementally instead.

PR #380, opened 2026-01-26
Fix workflow triggers to use file names instead of display names

lagergren:
## Summary

Fixes the CI failure on master where `gh workflow run "Publish Snapshots"` failed with:
```
could not find any workflows named Publish Snapshots
```

## Root Cause

The caching names changed, running a closed unsuccessful branch, trying to to use anchors to shorten the .github CI script code. Even though the branch was reverted and never pushed to master, (PR #378 closed without merge), GitHub's workflow name cache remained corrupted:

| Workflow File | Expected Name | GitHub's Cached Name |
|--------------|---------------|---------------------|
| `publish-snapshot.yml` | `Publish Snapshots` | `.github/workflows/publish-snapshot.yml` |
| `publish-docker.yml` | `Publish Docker Images` | `.github/workflows/publish-docker.yml` |
| `homebrew-update.yml` | `Update Homebrew` | `.github/workflows/homebrew-update.yml` |

## The Fix

Changed all `gh workflow run` commands to use **file names** instead of **display names**. This is more robust because:
- File names are stable and don't depend on GitHub's name cache
- Won't break if someone pushes invalid workflow syntax again

## Changes

- **commit.yml**: Use file names for triggering publish-snapshot, publish-docker, homebrew-update
- **RELEASE_PROCESS.md**: Use file names for prepare-release, promote-release commands
- **GITHUB_WORKFLOWS.md**: Use file names for prepare-release, validate-dependabot commands

## Test plan

- [ ] Merge this PR to master
- [ ] Verify the next master push successfully triggers publishing workflows

### DIFF
--- a/javatools_turtle/build.gradle.kts
+++ b/javatools_turtle/build.gradle.kts
@@ -2,15 +2,22 @@ import XdkDistribution.Companion.XDK_ARTIFACT_NAME_MACK_DIR
 import org.gradle.api.attributes.Category.CATEGORY_ATTRIBUTE
 import org.gradle.api.attributes.Category.LIBRARY
 import org.gradle.api.attributes.LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE
+import org.gradle.api.tasks.Copy
 
 plugins {
     alias(libs.plugins.xtc)
 }
 
+val processXtcResources by tasks.existing(Copy::class)
+
 val xdkTurtleProvider by configurations.registering {
     isCanBeResolved = false
     isCanBeConsumed = true
-    outgoing.artifact(tasks.processResources)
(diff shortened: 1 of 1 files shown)


javatools/src/main/java/org/xvm/asm/constants/IdentityConstant.java
ggleyzer, line 1: We have about 100 places in our code where we use "fall through" in switches; all commented with 
`// fall through`
Why did you choose to fix just three?

lagergren: I have had some rebase trouble with my working branches. I removed them when I noticed they were in the ci pr.

PR #403, opened 2026-03-05
Replace native tool deps (keytool/openssl/certbot) with pure Java APIs

lagergren:
## Summary
- Replaced all `ProcessBuilder` calls to `keytool`, `openssl`, and `certbot` in `xRTCertificateManager` with pure Java equivalents
- `keytool` → `java.security.KeyStore` API (JDK)
- `openssl` (RSA keygen, CSR generation, PEM→PKCS12 conversion) → BouncyCastle + JDK crypto
- `certbot` (ACME/Let's Encrypt protocol) → acme4j
- New versioned dependencies: **acme4j 3.5.1**, **BouncyCastle 1.82** (declared in `libs.versions.toml`)
- Extracted `KeyStoreOperations` utility class with no XVM runtime dependencies
- Added 13 unit tests covering all keystore operations
- Fat jar size increased from 5.7 MB → 15 MB (BouncyCastle is ~7.8 MB)

## Notes
- The `certbot`/`certbot-staging` provider names are preserved for backward compatibility
- TODO comments flag exception-swallowing patterns and lack of proper logging framework
- ACME challenge flow uses the same `.challenge` webroot directory as before

## Test plan
- [x] All 13 `KeyStoreOperationsTest` tests pass (self-signed cert, symmetric key, password, password change, key extraction, deletion)
- [x] Manual test with self-signed certificate provider
- [x] Manual test with Let's Encrypt staging provider (requires network + domain)

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
@@ -1,38 +1,44 @@
 package org.xvm.runtime.template._native.crypto;
 
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileWriter;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.OutputStream;
 
 import java.nio.file.Path;
 
+import java.security.GeneralSecurityException;
 import java.security.Key;
 import java.security.KeyStore;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
 
+import java.util.List;
 import java.util.concurrent.Callable;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.TimeUnit;
+
+import org.shredzone.acme4j.AccountBuilder;
+import org.shredzone.acme4j.Authorization;
+import org.shredzone.acme4j.Order;
+import org.shredzone.acme4j.Session;
+import org.shredzone.acme4j.Status;
+import org.shredzone.acme4j.challenge.Http01Challenge;
+import org.shredzone.acme4j.exception.AcmeException;
+import org.shredzone.acme4j.util.CSRBuilder;
+import org.shredzone.acme4j.util.KeyPairUtils;
 
 import org.xvm.asm.ClassStructure;
-import org.xvm.asm.ConstantPool;
 import org.xvm.asm.MethodStructure;
 import org.xvm.asm.Op;
 
 import org.xvm.asm.constants.TypeConstant;
 
-import org.xvm.runtime.ClassComposition;
 import org.xvm.runtime.Container;
 import org.xvm.runtime.Frame;
 import org.xvm.runtime.ObjectHandle;
 import org.xvm.runtime.ObjectHandle.ExceptionHandle;
 
 import org.xvm.runtime.template.xException;
-import org.xvm.runtime.template.xNullable;
 import org.xvm.runtime.template.xService;
 
 import org.xvm.runtime.template.collections.xArray;
@@ -52,6 +58,7 @@ import org.xvm.util.Handy;
  */
 public class xRTCertificateManager
         extends xService {
+
     public static xRTCertificateManager INSTANCE;
 
     public xRTCertificateManager(Container container, ClassStructure structure, boolean fInstance) {
@@ -70,6 +77,7 @@ public class xRTCertificateManager
         markNativeMethod("revokeCertificateImpl" , null, null);
         markNativeMethod("createSymmetricKeyImpl", null, null);
         markNativeMethod("createPasswordImpl"    , null, null);
+        markNativeMethod("changeStorePasswordImpl", null, null);
         markNativeMethod("extractKeyImpl"        , null, null);
 
         invalidateTypeInfo();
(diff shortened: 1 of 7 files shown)


thegridman (approved):

javatools/src/test/java/org/xvm/runtime/template/_native/crypto/KeyStoreOperationsTest.java
ggleyzer, line 1: I have not tested any of it with the platform; please hold on the merge

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
```diff
@@ -89,406 +97,379 @@ public TypeConstant getCanonicalType() {
...
-                            "--cert-name",  hName.getStringValue(),
-                            "--reason",     "unspecified"
-                          );
-                break;
-
-            default:
-                return xException.makeHandle(frame,
-                    "Unsupported certificate provider: " + hProvider.getStringValue());
+    private void pollAuthUntilValid(Authorization auth, String sDomain) throws Exception {
+        for (int i = 0; i < MAX_POLL_ATTEMPTS && auth.getStatus() != Status.VALID; i++) {
```
ggleyzer: This is a very questionable implementation. Sleeping for five seconds for something that may come back in a millisecond seems to be wrong. I realize that `Process.waitFor()` we are using now seems to use the same approach, but (i) they only sleep for 100 millis; (ii) jdk can improve or make it native at any time; (ii) one of the points of this exercise is to improve the current behavior.

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
```diff
@@ -89,406 +97,379 @@ public TypeConstant getCanonicalType() {
...
+     * Revoke a certificate using the ACME protocol via acme4j.
+     */
+    private void revokeWithAcme(String sStorePath, char[] achPwd, String sName, boolean fStaging)
+            throws Exception {
+        var keyStore = KeyStoreOperations.loadOrCreateKeyStore(sStorePath, achPwd);
+        var cert     = keyStore.getCertificate(sName);
+
+        if (cert instanceof X509Certificate x509Cert) {
+            var session = new Session(acmeServerUri(fStaging));
+            var accountKeyPair = KeyPairUtils.createKeyPair(2048);
```
ggleyzer: This seems to be wrong; the key pair we pass to the `org.shredzone.acme4j.Certificate.revoke()` API must be a key pair associated with the certificate being revoked, but a randomly generated one

ggleyzer (changes requested):
In addition to two reported issues I still didn't have time to test this in the real world, having to talk to the "real" CA authority. After the changes I requested are submitted I'm going to test it with the platform.

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
```diff
@@ -89,406 +97,379 @@ public TypeConstant getCanonicalType() {
...
-                    ? frameCaller.raiseException(xException.ioException(frameCaller,
-                        "Invalid or inaccessible key \"" + hName.getStringValue() + '"'))
-                    : frameCaller.assignValue(iReturn,
-                        xArray.makeByteArrayHandle(key.getEncoded(), Mutability.Constant));
+                        ? frameCaller.raiseException(xException.ioException(frameCaller,
+                                "Invalid or inaccessible key \"" + hName.getStringValue() + '"'))
+                        : frameCaller.assignValue(iReturn,
+                                xArray.makeByteArrayHandle(key.getEncoded(), Mutability.Constant));
             } catch (Throwable e) {
+                // TODO: catching Throwable and discarding the cause is bad practice; the
```
ggleyzer: To be fair we don't fully discard it; the message we log carries at least some information about it. Clearly we could improve it though

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
lagergren: Fixed in commit `89d4b800a` — replaced hand-rolled polling with acme4j's `waitForCompletion(Duration)` API. Both `order.waitForCompletion(ACME_TIMEOUT)` and `auth.waitForCompletion(ACME_TIMEOUT)` are now used with a 5-minute timeout. No more `Thread.sleep` loops.

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTCertificateManager.java
lagergren: Fixed in commit `b1d65ed3a` — revocation now extracts the domain private key from the keystore and uses it to authenticate the revocation request via `org.shredzone.acme4j.Certificate.revoke(Session, KeyPair, X509Certificate, ...)`. This is the correct approach per RFC 8555 §7.6 (domain-key-based revocation) and is actually more robust than the old certbot approach, since we always have the domain key in the keystore rather than depending on certbot's config directory.

lagergren:
What is the status here? When can we get a "real" CA test do you think?

ggleyzer:
It's on my list :)

lagergren:
So rebased and rerun - @ggleyzer did you have a chance to test what you wanted with the real certs?

ggleyzer:
Not yet; it's still in the middle of the task stack)

lagergren:
This actually work for a local x module I wrote + letsencrypt certs I created to host a small online game I wrote when I was 16, through a https connection on my home server (all in a docker container with nginx reverse proxy), so I am pretty sure it at least supports what we do with native tools at this stage. 

Michael also confirms that it works well in his tests in his deployment environment. 

It would be very good if we could close this off and get it merged ASAP, because if we wait longer, then I should probably modify our docker image recipe and brew recipe to install the native tools actually needed by the current ecstasy native implementation. 

So one more ping here with this info, @ggleyzer @cpurdy @mrcompator @thegridman

ggleyzer:
Working on it today

ggleyzer (approved):
Loo

PR #431, opened 2026-04-13
Java JIT project milestone

thegridman:
- Primitive Arrays
- Bit and Nibble support an corresponding arrays
- Fixed size initialised array support
- Range<Int constants
- Routing methods for caps
- Delegation support
- Map constants
- Various fixes


ggleyzer (approved):
Reviewed many times on-line

Issue #426, opened 2026-04-10
Stabilize JavaTools loading and avoid one-time XTC task classpath churn


lagergren:
## Summary

The Gradle plugin currently models JavaTools loading as a task (`loadJavaTools`) that mutates the Gradle daemon classloader. That appears to be the wrong architectural boundary.

The visible symptom is that after a clean main build, the first follow-up manual-tests invocation pays an unexpected one-time recompilation of `:manualTests:compileXtc` because Gradle sees the custom XTC task classpath as changed. This is not the XDK being rebuilt from scratch, and it does not appear to be a GitHub Actions cache-loss problem.

There is also ongoing overhead because `loadJavaTools` is explicitly never up to date.

A design note with more detail now lives in:

- `plugin/doc/plans/PLAN_JAVATOOLS_LOADING.md`

## Current architecture

Relevant code:

- `plugin/src/main/java/org/xtclang/plugin/tasks/XtcLoadJavaToolsTask.java`
- `plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java`

Today:

- `loadJavaTools` exists to make `javatools.jar` classes visible in the plugin/Gradle daemon classloader.
- All `XtcCompileTask`, `XtcRunTask`, and `XtcTestTask` depend on `loadJavaTools`.
- `loadJavaTools` is intentionally never up to date:

```java
getOutputs().upToDateWhen(_ -> false);
```

That makes some sense for a side-effect-only classloader mutation, but it does not fit Gradle's declarative task model well.

## Reproduction

Local reproduction used a CI-like sequence:

```bash
./gradlew clean --console=plain
./gradlew distZip check :plugin:compileJava \
  -Porg.xtclang.java.test.failFast=true \
  -Porg.xtclang.plugin.java.lint=true \
  --console=plain --info > /tmp/ci-build-info.log 2>&1
./gradlew manualTests:runXtc manualTests:runOne -PtestName=TestMisc \
  manualTests:runTwoTestsInSequence manualTests:runAllTestTasksParallel \
  --console=plain --info > /tmp/ci-manual-info.log 2>&1
./gradlew manualTests:runXtc manualTests:runOne -PtestName=TestMisc \
  manualTests:runTwoTestsInSequence manualTests:runAllTestTasksParallel \
  --console=plain --info > /tmp/ci-manual-info-2.log 2>&1
```

### The XDK is not the thing being rebuilt

[...]

PR #433, opened 2026-04-15
Harden plugin launcher runtime isolation and external consumer execution

lagergren:
## Summary

This PR finishes the first launcher-runtime hardening slice for the Gradle plugin.

The main goal is to keep the plugin usable in both of the important modes we care about:

- self-hosting builds where the XDK and plugin are built together in the composite build
- third-party consumers that resolve the plugin and XDK from Maven repositories, including local development via `publishLocal` and `../xtc-app-template`

The core design change from this branch is that the plugin no longer relies on daemon-global javatools loading or a pseudo-mutating `loadJavaTools` task. Instead, launcher runtime resolution is explicit, direct mode is isolated through a build-scoped runtime service, and forked execution uses a real resolved launcher runtime classpath.

This PR also adds a build-level execution-mode override so external consumers and alias tasks can switch the plugin defaults without needing task-local `--mode` handling.

This is related to #426, but I am not marking it as closing that issue because I do not want to claim that every historical symptom tied to that issue is fully proven closed by this PR alone.

### 1. Build-level execution mode override

Added a new Gradle property:

- `xtcDefaultExecutionMode`

Supported values are:

- `ATTACHED`
- `DETACHED`
- `DIRECT`

This property is now part of the launcher-task default convention, so plugin-created tasks inherit it automatically. That matters for real consumer builds, because alias tasks like `greet` do not own the `--mode` CLI option themselves.

This makes flows like the following work cleanly:

- `./gradlew greet -PxtcDefaultExecutionMode=DIRECT -PlocalOnly=true`

### 2. Direct-mode test execution fix

While doing the final verification pass, I found a real bug in direct-mode test execution for external consumers.

The isolated direct path was constructing `TestRunnerOptions` as if they accepted generic runner method semantics, which injected an unsupported `-M` shape into the test-runner option set. The result was that external-consumer `testXtc` in `DIRECT` mode failed even though compile/run direct-mode paths were working.

This PR fixes that by aligning the isolated direct test-runner option builder with the actual `TestRunner` option contract.

### 3. CI integration coverage for external consumers

The `commit.yml` integration-test job now exercises `xtc-app-template` in both modes:

- default attached mode
- direct mode via `-PxtcDefaultExecutionMode=DIRECT`

That keeps CI aligned with the supported consumer behaviors that were manually verified during this refactor.

### 4. Documentation / research notes

`launcher-runtime-brittleness-gpt.md` has been updated with:

- the current completed task list
- verification notes from self-hosting and external-consumer flows
- the build-level execution-mode override
- the direct-mode test-runner discovery/fix
- the next-step architecture note for a future reused external runtime / worker investigation

## Why This Refactor Exists

The old launcher architecture had a brittle combination of behaviors:

- daemon-global classloader mutation
- task setup that depended on pseudo-mutating `loadJavaTools`
- runtime resolution that was only partially modeled by Gradle
- direct mode behavior that could be poisoned by previous daemon state

That was especially risky when switching between:

- different builds
- different execution modes
- self-hosting and external-consumer scenarios
- republished local artifacts

The current architecture is better because:

- `loadJavaTools` is gone from active execution
- daemon-global javatools loader state is gone
- direct mode uses build-scoped isolation
- runtime reuse is keyed by runtime content fingerprint
- attached and direct paths now both work against the same explicit launcher runtime model
- external consumers can opt into direct mode at the build level without task-specific hacks

### Plugin and local verification

I verified the following locally during this slice:

- `./gradlew :plugin:test --console=plain`
- `./gradlew :manualTests:runTestAllExecutionModes --console=plain`
- `./gradlew :manualTests:runTestWithDirect --console=plain`
- `./gradlew :manualTests:runTestWithAttached --console=plain`
- `./gradlew :manualTests:runXunitTests --console=plain`

### Publish and external-consumer verification

I also verified the Maven Local consumer flow end to end:

- `./gradlew publishLocal --console=plain`

Then from `../xtc-app-template`:

- `./gradlew greet -PlocalOnly=true --refresh-dependencies --console=plain --info`
- `./gradlew greet -PxtcDefaultExecutionMode=DIRECT -PlocalOnly=true --refresh-dependencies --console=plain --info`
- `./gradlew greet -PlocalOnly=true --console=plain --info`
  - this confirmed that switching back from direct mode invalidates configuration cache and reruns the relevant tasks because `executionMode` changed
- `./gradlew testXtc -PxtcDefaultExecutionMode=DIRECT -PlocalOnly=true --refresh-dependencies --console=plain --info`
  - this was the case that exposed the direct-mode test-runner bug, and it now passes after the fix in this PR

The direct-mode external-consumer logs showed:

- the published `xtc-plugin` jar from Maven Local being used by the consumer build
- the published `xdk` zip being extracted from Maven Local
- direct compile/run execution in the current process
- build-scoped isolated runtime creation and reuse

### Known out-of-scope behavior

One known runtime/interpreter issue still exists in `manualTests:runParallel`, but that is not treated as a blocker for this PR because it is a separate interpreter/runtime race and not caused by the launcher-runtime hardening work in this branch.

## Scope Boundary

This PR is intended to merge the first safe functionality slice to `master`.

It is **not** the follow-up investigation into whether XDK aggregate builds can be made faster with a reused external runtime or worker model. That should be handled separately after this foundation is merged.


AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+- Do not use string-based untyped task wiring when typed providers are available.
+- Do not capture `project`, `logger`, or other script objects inside task actions.
+- Prefer Provider APIs, declared inputs/outputs, injected services, Worker API, or convention plugins.
+- Test Gradle changes with real tasks after editing.
+
+## Java/Kotlin Style
+
+- Always end files with a newline.
```
ggleyzer: What is the purpose of the "empty line" rule? We never do that.

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+- Never use star imports.
```
ggleyzer: Why such a strict rule? Our convention is to use "*" if a threshold (20) is reached

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+- Never use fully qualified Java type names in Java source when an import should be used.
+- Prefer `var` when the type is obvious from the right-hand side.
```
ggleyzer: I disagree with this; we should make things easier for the reader, not the writer. Why should the reader guess what the actual type is? Yes, sometimes it makes sense for a well known methods (e.g. Map.entrySet(), Set.iterator(), etc), but to make this into a hard rule is not right IMHO

ggleyzer (changes requested):
I very much like the idea of common AGENTS.md file with specific agents pointing it, but I disagree with some Java rules that you put there. Maybe we should do that change as a separate PR, asking every one's input and coming to a consensus of the coding style

AGENTS.md
ggleyzer: Everything else looks good though

lagergren: This has always been there. I have put the agents in a separate PR and if you could approve the gradle stuff so we can start integrating it it would be great.

AGENTS.md
lagergren: It is how I was told to write an AGENTS.md file. Different paragraphs tend to become different rules.

AGENTS.md
lagergren: We have always had these. I haven't changed a single electron.

ggleyzer (approved):

lagergren:
> I very much like the idea of common AGENTS.md file with specific agents pointing it, but I disagree with some Java rules that you put there. Maybe we should do that change as a separate PR, asking every one's input and coming to a consensus of the coding style

I actually have not changed anything from what has been committed in CLAUDE.md since for a very long time ago.  The xtc plugin was written using these prompting rules. So either way, I am not adding anything new here, and we can discuss contents later.

thegridman (approved):

PR #440, opened 2026-04-21
Fix xunit BaseTestRunner always returning success

lagergren:
## Summary

`BaseTestRunner.run()` constructed two separate `StatusListener` instances — one kept as a local and one inlined into `listeners.add(new StatusListener())`. Only the inline instance received execution events; the outer local (which was returned) stayed at its initial `success = True` forever.

Net effect: every `xtc test` exited 0 regardless of how many tests actually failed. The XTC Gradle plugin's `testXtc` relies on that exit code (`XtcRunTask.runSingleModule`'s `if (exitCode != 0)` → throw), so `./gradlew build` reported `BUILD SUCCESSFUL` even when the console output showed `ChessFooTest Failed (passed=X failed=Y skipped=0 errors=0)` for one or more modules.

Fix: reuse the outer `statusListener` in the listener composite so the instance the runner returns is the one actually driven by test-outcome events.

## Test plan

- [ ] From `../examples` checkout, rebuild the XDK from this branch and re-run `./gradlew :chess-game:app:testXtc` against the examples branch `chess-game-test` (which has intentionally `@Disabled` failing tests).
- [ ] With the current `@Disabled` set, the build should still succeed (no uncaught failures).
- [ ] Remove one `@Disabled` from `ChessMoveValidatorTest.x` and confirm the build now fails on that test.
- [ ] Restore `@Disabled` and confirm the build goes green again.

## Context

Discovered while diagnosing why `./gradlew build` in the `examples` repo printed `MoveValidatorTests Failed (passed=3 failed=1)` but still reported `BUILD SUCCESSFUL`. Root-cause walk-through: `XtcTestTask.executeTask` → `XtcRunTask.runSingleModule` checks exit code → exit code came from `xunit_engine.x:140 return t[0].as(Boolean) ? 0 : 1` → generated test module's `run` returns `engine.executor.runTests(m)` → `runTests` = `new console.ConsoleRunner().run(m)` → `BaseTestRunner.run` always returned `True`.

Also touches on a separate, latent issue: `XtcProjectDelegate.java:540` hard-codes `task.getFailOnTestFailure().convention(true)` which overrides the convention wired from the `xtcTest` extension in `XtcTestTask`'s constructor — so `xtcTest { failOnTestFailure = false }` wouldn't propagate either. Not fixed here; worth a follow-up if that DSL path is intended to work.

### DIFF
--- a/lib_xunit_engine/src/main/x/xunit_engine/executor/BaseTestRunner.x
+++ b/lib_xunit_engine/src/main/x/xunit_engine/executor/BaseTestRunner.x
@@ -15,7 +15,7 @@ import xunit.extensions.Extension;
         StatusListener            statusListener = new StatusListener();
         ExecutionListener[]       listeners      = new Array();
 
-        listeners.add(new StatusListener());
+        listeners.add(statusListener);
 
         ExecutionListener? runnerListener = createExecutionListener(testModule);
         if (runnerListener.is(ExecutionListener)) {
--- a/plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
+++ b/plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
@@ -537,7 +537,12 @@ public class XtcProjectDelegate {
         testTask.configure(task -> {
             task.setGroup(LifecycleBasePlugin.VERIFICATION_GROUP);
             task.setDescription("Run XTC xunit tests. Skip with -P" + PROPERTY_SKIP_TESTS + " or -P" + PROPERTY_SKIP_ALL_TESTS + ".");
-            task.getFailOnTestFailure().convention(true);
+            // NOTE: do not set task.getFailOnTestFailure().convention(...) here.
+            // The task's constructor already wires its convention to the
+            // xtcTest extension's failOnTestFailure property; setting another
+            // convention at registration time replaces that wiring, which
+            // breaks the `xtcTest { failOnTestFailure = ... }` DSL. The
+            // default value comes from DefaultXtcTestExtension (true).
             task.dependsOn(compileTaskNames);
             task.onlyIf(t -> !shouldSkip);
             if (shouldSkip) {


lagergren:
### Verified end-to-end against the examples repo

Ran `./gradlew publishLocal` on this branch (produced `~/.m2/repository/org/xtclang/xdk/0.4.4-SNAPSHOT/xdk-0.4.4-SNAPSHOT.zip` SHA `c0e0c263...`), enabled `mavenLocal` in the `xtclang/examples` repo's `settings.gradle.kts` + `build-logic/settings.gradle.kts` + `chess-game/settings.gradle.kts` + `webapp-conventions.gradle.kts` so the chess-game build picks up the fixed XDK, then temporarily removed one `@Disabled` from `chess-game/app/src/test/x/ChessPieceValidatorTest.x` (the intentionally-broken `shouldValidateBasicPieceMovement`) and ran:

```
./gradlew :chess-game:app:testXtc --refresh-dependencies
```

**Result with the fix applied:**

```
Finished: AIPositionEvaluatorTests Failed in 0.072 (passed=4 failed=1 skipped=0 errors=0)
Finished: ChessAIEvaluatorTest Failed in 0.072 (passed=4 failed=1 skipped=0 errors=0)
> Task :chess-game:app:testXtc FAILED
  > [plugin] Test failure.
BUILD FAILED in 24s
```

Before the fix (remote `0.4.4-SNAPSHOT` XDK), the exact same code / same failing test produced `BUILD SUCCESSFUL`. After the fix, the xunit engine returns a real exit code, `XtcRunTask.runSingleModule` throws on `exitCode != 0`, `XtcTestTask` re-throws because `failOnTestFailure = true`, and Gradle fails the build. Propagation works.

### Bonus finding

The fix surfaced a previously-hidden failing test that was silently ignored by the old `always-return-True` path:

```
Finished: AIPositionEvaluatorTests.shouldEvaluatePositionWithMaterialAdvantage() Failed in 0.009
IllegalState: "eval > 0": eval=-975
```

Unrelated to this PR — it's a real chess-evaluation bug in `AIPositionEvaluator.x` that was masked by the xunit bug. I'll track it separately in the `examples` repo; mentioning it here only as evidence that the runner is now actually reporting what the tests say.

lagergren:
Added follow-up commit `e5ac18b97` that fixes the second issue mentioned in the PR description.

**Context:** `XtcTestTask`'s constructor wires its `failOnTestFailure` convention to the `xtcTest` extension's property (`this.failOnTestFailure = objects.property(Boolean.class).convention(testExtension.getFailOnTestFailure());`). Then in `XtcProjectDelegate.configureTestTask(...)` we immediately replaced that wiring with `task.getFailOnTestFailure().convention(true);` — calling `convention()` a second time replaces the first, so the task convention became a hard-coded `true` detached from the extension.

That meant `xtcTest { failOnTestFailure = false }` in a consumer build had no effect: the DSL looked plumbed but the extension's value never reached the task.

**Fix:** drop the redundant `convention(true)` call in `XtcProjectDelegate` and rely on `DefaultXtcTestExtension`'s own `.convention(true)` (line 24) as the source of the default. Out-of-the-box behaviour is unchanged (default still `true`), but the DSL path is now live and overridable.

thegridman (approved):

PR #442, opened 2026-04-23
Auto-provision Gradle daemon JDK 25; rename XTC→Ecstasy in prose

lagergren:
## Summary

Two related changes on one branch:

1. **Gradle daemon JDK 25 auto-provisioning** (`e30a811ee`). Commits `gradle/gradle-daemon-jvm.properties` with per-platform Foojay redirect URLs, generated by `./gradlew updateDaemonJvm --jvm-version=25`. A contributor or CI runner with only a Gradle-minimum bootstrap JDK (17+) no longer needs JDK 25 installed separately — Gradle auto-provisions it on first launch. Matches the mechanism on the examples repo's `lagergren/examples-ci` branch.

2. **Rename "XTC language" → "Ecstasy" in user-visible prose** (`8f2d69b46`). The language is Ecstasy; XTC is the toolchain (CLI, Gradle plugin, IntelliJ plugin, tree-sitter grammar id, file extensions). Updates READMEs, plan docs, LSP server log messages, CLI banners, editor-support generator headers, and the Gradle plugin portal display name/description in `xdk.properties`.

   Identifiers kept as "xtc": the IntelliJ `notificationGroup` id "XTC Language Server" (referenced by `getNotificationGroup` lookups), the tree-sitter C symbol `tree_sitter_xtc`, and the Gradle SourceSet language name `XTC_LANGUAGE_NAME = "xtc"`.

   `lang/intellij-plugin/.../plugin.xml` intentionally not touched — user-facing display strings should be updated in a separate pass that preserves every `id="..."` attribute for code compatibility.

## Test plan

- [ ] CI passes (Verify Commit) on `lagergren/update-daemon`
- [ ] Fresh-machine sanity check: with only JDK 21 on PATH, `./gradlew build` auto-provisions JDK 25 via Foojay and completes the build
- [ ] Spot-check renamed log banner: `./gradlew :lang:lsp-server:run` (or any path that starts the LSP server) shows `Ecstasy Language Server v…` in the banner
- [ ] Spot-check generator output: regenerate editor support (`./gradlew :lang:dsl:generateAllEditorSupport`) and confirm the Emacs/Vim/Sublime syntax file headers now say "Generated from Ecstasy language model"


.devcontainer/README.md
```diff
@@ -12,16 +12,16 @@ This devcontainer provides a complete development environment for the XTC (Ecsta
...
-- **Java Runtime**: Optimized JRE for running XTC programs
+- **Java Runtime**: Optimized JRE for running Ecstasy programs
 - **Git**: Version control tools
 - **GitHub CLI**: For interacting with GitHub repositories
 
 ## Available Tools
 
-- `xtc` - XTC/Ecstasy compiler
-- `xec` - XTC/Ecstasy program executor  
+- `xtc` - Ecstasy compiler
```
cpurdy: xcc is the compiler tool
xtc is the "everything" tool

gradle/gradle-daemon-jvm.properties
```diff
@@ -0,0 +1,12 @@
+#This file is generated by updateDaemonJvm
```
cpurdy, line 1: Is this file supposed to be checked in to git?

lang/doc/plans/vscode-specific.md
```diff
@@ -468,12 +468,12 @@ Add a status bar item in `extension.ts` (~15 lines):
 ```typescript
 const statusBar = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, 100);
 statusBar.text = '$(gear~spin) XTC';
```
cpurdy, line 470: Should this (and the same a few lines down) also be changed "XTC" -> "Ecstasy"?

cpurdy (changes requested):
Looks good overall. Just a few questions I added comments for.

gradle/gradle-daemon-jvm.properties
lagergren: Yes.

lang/doc/plans/vscode-specific.md
lagergren: Yes, but not right now, since they carry programmatic semantic weight and I deliberately don't want to change the names of things like language implementation tags in plugin code in a mostly semantic change.

.devcontainer/README.md
lagergren: Good catch, fixing that.

gradle/gradle-daemon-jvm.properties
lagergren: At least if you want to be able to run "offline". The gradlew will use any system java 17+ to acquire a java 25 if there is none available, and if none is, foojay will resolve and download one from the latest repos. The repos should be checked into the code to provide locations to retrieve validated and compatible java 25s. 
Very easy to check. Build a dockerfile with java 17 or 21 as the system java and run gradlew a couple of times with and without offline settings. 

This doesn't change anything in the current build or for any build that has a java 25 installed already

lagergren: Or as Claude explains it, much better than I do: 

-

                                                                                                                      
❯ So is the convention that we check that update daemon properties thing into the code base                             
                                                                                                                      
⏺ Yes — gradle/gradle-daemon-jvm.properties is meant to be committed. It's part of the wrapper plumbing, just like      
  gradle-wrapper.properties and gradle-wrapper.jar.                                                                   
                                                                                                                        
  Reasoning:                                                                                                            
  - It's the deterministic output of ./gradlew updateDaemonJvm --jvm-version=N, regenerated only when you bump the      
  target JDK.                                                                                                           
  - Committing it pins the daemon-JVM contract for every contributor + CI: same version, same Foojay fallback URLs.   
  - Without it, Gradle falls back to ad-hoc auto-detect logic with no per-platform download URLs — fresh-machine builds 
  break unless the user pre-installs the right JDK.                                                                     
  - Gradle's own docs (and the file's #This file is generated by updateDaemonJvm header) treat it as source-controlled  
  config, not build output.                                                                                             
                                                                                                                        
  Convention summary:                                                                                  
  - ✅ Commit gradle/gradle-daemon-jvm.properties                                                                       
  - ✅ Regenerate with ./gradlew updateDaemonJvm --jvm-version=<N> when the JDK target changes; commit the result       
  - ❌ Don't .gitignore it                                                                                            
  - ❌ Don't hand-edit it (the task overwrites)

cpurdy (approved):
"Make it so, number one."

PR #444, opened 2026-04-25
Add disass --module-name-only flag

lagergren:
## Summary

- New `xtc disass --module-name-only` flag prints just the fully qualified module name (e.g. `web.xtclang.org`) of the target, then exits. Works against both an `.xtc` file path and a module name resolved against `-L`.
- Drive-by fix to a pre-existing bug in `Disassembler.validateOptions()`: it unconditionally rejected any target that did not exist as a file on disk, which made `disass <name> -L <path>` unreachable. Validation now only checks file existence for `.xtc` paths; bare module names fall through to the module-path repository as the code in `process()` always intended.

## Motivation

CI pipelines that publish a compiled module to a service which requires its FQN (e.g. registering a `@WebApp` module via `curl`) currently have to hardcode the qualified suffix as an env var, because nothing on the published `.xtc` byte stream is exposed via the CLI. With this flag a CI script can do:

```bash
FQN=$(xtc disass --module-name-only build/.../web.xtc)
curl ... --data-urlencode "module=$FQN" ...
```

## Test plan

- [x] `xtc disass --module-name-only <file>.xtc` prints the FQN for `net.xtc`, `web.xtc`, `aggregate.xtc`, `collections.xtc`
- [x] `xtc disass --module-name-only -L <lib> <module-name>` prints the FQN for `net.xtclang.org`, `web.xtclang.org`, `collections.xtclang.org`, `ecstasy.xtclang.org`
- [x] Nonexistent `.xtc` path → clean error, non-zero exit
- [x] Unknown module name on `-L` → clean error, non-zero exit
- [x] Regular `xtc disass` (no flag) still works for both `.xtc` paths and module names
- [x] `xtc disass --help` lists `--module-name-only` with description
- [x] Full `./gradlew clean` then `./gradlew build` succeeds

### DIFF
--- a/javatools/src/main/java/org/xvm/tool/Disassembler.java
+++ b/javatools/src/main/java/org/xvm/tool/Disassembler.java
@@ -18,6 +18,7 @@ import org.xvm.asm.Constant.Format;
 import org.xvm.asm.ErrorListener;
 import org.xvm.asm.FileStructure;
 import org.xvm.asm.MethodStructure;
+import org.xvm.asm.ModuleStructure;
 
 import org.xvm.asm.constants.FSNodeConstant;
 import org.xvm.asm.constants.StringConstant;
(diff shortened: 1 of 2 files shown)


ggleyzer (approved):
Looks good

PR #445, opened 2026-04-25
Elvis/vscode extension

ezekielelvis:
[WIP]This pull request introduces significant improvements to the VS Code extension for XTC, focusing on enhanced debugging support, improved task and command integration, and better developer experience through configuration and documentation updates. The main changes are the addition of a Debug Adapter Protocol (DAP) integration, new commands and tasks for running and managing XTC modules, and updates to build scripts and configuration files to support these features.

**Debugging and Task Integration**

* Added a new Debug Adapter Protocol (DAP) integration, including a `XtcDebugAdapterDescriptorFactory` and `XtcDebugConfigurationProvider` in `debug-adapter.ts`, enabling users to debug XTC modules directly from VS Code. This includes logic to find the Java executable, launch the DAP server JAR, and provide debug configuration defaults.
* Introduced a new task provider and command registrations in `commands.ts`, allowing users to run XTC modules, create new projects, and show language server output directly from the command palette.
* Updated `package.json` to declare new commands (`xtc.runModule`, `xtc.restartServer`, `xtc.showServerOutput`), a custom XTC task type, and a new debugger type (`xtc`) with configuration snippets and attributes for module debugging. [[1]](diffhunk://#diff-03fc9f87a7e7a8faf7d3a2a2d2d128e61f79b979a5d02938d729bffce83002f2L23-R25) [[2]](diffhunk://#diff-03fc9f87a7e7a8faf7d3a2a2d2d128e61f79b979a5d02938d729bffce83002f2R63-R132) [[3]](diffhunk://#diff-03fc9f87a7e7a8faf7d3a2a2d2d128e61f79b979a5d02938d729bffce83002f2R161-R223)

**Build and Launch Configuration**

* Updated the `build.gradle.kts` build script to copy the DAP server JAR into the extension's `server` directory, and ensured that both the LSP and DAP servers are included in packaging and assembly tasks. The VS Code launch command now opens the test fixtures folder for easier testing. [[1]](diffhunk://#diff-72d70a97ba83357b6138e2ff363a126c23cbdc56138c232a546b0e50ff4e96c3L44-R63) [[2]](diffhunk://#diff-72d70a97ba83357b6138e2ff363a126c23cbdc56138c232a546b0e50ff4e96c3L81-R92) [[3]](diffhunk://#diff-72d70a97ba83357b6138e2ff363a126c23cbdc56138c232a546b0e50ff4e96c3L95-R106) [[4]](diffhunk://#diff-72d70a97ba83357b6138e2ff363a126c23cbdc56138c232a546b0e50ff4e96c3L105-R117)
* Added and updated VS Code workspace configuration files (`.vscode/launch.json`, `.vscode/tasks.json`) to support extension development, building, and debugging workflows for both the root and extension subfolders. [[1]](diffhunk://#diff-bd5430ee7c51dc892a67b3f2829d1f5b6d223f0fd48b82322cfd45baf9f5e945R1-R18) [[2]](diffhunk://#diff-7d76d7533653c23b753fc7ce638cf64bdb5e419927d276af836d3a03fdf1745aR1-R20) [[3]](diffhunk://#diff-5dac5475c87df10e74302ae8a64c1894033bfd39ee2b9df7aa19787c39ecea40R1-R19) [[4]](diffhunk://#diff-84bab7d8b5b5fb2254bc413a6252a4a487171aa28ec704c4b84fe2591a8bcce7R1-R30)
* Updated `.gitignore` to exclude `.vscode-test/` from version control.

**Documentation and Developer Experience**

* Enhanced the `README.md` with clearer setup instructions, improved project structure documentation, and step-by-step guides for building, running, and testing the extension locally. [[1]](diffhunk://#diff-dbf543b17125723cbc24d85986f17cfe62d33fc477c76a029cec52fc3918f29aR54) [[2]](diffhunk://#diff-dbf543b17125723cbc24d85986f17cfe62d33fc477c76a029cec52fc3918f29aL64-R77) [[3]](diffhunk://#diff-dbf543b17125723cbc24d85986f17cfe62d33fc477c76a029cec52fc3918f29aL91-R99) [[4]](diffhunk://#diff-dbf543b17125723cbc24d85986f17cfe62d33fc477c76a029cec52fc3918f29aL100-R108) [[5]](diffhunk://#diff-dbf543b17125723cbc24d85986f17cfe62d33fc477c76a029cec52fc3918f29aL110-R130) [[6]](diffhunk://#diff-dbf543b17125723cbc24d85986f17cfe62d33fc477c76a029cec52fc3918f29aR156)

These changes collectively make the XTC VS Code extension much more robust and user-friendly, especially for debugging and automation workflows.


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Run XTC Extension",
+            "type": "extensionHost",
```
lagergren: Why are we adding a vscode catalog with configuration to root and to lang/vscode-extension? Is this intentional.  A template file for launch and tasks is probably a good idea but not sure we should check those into the either the repo root or lang root as this - aren't they strictly developer specific ones, or does the plugin need to use them ?

lagergren (changes requested):

javatools/bin/default/README.md
ggleyzer, line 1: It looks like you moved `javatools` to `javatools/bin/default` which is very strange. Was that an accident?

ggleyzer (dismissed):

.vscode/launch.json
lagergren: After our discussion these should be removed, or perhaps put under some resources dir so the build can still use and bootstrap them as needed but should not be in the root.

javatools/bin/default/.classpath
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path=""/>
```
lagergren: Delete this. It is some ide metadata isn't it?

javatools/bin/default/build.gradle.kts
```diff
@@ -0,0 +1,228 @@
...
+import org.gradle.api.tasks.PathSensitivity
+import java.util.Properties
+
+/**
+ * Build file for the Java tools portion of the XDK.
+ */
+
+plugins {
+    alias(libs.plugins.xdk.build.java)
+    alias(libs.plugins.palantir.git.version)
```
lagergren: This is a very strange place for a build file and I think this has to be some mistake?

javatools/bin/default/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
```
lagergren: You seem to have moved a number of files - including this LICENSE to bin default. Revert that

lang/vscode-extension/build.gradle.kts
```diff
@@ -78,7 +89,7 @@ val npmCompile by tasks.registering(Exec::class) {
 // Package the extension
 val packageExtension by tasks.registering(Exec::class) {
     description = "Package VS Code extension"
-    dependsOn(npmCompile, copyTextMateGrammar, copyLanguageConfig, copyIcon, copyLspServer)
+    dependsOn(npmCompile, copyTextMateGrammar, copyLanguageConfig, copyIcon, copyLspServer, copyDapServer)
     workingDir = layout.projectDirectory.asFile
     commandLine("npm", "run", "package")
```
lagergren, line 114: I don't think you should run npm from the command line. You should grab it through the gradle node plugin, pinned to a tested version in the libs.versions.toml and do it like e.g. the platform does it for Yarn. We do NOT want to create the dependency to any random binaries that may or may not be installed on the user's machine for a build. A build should always be self contained and retrieved artifacts at fixed versions for reproducibility and run-anywhere properties.

lang/vscode-extension/src/lsp-client.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+export function startLanguageClient(context: vscode.ExtensionContext, serverJar: string, outputChannel: vscode.OutputChannel): void {
+    const javaExecutable = findJavaExecutable();
+    const logLevel = process.env.XTC_LOG_LEVEL?.toUpperCase() ?? 'INFO';
+    const jvmArgs = buildJvmArgs(serverJar, logLevel);
+
+    outputChannel.appendLine('Starting XTC Language Server...');
+    outputChannel.appendLine(`Java: ${javaExecutable}`);
+    outputChannel.appendLine(`Args: ${jvmArgs.join(' ')}`);
```
lagergren, line 48: Is this typically how you get java in a vscode plugin? This looks like it assumes it has to be on the system path. Is there no way to have an extension depend on a java version and have it acquire it? I have't checked. But that is what Intellij plugins does, and probably the Disco API has some extension that can be inherited for acquiring a correct JDK/JRE in a vscode plugin. I hope

.vscode/launch.json
ezekielelvis: Done with this

ggleyzer (commented):
I'm really out of my depth here and would completely defer to Markus.

My only comment is: this PR seems to be an indicator that it wasn't the best idea to keep the plugin as a part of the xvm repo. May be we should start a new "plugin" repo using this PR?

lagergren:
> I'm really out of my depth here and would completely defer to Markus.
> 
> My only comment is: this PR seems to be an indicator that it wasn't the best idea to keep the plugin as a part of the xvm repo. May be we should start a new "plugin" repo using this PR?

There is nothing in the PR after updates that leaks out of lang, and as we have talked a lot about there is a migration plan to standalone implementation when I have finished the type adapter work, and there is a good reason that this lives under lang until that integration path is finished. 

I am still in deployment hell but I will try to review tonight or tomorrow.  Now, however, it looks safe, in that it doesn't bleed out state into the rest of the code and lives self contained under lang just like the intellij plugin and the lsp server that both use.  The only concern I have is about the node dependency potentially being resolved even if we are not creating a vscode-plugin build, and that, like the intellij toolkit, also is a very large download/install (although cached of course):

lagergren:
There is an issue with the CI / build and test trying to resolve the libs.versions.toml gradle plugin but fails. Not sure why, but I think that it needs to live only under lang an only when the vscode plugin is being built.

lagergren:
Where:
Build file '/home/runner/work/xvm/xvm/lang/vscode-extension/build.gradle.kts' line: 5

* What went wrong:
31 actionable tasks: 11 executed, 2 from cache, 18 up-to-date
Script compilation error:

  Line 5:     alias(libs.plugins.langNodeGradle)
                                 ^ Unresolved reference 'langNodeGradle'.

1 error

lang/vscode-extension/build.gradle.kts
```diff
@@ -69,9 +69,17 @@ val copyDapServer by tasks.registering(Copy::class) {
     rename { "dap-server.jar" }
 }
 
+// Copy LICENSE from repository root
+val copyLicense by tasks.registering(Copy::class) {
+    description = "Copy LICENSE from repository root"
+    from(layout.projectDirectory.file("../../LICENSE.md"))
+    into(layout.projectDirectory)
```
lagergren, line 71: Don't use relative paths in the composite build. There is a compositeRoot property in the build-logic.  We don't know where this code will live, and it is unsafe to assume that you can find a repo root with a license file two directories up.

ezekielelvis: Alright on it

lagergren:
The issue is that one of x files no longer parse equivalently between the lsp parser and the Java parser. This is likely an issue in treesitter caused by someone changing parser behavior or grammar in master, but not in the treesitter grammar template. It hasn't got anything to do with your branch. If you want, add the falling file to the skip list documented as a known outage and we can approve this for merge. 

The treesitter grammar validation tests only run for a branch that changes stuff in the Lang tree. This means that I think the test would fail if running in master itself. I am on the road most of the day, but if you can confirm thst validateTreeSitter grammar fails in master too with Lang attached, then the theory is confirmed. You can check the commit.yml to see how to run the validation test, or you could just trigger a GitHub workflow on master with workflow dispatch (manually) with the "test lang" boxes checked and it should do it automatically for you. 

I will take a look in a few hours when I return to civilization.

lagergren:
**Why the tree-sitter validation went red on this PR but is green on master**

Short version: this isn't a tooling or grammar regression — it's two new XTC syntax patterns that were added to the `manualTests` corpus between May 6 and May 17 and that the tree-sitter grammar never had test cases for. The path-filter gate in `.github/workflows/commit.yml` (`dorny/paths-filter` on `lang/**`) hid both of them from master CI because the PRs that introduced them only touched `manualTests/` and `javatools/`.

| Date | Commit | Test addition |
|------|--------|---------------|
| 2026-05-06 | `0120712a1` *Fix NameExpression code production for parameterized virtual child classes* | `generics.x:27` — `Type t2 = Base<Int>.Child2<String>;` (value-position virtual-child type with generic args) |
| 2026-05-17 | `91f6d7d96` *Fix a compiler exception (#463)* | `annos.x:141` — `annotation PreComputed(...) into Predicate { ... }` *inside a method body* (local annotation declaration) |

Both PRs were compiler bugfixes that added a single new line of XTC to verify the fix. Because neither touched `lang/**`, the tree-sitter validation step on master was skipped (path-filter conclusion: `lang=false`). On this PR, the `lang/vscode-extension/**` changes trip the same filter (`lang=true`), so the validation step runs — and surfaces both pre-existing latent failures.

The skip lists (`treeSitterParseSkipList.parked.txt`, `treeSitterParseSkipList.intentional.txt`) and the grammar template itself were unchanged between PR #452 (May 1, last grammar work) and the May 6 / May 17 corpus additions. The comment in `parked.txt` claiming "generics.x parses cleanly now — removed from this list" is incidentally misleading: `git log -S "manualTests/src/main/x/generics.x" -- treeSitterParseSkipList.parked.txt` returns empty, so generics.x was never actually in any skip list.

**Fix in this PR**: two grammar template edits — `annotation_declaration` added to `_statement`, and `member_type` (with-type-arguments form at `prec.left(MAX_PRECEDENCE_PLUS_3)`) added to `_expression`. Local test now reports `881 passed, 0 failed` (was 879 / 2).

**Suggested follow-up (not in this PR)**: the path-filter for the tree-sitter step in `commit.yml` could also include `manualTests/**/*.x` so that new corpus shapes get grammar-checked on master, not just on PRs that happen to touch `lang/**`. That would have caught both regressions at the source.

lagergren (approved):

PR #447, opened 2026-04-27
Replace LinkedIterator with Stream.concat (Work with the Java type system instead of bypassing it)

lagergren:
`LinkedIterator` was an internal utility class whose entire job was to splice two iterators end-to-end so that `getContained()` overrides could chain a parent's children with their own annotations. It had been in the tree since the initial commit in 2016, predating widespread use of `Stream.concat`.

This PR retires the class, replaces it with `Stream.concat(a, b).iterator()`, and follows through on the deeper design point by changing the `getContained()` API itself from `Iterator` to `Iterable` — which is what makes the rest of the cleanup possible.

## Where the original design went wrong

A few things piled up:

- **The constructor had an avoidable type-system trap.** `LinkedIterator(Iterator<E>...)` — without `? extends` — couldn't accept `Iterator<Component>` and `Iterator<Annotation>` together as an `Iterator<XvmStructure>`, even though both are `XvmStructure` subtypes. Java generics are invariant, so the right signature was `Iterator<? extends E>...`. One missing keyword was enough to push every call site into raw-type usage. With that single character fix, the class would have been entirely type-safe; instead, four raw call sites became the path of least resistance.

- **`@SafeVarargs` on a generic-array varargs constructor is always a smell** — it suppresses heap-pollution warnings, which is exactly what the four raw call sites were generating. The annotation should have been a signal that the API contract wasn't expressing the intent properly.

- **The class mutated its caller's varargs array** (`f_aIter[i] = null`, for GC hygiene). Documented in a comment, but action-at-a-distance on a shared input array. The lazy-iteration goal was reasonable; the implementation didn't need to scribble on the input to achieve it.

- **`remove()` was a footgun.** `getContained()` is documented "treat the return as immutable", but `LinkedIterator.remove()` happily forwarded to the underlying iterator. A documented read-only API leaking write capability — the iterator should have unconditionally thrown `UnsupportedOperationException` to match the contract.

- **It's 150 lines of hand-rolled state machine** (`NOT_READY` / `HAS_NEXT` / `CAN_REMOVE`) reimplementing what `Stream.concat` does in one line.

None of these individually are crimes — together they signal that the type system was being worked around rather than leveraged.

## The deeper design point — and why fixing it pays off

The reason `LinkedIterator` needed to exist at all is that `XvmStructure.getContained()` returned `Iterator`, not `Iterable` or `Stream`. Iterators are stateful, single-use, and don't compose — there's no `Iterator.concat`, no `for-each` over them, no native pipeline. The moment that's the return type, you immediately need a tool to combine iterators, and `LinkedIterator` is what filled the gap.

Switching the contract to `Iterable<? extends XvmStructure>` collapses a surprising amount of complexity:

- **`Component.getContained()` becomes a one-liner: `return children();`** Because `Collection extends Iterable`, the previous `.iterator()` indirection just falls away.
- **Seven `for (Iterator iter = getContained(); iter.hasNext(); )` loops** in `XvmStructure` rewrite to `forEach` / `anyMatch` / `allMatch` — `Iterable` brings `forEach` for free, and the early-exit predicates want stream operators anyway.
- **`ConstantPool.getContained()` collapses from an 18-line anonymous `Iterator` to `return List.copyOf(m_listConst);`** — same snapshot semantics, drastically less code.
- **The three Component subclass overrides** (`MethodStructure`, `ClassStructure`, `PropertyStructure`) DRY up via a small `Component.containedWith(Iterable extras)` helper. Each override now reads `cond ? super.getContained() : containedWith(extras)`. The `Stream.concat` + lambda boilerplate lives in one place.
- **`Handy.iteratorStream` becomes orphaned** — gone — replaced by the more useful `Handy.stream(Iterable)`.
- Several `java.util.Iterator` imports drop out of the affected files.

Net effect across the PR: **−295 / +138 = ~−157 lines**, one fewer class, no new warnings.

## Behavioral note: `ConstantPool.getContained()` semantics

The old code captured a *reference* to `m_listConst` when `getContained()` was called and indexed into it — surviving field reassignment but vulnerable to in-place mutation during iteration (which would have been undefined behavior anyway). The new code uses `List.copyOf(m_listConst)`, which snapshots the contents — strictly more protective. The only behavioral delta would matter if something mutated the constant pool's list while another caller was mid-iteration, which shouldn't happen, but the new behavior is the safer default.

## Tests

- Full `./gradlew clean && ./gradlew build` (242 tasks, including unit tests and the XTC test suite that exercises XVM file assembly/disassembly) — green.
- CI commit.yml manual-test commands (`manualTests:runXtc`, `runOne -PtestName=TestMisc`, `runTwoTestsInSequence`, `runCiTestTasksParallel`) — green.
- Four new tests added in `HandyTest` covering `Handy.stream(Iterable)`, including the load-bearing one — `testStreamFromLazyIterable` — that verifies a lazy `() -> Stream...iterator()` Iterable can be iterated multiple times producing the same sequence each time. That's the contract `Component.containedWith` depends on.

## The incremental path forward

There's plenty more raw-type and wildcard cleanup we could do — `Component.unlinkSibling`'s raw `Map` parameter, `Context.discardBranch`'s raw `Map` local, `SwitchAST.contents()` returning a raw `Iterator` from a public API — but each of those wants its own narrow PR with its own thinking. The point isn't to refactor aggressively; it's to gently retire the bypasses one at a time, so the type system stops being something to work around and starts being something we can lean on.

## Commits

1. **Replace `LinkedIterator` with `Stream.concat`** — main change, deletes the class, adds `Handy.iteratorStream` helper.
2. **`Set.of` for raw `HashSet` equality check** in `MethodStructure.reorderAnnotations`.
3. **`var` for raw `Map` locals** in `ClassStructure`'s equality check.
4. **Tighten `iteratorStream` variance and drop redundant `Stream.concat` witness**.
5. **Switch `getContained()` to return `Iterable`** — adds `Component.containedWith` helper, collapses the for-loops in `XvmStructure` to stream/forEach, simplifies `ConstantPool.getContained` to `List.copyOf`, replaces `Handy.iteratorStream` with `Handy.stream(Iterable)`.
6. **Add `Handy.stream(Iterable)` test cases** covering empty input, content/order preservation, lazy-Iterable multi-iteration, and the live (non-snapshot) behavior of the helper.


javatools/src/main/java/org/xvm/asm/MethodStructure.java
```diff
@@ -240,8 +239,7 @@ public Annotation[] getAnnotations() {
      * Replace the annotations with an equivalent re-ordered array.
      */
     public void reorderAnnotations(Annotation[] annotations) {
-        assert new HashSet(Arrays.asList(annotations)).equals(
-               new HashSet(Arrays.asList(m_aAnnotations)));
+        assert Set.of(annotations).equals(Set.of(m_aAnnotations));
```
cpurdy: This may be doing the same thing under the hood (or not), but the point of the "new HashSet" was that these objects need to be tested both for equality but also for hashing. (Less important today, arguably, now that the bugs are largely worked out, but a reasonable check to avoid deep regressions.)

javatools/src/main/java/org/xvm/asm/ConstantPool.java
```diff
@@ -2406,25 +2404,8 @@ public ConstantPool getConstantPool() {
...
-            public XvmStructure next() {
-                if (iNext >= listConst.size()) {
-                    throw new NoSuchElementException();
-                }
-
-                return listConst.get(iNext++);
-            }
-        };
+    public Iterable<? extends XvmStructure> getContained() {
+        return List.copyOf(m_listConst);
```
cpurdy: Just for the record: copies aren't free. A lot of the choices we made were to avoid needless copies. Were the data structures immutable or lazy immutable, we wouldn't need the copies, but that is a larger change, and it's not clear that the benefit at this point is worth it.

javatools/src/main/java/org/xvm/asm/ClassStructure.java
```diff
@@ -3604,8 +3594,8 @@ public boolean equals(Object obj) {
         }
 
         // type parameters
-        Map mapThisParams = this.m_mapParams;
-        Map mapThatParams = that.m_mapParams;
+        var mapThisParams = this.m_mapParams;
```
cpurdy: I think replacing the Map type here with var is a change for the worse. OTOH the residual hungarian notation ...

javatools/src/main/java/org/xvm/asm/ClassStructure.java
```diff
@@ -3354,24 +3353,15 @@ protected void assemble(DataOutput out)
     }
 
     @Override
-    public Iterator<? extends XvmStructure> getContained() {
+    public Iterable<? extends XvmStructure> getContained() {
```
cpurdy, line 3352: In general, I like the result of this type of change. Here we see that no captures are occurring, no copies are occurring, etc., so while the heap churn will increase by a small amount, the result is still quite arguably a net improvement.

javatools/src/main/java/org/xvm/asm/PropertyStructure.java
```diff
@@ -733,24 +731,15 @@ protected void assemble(DataOutput out)
...
+                .filter(c -> c.getComposition() == Composition.Annotation)
+                .map(Contribution::getAnnotation)
+                .toList();
 
-        return listAnno == null
-                ? super.getContained()
-                : new LinkedIterator(
-                        super.getContained(),
-                        listAnno.iterator());
+        return listAnno.isEmpty() ? super.getContained() : containedWith(listAnno);
```
cpurdy, line 742: This isEmpty call may hide an eval of the filter. (It would have in the original streams impl.) It's one of the reasons I still hesitate to reach for streams blindly. OTOH, the old code was just flat out doing a full eval up front, but at least it was an obvious cost.

Just to be clear, I'm not suggesting that this change is wrong. I'm more just thinking aloud about how we should/shouldn't represent/hide cost models. Not expecting a clear cut answer.

cpurdy (changes requested):
Only minor changes indicated (see comments). Looks good overall. I'm not convinced that this is the best investment of cycles right now (including reviews etc.), but I also recognize that ratcheting up the old code is something that (carefully done) pays dividends over time.

javatools/src/main/java/org/xvm/asm/MethodStructure.java
lagergren: Yes. Which every set does - there is no set without equals and hashcode semantics, or collection for that matter, unless you deliberate override and break the contract, so this is the preferred idiom.

javatools/src/main/java/org/xvm/asm/ClassStructure.java
lagergren: It is a change for the better since you use a raw type which messes up code inspection and produces (rightly so) yellow warnings. It is clear that this is a map from the function name, and it now compiles without raw type warnings. It is preferable to use a generic wildcard, or if you must, use a suppression for raw types, but it is not preferable to ever leave raw types in Java after 2004.

javatools/src/main/java/org/xvm/asm/ConstantPool.java
lagergren: Agreed. Switching defensive copies to `Collections.unmodifiableList(m_listConst)` — that's O(1) per call (one small wrapper, no array copy, no Object[N] allocation) instead of O(N), which MAY matter here because the constant pool is the largest list in the structure and `getContained()` is on the hot path for every full traversal (`registerConstants`, `validate`, `isModified`, `resetModified`, `resolve`, `purgeCondition`).

The contract on `XvmStructure.getContained()` already says "the caller should treat the return value as if it were immutable", and `Component.getContained()` returns `children()`, a live `values()` view, with no defensive copy. So as long as we trust the contract, we could equally well just return `m_listConst` directly (it's `final`); the `unmodifiableList` wrapper is the same cost class and adds a belt-and-suspenders runtime check against accidental mutation. Went with the wrapper for that reason — happy to drop to the bare field if you'd prefer to match `Component` exactly.

(There isn't a more modern Java 9+ replacement for the non-copying view — `List.of` and `List.copyOf` both copy. `Collections.unmodifiableList` is still the right tool.)

javatools/src/main/java/org/xvm/asm/ClassStructure.java
lagergren: There is no conceivable way the difference in references required will matter either way for footprint or performance, and it is also Java idiomatic, since an Iterator needs to implement remove, which may or may not be allowed, whereas the API contract is now exactly what we do want to provide, so I think it is a good idea to stick with this change.

javatools/src/main/java/org/xvm/asm/PropertyStructure.java
lagergren: The good news for this specific call site: `isEmpty()` here *can't* hide a filter eval, because `.toList()` is the terminal op. By the time the ternary runs, the stream is fully evaluated and `listAnno` is a concrete `List<Annotation>`, so `listAnno.isEmpty()` is just `List.isEmpty()` — O(1).

Cost shape vs. the old code:

- **Old:** `for` walks all contributions, lazily allocates an `ArrayList` only if any annotation is found, then `LinkedIterator` lazily concats during iteration.
- **New:** stream walks all contributions, `.toList()` materializes the list, then `containedWith(listAnno)` returns `() -> Stream.concat(...).iterator()` that lazily concats during iteration.

Both eagerly compute the annotation list. Both lazily concat at iteration. The lazy/eager boundary hasn't moved — the only real delta is short-lived stream-pipeline objects vs. the `for` loop's branch/null check, which HotSpot doesn't lose sleep over.

Your broader concern absolutely stands, though, and I share it. The place laziness leaks into apparent simplicity is when you inspect a *non-terminated* stream (`.findFirst().isPresent()`, `.count()` over a filter, etc.). The rule I see and keep is "always materialize before you inspect" — which is what the explicit `.toList()` here is doing, and I'd argue it actually makes the cost model *more* visible than the old `listAnno == null` sentinel did, not less: the eager pass is now a single named operation on the chain rather than spread across a four-line `for` body.

javatools/src/main/java/org/xvm/asm/ClassStructure.java
```diff
@@ -3354,24 +3353,15 @@ protected void assemble(DataOutput out)
...
         // we cannot use the "collectAnnotations" API at this time, since our module may not yet
         // be linked
-        List<Annotation> listAnno = null;
-        for (Contribution contrib : getContributionsAsList()) {
-            if (contrib.getComposition() == Composition.Annotation) {
-                if (listAnno == null) {
-                    listAnno = new ArrayList<>();
-                }
-                listAnno.add(contrib.getAnnotation());
```
lagergren, line 3362: Btw @cpurdy This pattern "null collection initialized in a loop if null", is in probably 100 places. It is a false optimization and there is absolutely nothing to be gained from not creating an empty collection before a loop that runs 0 to n iterations. It breaks streamable fluent ops and it introduces scaffolding logic everywhere, which makes it had to read the actual semantics of the xtc implementation. I would strongly suggest you start incrementally removing this when you are working with code that contains it.  In some places, I think only the parser, there is a difference between "absent" and "empty" so null and empty lists actually mean different things. The GC overhead is zero, you are just messing with the branch predictor (even though the overhead for that is zero too), and breaking opportunities to write chained fluent transforms that would be way easier to read.

javatools/src/main/java/org/xvm/asm/ConstantPool.java
lagergren: ( However, even this is a constant pool entry table of tens of thousands of entries, that ref copy is not going to matter much anyway, in the state of things. Most of the complexity is doing deep equals on type info, and you are (horribly) using _clones_ for object AST node copies, so there is overhead that is way more significant for constant pool and type info handling than a defensive shallow copy, even if the list is wide. I would be surprised if any measurable performance regression came out of the defensive copy approach )

javatools/src/main/java/org/xvm/asm/MethodStructure.java
lagergren: One more comment on this:  I have confirmed that the construct is semantically equivalent in any setting, including where we are doing brittle stuff like computing a hashCode lazily, which is a very dangerous thing to do in Java, since every module author or library provider will assume that if hashCode is overridden then that is final, and will stay the same as long as an object is equals to its last equality contents. This is typically architectured as a copy on write instead of mutable object field modification that may cause hash codes to change, if the computation is heavy or there is another good reason to want to compute it "late" from some "0" value. 

Anyway, more generically for Java in: 

* Set.equals is implementation-independent. Both HashSet and the immutable set returned by Set.of(...) honor the Set.equals contract from java.util.Set (same size, each element of one contained in the other), which delegates to the elements'        
  equals/hashCode. The internal storage strategy (open-addressed table in ImmutableCollections.SetN vs. HashMap-backed HashSet) is irrelevant to equals ** both will ALWAYS produce the same result for the same elements.                                          

And here for Annotation instances specifically: 
                                                                                                                                                                                                                                                            
* Annotation has standard equals/hashCode. Annotation extends Constant (javatools/src/main/java/org/xvm/asm/Annotation.java:37-38). Constant defines equals (Constant.java:598) and hashCode (Constant.java:560) at the base class level — equals        
  resolves both sides and compares format + compareDetails, hashCode is cached and derived from computeHashCode(). They're consistent with each other and follow the standard contract. There is no per-subclass override that violates                   
  symmetry/transitivity in this hierarchy.                                                                                                                                                                                                                  
                                                                                               
  The one real semantic difference (which your commit message already calls out and which is desirable in an assertion):                                                                                                                                    
  
  ┌───────────────────┬─────────────────────────────────┬─────────────────────────────────┐                                                                                                                                                                 
  │                   │ new HashSet(Arrays.asList(arr)) │           Set.of(arr)           │    
  ├───────────────────┼─────────────────────────────────┼─────────────────────────────────┤                                                                                                                                                                 
  │ null element      │ silently allowed                │ throws NullPointerException     │    
  ├───────────────────┼─────────────────────────────────┼─────────────────────────────────┤                                                                                                                                                               
  │ duplicate element │ silently deduped                │ throws IllegalArgumentException │                                                                                                                                                                 
  └───────────────────┴─────────────────────────────────┴─────────────────────────────────┘                                                                                                                                                                 
                                                                                                                                                                                                                                                            
  So the old form would silently pass an assertion where the array contained nulls or had a duplicate present in both arrays in the same pattern; the new form fails fast. Both of those would be real bugs in the reorder contract, so the stricter        
  behavior is the right call. cpurdy's hashing concern is also covered — Set.of's storage exercises hashCode just as much as HashSet does.                                                                                                                
****

lagergren:
@cpurdy So can we merge this with the potentially large copy behavior reverted then?

lagergren:
> Only minor changes indicated (see comments). Looks good overall. I'm not convinced that this is the best investment of cycles right now (including reviews etc.), but I also recognize that ratcheting up the old code is something that (carefully done) pays dividends over time.

We have spent the time already - a month ago. So it's not a lot of work to do if it can come in.

lagergren:
Pushed `f1496d277` as a review follow-up.

Changes made:
- restored the `HashSet<>(Arrays.asList(...))` assertion shape in `MethodStructure.reorderAnnotations()` to keep the explicit hashing/equality check
- replaced the `var` locals in `ClassStructure.equals()` with explicit generic `Map<StringConstant, TypeConstant>` locals

The earlier constant-pool copy concern is already addressed on this branch by the index-based live iterator, not a `List.copyOf`/snapshot. Verified with `./gradlew :javatools:compileJava :javatools_utils:test --console=plain`.

@cpurdy could you re-review when you have a chance? The PR is still blocked on the old Changes Requested review even though the threads are resolved.

lagergren:
One correction on the `HashSet` assertion point, because the official Java docs settle the semantic part.

The predicate being asserted here is set equality. The `Set` spec defines equality as same size plus same members and explicitly says this works across different `Set` implementations. The same interface defines set `hashCode()` as the sum of element hash codes, preserving the equal-set hash contract. `HashSet` is simply one implementation of `Set` backed by a hash table. `Set.of(E...)` also returns a `Set`, rejects duplicate elements, and rejects nulls at creation.

So for valid annotation arrays, meaning non-null and duplicate-free, this:

```java
new HashSet<>(Arrays.asList(a)).equals(new HashSet<>(Arrays.asList(b)))
```

has no additional specification-level meaning over this:

```java
Set.of(a).equals(Set.of(b))
```

Both assert the same `Set.equals` contract. The `HashSet` spelling chooses a mutable implementation and extra allocation; it does not assert a stronger semantic property. The only real semantic difference is that `Set.of(...)` is stricter: nulls fail with `NullPointerException` and duplicates fail with `IllegalArgumentException`, which is better for this assertion because either case would be a malformed annotation reorder input.

For idiom, the Oracle core-libraries guide specifically shows the old JDK 8 pattern using `new HashSet<>(Arrays.asList("a", "b", "c"))` plus `Collections.unmodifiableSet(...)`, and the JDK 9-and-later replacement as `Set.of("a", "b", "c")`. In 2026 Java, `Set.of(...)` is the intended factory for this assertion-local immutable set construction.

Official docs:

- Java SE 26 `Set.equals`: https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/Set.html#equals(java.lang.Object)
- Java SE 26 `Set.hashCode`: https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/Set.html#hashCode()
- Java SE 26 `Set.of(E...)`: https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/Set.html#of(E...)
- Java SE 26 `HashSet`: https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/HashSet.html
- Oracle core-libraries guide, unmodifiable set factory methods: https://docs.oracle.com/en/java/javase/26/core/creating-immutable-lists-sets-and-maps.html#GUID-CAC7ED7C-A9DD-4D30-9EB5-5C2C95AC21D1

cpurdy (approved):
Looks fine. Gene also looked over my shoulder a bit while I was going through the changes (I do attempt to look at and through about every change!) and he does like the change from Iterator to Iterable.

PR #448, opened 2026-04-29
Gene's LSP report: PR A (grammar) + PR B (scope-aware definition + completion)

lagergren:
## Summary

Bundle of fixes for issues Gene reported in his email (`gene1.md`):

- **Issue 3 (PR A)**: `Int val2.get() = 43;` rejected at module scope.
- **Issue 1 (PR B-1)**: cmd-click on a local variable resolves to a same-named workspace symbol.
- **Issue 2 (PR B-2)**: completion popup misses module-level `@Inject` properties, function-local variables, and parameters.

Issues 4 (manualTests parse failures) and the doc/jump-to-comment issue are tracked separately. The full plan lives in `lsp-bugfixes-gene-plan.md` (not in this branch).

## Commits (review in order)

| # | Commit | What |
|---|--------|------|
| 1 | `grammar: accept short-form property getter at module/package level` | Issue 3. Adds `property_getter_declaration` to `module_body` and `package_body` in `grammar.js.template`. Two `TreeSitterAdapterTest` cases. |
| 2 | `chore(lang): refresh generated-examples from current DSL/template` | Routine refresh of auto-generated documentation files. No source behavior changes. |
| 3 | `deps(lang): bump lsp4ij 0.19.2 -> 0.19.3` | Patch bump on the Red Hat IntelliJ LSP4IJ plugin. Notable fixes: inverted semantic-token offsets, codelens display, IntelliJ Platform 2026.1 support. |
| 4 | `completion: include file declarations and imports in BODY context` | Issue 2 part 1. The classifier returns `BODY` for cursors in function bodies; that branch was skipping `findAllDeclarations`/`findImports`, so module-level `@Inject Console console;` never reached the popup. Mirrors what `DEFAULT` already does. One regression test reproducing Gene's exact `co` → `console` scenario. |
| 5 | `lsp: scope-aware definition lookup, prefer locals over workspace symbols` | Issue 1. Adds `XtcQueryEngine.resolveByNameInScope` which walks the AST upward from the cursor and matches against locals, parameters, and class/module members in the enclosing scope chain. Used by the definition handler before the workspace fallback. Two `NavigationTest` cases. |
| 6 | `completion: surface in-scope locals/params and order by scope distance` | Issue 2 part 2 + ordering. Adds `enumerateInScope` to share scope-walk infrastructure with definition. BODY context now lists locals/params/enclosing-scope members, **ranked ahead of built-in types and keywords**. Five new tests including a complex mixed-kind ordering test. |

## Test plan

- [x] `:lang:tree-sitter:validateTreeSitterGrammar` — grammar regenerates and validates.
- [x] `:lang:tree-sitter:testTreeSitterParse` — XDK corpus still parses.
- [x] `:lang:lsp-server:test` — full suite green; 9 new tests added across `TreeSitterAdapterTest`, `NavigationTest`, `CompletionTest`.
- [x] Direct `tree-sitter parse` on a hand-written `.x` file with `Int val2.get() = 43;` at module scope — clean parse tree, no `ERROR` nodes.
- [x] `:lang:intellij-plugin:compileKotlin` — IntelliJ plugin still compiles against lsp4ij 0.19.3.
- [ ] Manual smoke test in IntelliJ: cmd-click on a local variable, completion in a function body listing locals/params, no underline on imports (verified earlier on master).

## Scope notes

- **Workspace fallback for cross-file navigation is preserved.** The new scope walk only short-circuits when it finds a match; if the local scope has no matching name, the existing workspace index is consulted exactly as before. So `cmd-click` on a free function defined in another file still works.
- **`property_getter_declaration` is still not captured by the `allDeclarations` query**, so module-level getters don't appear as workspace symbols. Pre-existing behaviour at every scope level. Worth a follow-up but not in this PR.
- **`sortText` is not yet emitted** on completion items. The BODY-block ordering controls the response array order; the IDE may re-rank by its own fuzzy/relevance score on top. If we want hard control over IDE ordering we'd need to add `sortText` to `CompletionItem` and the LSP4J protocol mapping — separate piece of work.


lagergren:
## Update — additional tests and manual test plan refresh

Pushed two more changes addressing earlier review questions.

### New regression tests for scope-walk edge cases (commit `b49b243`)

Three behaviors were correct in the implementation but lacked explicit coverage. Now in `NavigationTest.DefinitionTests`:

- **`should not resolve to local declared after the cursor`** — forward references aren't legal Ecstasy. A reference to `globalThing` above its local declaration must skip the forward-declared local and walk up to the module-level property. Guards the `endsBefore` filter on `variable_declaration`.
- **`should prefer inner-block local over outer-block local`** — a name redeclared in a nested `if`/`while` body must shadow the outer-scope name.
- **`should fall through to same-file declarations when name is not in scope`** — when scope walk finds no match, resolution continues to `findAllDeclarations` (and ultimately the workspace index). Verifies my changes don't accidentally break cross-scope navigation.

### `lang/doc/MANUAL_TEST_PLAN.md` updates

- **Section 3 (Code Completion)**: 5 new entries for in-scope completion (module-level `@Inject` in body, function-local variable, method parameter, class member from method body, scope ordering check).
- **Section 4 (Go to Definition)**: 5 new entries for scope-aware resolution (parameter, function-local, local shadows class member, inner-block shadows outer, forward-reference rejection). Updated the status to "scope-aware same-file + cross-file via workspace index" and documented the resolution order. Also documented the known doc-comment-landing limitation as a follow-up.
- **Section 7 (Diagnostics)**: 2 new entries for module-level and package-level property getter parsing cleanly (Issue 3 coverage).

### Remaining issues from Gene's email after PR A + B (deferred)

Both items below are **independent of this PR** and should land in separate PRs. They don't block #448.

1. **Doc-comment hijacks the jump target.** From Gene's preamble: *"jumping to the comment rather than the actual declaration"*. `findAllDeclarations` returns the *outer* declaration node's start position, which includes the optional `/** doc */` prefix. Fix: return the `name` field's location instead (same approach `resolveByNameInScope` already uses). Small change to `XtcQueryEngine.findAllDeclarations` + `SymbolInfo` location, plus tests.
2. **Issue 4 (14 manualTests files fail to parse).** Heterogeneous grammar gaps — `@Inject` annotations on local declarations, package-import `inject(args) using ...` clauses, postfix async `^()`, modifier/annotation ordering on property getters, etc. Tracked as PR C in `lsp-bugfixes-gene-plan.md`. File-by-file campaign with corpus tests.

ggleyzer (approved):

PR #451, opened 2026-04-30
Allow custom names for properties in JSON serialization

thegridman:
Sometimes we need to use a name for a property in JSON that is not a valid identifier in Ecstasy. We could use a custom mapping for this, but that means hand coding serialisation just for a property name change. This PR adds the `JsonProperty` annotation that can be applied to a property to specify a different name to use when the type is being serialised using the `ReflectionMapping`.

### DIFF
--- a/lib_json/src/main/x/json/annotations/JsonProperty.x
+++ b/lib_json/src/main/x/json/annotations/JsonProperty.x
@@ -1,0 +1,8 @@
+/**
+ * An annotation to allow customization of the JSON name of a value.
+ *
+ * @param jsonName - the customized name of the field
+ */
+annotation JsonProperty(String jsonName)
+        into Class | Property {
+}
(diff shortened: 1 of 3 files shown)


lib_json/src/main/x/json/annotations/JsonProperty.x
ggleyzer, line 1: I'd suggest adding a simple usage example to the doc here

ggleyzer (approved):

ggleyzer:
The only thing that gives a bit of a pause - we usually prefer the data classes **not** to be aware of the way they are being serialized (be it with JSON serialization or any other methods). In the case of the JSON, the specialization/customization is handled by the Schema, Mappings etc.

Your approach goes against that design... but I don't have a better proposition

thegridman:
@ggleyzer I can see how separating a data class from how it is serialized makes sense. I'm happy to hold off merging this.

lagergren (approved):

PR #452, opened 2026-05-01
LSP follow-up: tree-sitter grammar gaps closed (callTests + literals + multiline-template + C5 + skip-list cleanup)

lagergren:
## Summary

Multi-session sweep over remaining tree-sitter grammar gaps. Active sweep moves from 853/876 (with 7 parked entries) to **880/880 with 13 parked entries** — the parked count *grew* only because re-categorizing \`intentional.txt\` uncovered 13 real grammar gaps that were previously hidden under the \"intentionally not parseable\" label.

## Highlights

**callTests.x** — \`(i1, Int i2) = testMultiReturns(0);\` (mixed bare-ident + typed-decl tuple_assignment). Closed by combining:
- An external-scanner \`TUPLE_ASSIGN_LPAREN\` token that peeks past the matching \`)\` for \`=\` (and not \`==\` / \`=>\`) before any \`;\` / \`{\` / \`}\` / EOF, requires at least one \`,\` at depth 1, and only fires at statement start. Forces LR commit to \`tuple_assignment\` at the \`(\` shift.
- Narrowing \`tuple_assignment_element\`'s bare alt from \`prec.dynamic(1, \$._expression)\` to \`choice(\$.identifier, \$.member_expression, \$.index_expression)\` — the three node shapes that actually appear as bare LHS slots in real code. With the scanner suppressing the container-level fork, a full \`_expression\` alt was a trap (typed elements like \`UInt8 left\` could lose to a failing _expression parse with no backtrack).
- Scanner skips leading whitespace via \`lexer->advance(skip=true)\` (tree-sitter doesn't auto-skip extras before invoking external scanners) but **defers to the TODO scanner** when both are valid, otherwise it eats the hspace TODO needs.

**literals.x** — bare absolute-path primary expression (\`Path p = /allow.txt;\`). Closed by adding the existing \`path_literal\` rule to \`_literal\`'s choice and tightening the regex first-char class to disambiguate against \`//\` (line comment) and \`/*\` (block comment) starts.

**reflect.x** — closed by the multiline-template interpolation work (template interpolation expressions can now span physical lines via the parent \`|\` continuation; scanner emits \`MULTILINE_CONTINUATION\` when persistent state shows we're inside a multiline-template interpolation expression).

**generics.x** — closed by the C5 LHS work (split-shift operators \`<<\` / \`>>\` / \`>>>\` so the lexer never produces a multi-char shift token, unblocking nested type-argument lists like \`Function<<Int, String>, <Int>> f = ...\`).

**Skip-list re-categorization** — the previous \`intentional.txt\` claimed 19 files were \"intentionally not parseable by tree-sitter\". Re-checking against both parsers shows:
- 4 files parse cleanly with both — moved into the active sweep.
- 13 files have no Java \`PARSER-NN\` errors but produce ERROR nodes in tree-sitter — these are real grammar debt; moved to \`parked.txt\` for follow-up. Tracked by issue (filed alongside this PR).
- 2 files (\`socks-catalog.x\`, \`socks-user.x\`) have genuine \`PARSER-03\` errors from the Java compiler too; remain in \`intentional.txt\` with a corrected header documenting the real criterion (\"xcc emits PARSER-NN\").

## Verification

\`\`\`
./gradlew :lang:dsl:generateScannerC :lang:dsl:generateTreeSitter \\
          :lang:tree-sitter:validateTreeSitterGrammar \\
          :lang:updateGeneratedExamples \\
          :lang:tree-sitter:testTreeSitterParse \\
          :lang:lsp-server:test \\
          :lang:intellij-plugin:verifyPlugin \\
          -PincludeBuildLang=true -PincludeBuildAttachLang=true
\`\`\`

- Sweep: **880/880** passing
- \`validateTreeSitterGrammar\`: no \"unnecessary conflict\" warnings (also dropped one stale conflict that became redundant after the bare-element narrowing)
- \`lsp-server:test\`: green
- \`intellij-plugin:verifyPlugin\`: Compatible
- Configuration cache: stored / reused throughout

## Test plan

- [ ] Reviewers spot-check the four formerly-\`intentional\` files now in the sweep parse to non-trivial trees: \`archive/web-static.x\`, \`dbTests/PeopleDB.x\`, \`jsondb/.../JsonValueStoreTest.x\`, \`webTests/Hello.x\`.
- [ ] Reviewer eyeballs the new TUPLE_ASSIGN_LPAREN scanner block in \`ScannerCGeneratorDsl.kt\` (TODO-deferral guard is the load-bearing detail).
- [ ] Reviewer eyeballs the narrowed \`tuple_assignment_element\` and confirms \`choice(\$.identifier, \$.member_expression, \$.index_expression)\` covers all bare LHS shapes used in this codebase (the parked.txt comment lists the working cases).


ggleyzer (approved):
Looks so much cleaner now!

lagergren:
> Looks so much cleaner now!

Thx. This was HARD.

PR #457, opened 2026-05-05
[WIP / POC] injectable logging APIs - Not intended as a PR to merge.

lagergren:
## 📖 Start here → [`doc/logging/README.md`](https://github.com/xtclang/xvm/blob/lagergren/logging/doc/logging/README.md)

That README is the single entry point. It carries the goals, the requirements (with a status column linking to evidence), the non-goals, the reading paths for different reviewer roles, and the full doc index. Every other doc in this branch links back to it via prev/next/up footers.

Other top-level entry points reviewers most often want:

- [`roadmap.md`](https://github.com/xtclang/xvm/blob/lagergren/logging/doc/logging/roadmap.md) — what blocks merge, what lands in v1, what is explicit future
- [`lib-logging-vs-lib-slogging.md`](https://github.com/xtclang/xvm/blob/lagergren/logging/doc/logging/lib-logging-vs-lib-slogging.md) — the API-choice comparison
- [`open-questions.md`](https://github.com/xtclang/xvm/blob/lagergren/logging/doc/logging/open-questions.md) — open questions for the XTC language/runtime designers (Q-D1…Q-D7)
- [`usage/injected-logger-example.md`](https://github.com/xtclang/xvm/blob/lagergren/logging/doc/logging/usage/injected-logger-example.md) — recommended first stop if you've never touched the library

---

**[WIP / POC]** This PR explores two parallel injectable XDK logging POCs so the project can compare a familiar SLF4J/Logback-shaped API against a Go slog-shaped API before choosing one canonical XDK logging facade. It is not intended to merge as-is; the merge-blocking decisions are listed in [`roadmap.md`](https://github.com/xtclang/xvm/blob/lagergren/logging/doc/logging/roadmap.md).

## Summary

- Adds `lib_logging` as the SLF4J-shaped POC with named loggers, levels, MDC, markers, message formatting, structured key/value events, JSON output, and sink extension points.
- Adds `lib_slogging` as the slog-shaped comparison POC with records, attrs, groups, handlers, JSON/text rendering, context helpers, and handler extension points.
- Wires temporary runtime injection for both logger shapes while tolerating runners that only carry one POC module.
- Ships `doc/logging/` as a hierarchical, depth-first doc tree starting at `README.md`, with goals/requirements/non-goals split, status table linking to evidence, prev/next navigation between docs, and per-doc audience tags.

## Verification

- `./gradlew build` — full root build, 250 tasks
- `./gradlew :xdk:lib-logging:testXtc` — 70/70 XTC tests pass
- `./gradlew :xdk:lib-slogging:testXtc` — 41/41 XTC tests pass
- `./gradlew :manualTests:runOne -PtestName=TestLogging` — end-to-end demo
- `./gradlew :manualTests:runOne -PtestName=TestSLogging` — end-to-end demo


lib_slogging/src/main/x/slogging/Logger.x
```diff
@@ -0,0 +1,264 @@
...
+    construct() {
+        construct Logger(new TextHandler());
+    }
+
+    /**
+     * Compatibility convenience: handler plus pre-bound attrs. The attrs are handed to
+     * [Handler.withAttrs] immediately, which is the slog contract and what allows a
+     * handler to pre-render or cache its own derived state.
+     */
+    construct(Handler handler, Attr[] attrs) {
```
thegridman: Do we need `Attr` as it is basically a key/value pair where they key is a String so `attrs` could just be a Map.
```
construct(Handler handler, Map<String, Attribute> attrs = [])
```
Where `Attribute` above would probably be some sort of union typedef as I'm pretty sure we'd probably restrict the types. They at least need to be `Sharable` as they are passed to a service.

Then instead of having to do things like `[Attr.of("foo", "bar")]` in the constructor you can do
```
new Logger(handler, ["foo"="bar"])`
```

lib_slogging/src/main/x/slogging/Logger.x
```diff
@@ -0,0 +1,264 @@
...
+     * resource supplier is registered (parallel to the SLF4J library's
+     * `NativeContainer.ensureLogger`).
+     *
+     * Explicit delegating constructors (rather than `attrs = []` default on the
+     * primary form) for the same reason as the SLF4J side: cross-module
+     * default-argument resolution on `const` constructors does not always synthesise
+     * the shorter form. See `lib_logging.BasicLogger`.
+     */
+    construct() {
+        construct Logger(new TextHandler());
```
thegridman: You do not need three constructors, just use default values.
```
construct(Handler? handler = Null, Attr[] attrs = []) {
    this.handler = handler ?:  new TextHandler();
    if (!attrs.empty) {
        this.handler =this.handler.withAttrs(attrs);
    } 
}

Handler handler;
```

thegridman:
As I said on our weekly call, I prefer the Slog implementation to SLF4J as I feel it is simpler.

I have done some refactoring of it.
- Refactored various types that had multiple constructors and used a single constructor with default parameter values instead.
- Removed the `Attr` type. As I said in a comment, this is just a `String` to `Object` key value pair and we can use a `Map` to represent attributes instead. In `slogging.x` I added a few `typedef` statements to define attributes. This also restricts the types that can be used for attribute values. I have also defined a typedef called `Attributes` which is a `Map<String, AnyValue>` where `AnyValue` is also a typedef of allowed attribute values.

I changed the allowed attribute types to match those allowed by Open Telemetry as this is a standard with a spec and I think if there is a standard we should use it. I may make other changes to align better with Open Telemetry logging. After all, I would think in production we will end up shipping logs to a log collector as writing to and handling log files is so "last decade" Dev Ops.

thegridman:
I refactored `lib_slogging` to align the log `Record` closer to the Open Telemetry logging data model. This allows `AnyValue` as the log message type instead of just a `String`. In fact, when the Open Telemetry docs discuss structured logging, the recommendation is to not use a "message" field and put everything into the `attributes`. This is because Open Telemetry log collectors will typically index the attributes to allow faster and easier searching of logs, but they tend to not do full text indexing of the message field.

thegridman:
Rebased from master

PR #463, opened 2026-05-17
Fix a compiler exception

ggleyzer:
This issues was [reported by Dima on Discord](https://discord.com/channels/837372142576205885/1406009322307190834/1505484174197260468).

The gist of the problem lies in the flow of processing by the StageManager:

1) during the "resolveName" stage, we resolve all the names with a possible exception of the annotation parameters, since they could be represented by expressions and need to be "validated"

2) during the "validateContent" stage, we validate those "left behind" annotation parameters; if detect that some of them are on not-yet-validated, we repeat the cycle via `mgr.requestRevisit()` API

The problem arises when the validation requires using the TypeInfo for classes that depend on those "not-yet-validated" ones. In that case, the `UnresolvedNameConstant` or `UnresolvedTypeConstant` would throw an `UnsupportedOperationException`, causing the compiler termination.

The proposed fix replaces the `UnsupportedOperationException` with the `CompilerException` and handles it in the `StageManager` in the same way a standard "Name not resolved" error

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/constants/UnresolvedNameConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/UnresolvedNameConstant.java
@@ -8,6 +8,9 @@ import java.util.function.Consumer;
 
 import org.xvm.asm.Constant;
 import org.xvm.asm.ConstantPool;
+
+import org.xvm.compiler.CompilerException;
+
 import org.xvm.util.Hash;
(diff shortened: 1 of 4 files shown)


cpurdy (approved):
Looks like a reasonable solution to avoid the exception. Are there still any remaining cases where we should be able to resolve something but don't?

PR #465, opened 2026-05-18
Fix HMAC key sizes for HmacSHA256 and HmacSHA512

mrcompator:
Added 44 bytes HmacSHA256 secret key length to alllow AWS SigC4 keys
Enhance HMAC key handling to support PublicOrSecret and PrivateOrSecret forms
Update key extraction to support PrivateOrSecret form in xRTSigner

### DIFF
--- a/javatools_bridge/src/main/x/_native/crypto/RTAlgorithms.x
+++ b/javatools_bridge/src/main/x/_native/crypto/RTAlgorithms.x
@@ -124,8 +124,8 @@ service RTAlgorithms {
      */
     static Tuple<String, KeySize>[] macs = [
         ("HmacSHA1",    16),
-        ("HmacSHA256", [16, 128]),
-        ("HmacSHA512", [32, 128]),
+        ("HmacSHA256", [16, 32, 44]),
+        ("HmacSHA512", [32, 64]),
     ];
 
     /**
--- a/javatools_bridge/src/main/x/_native/crypto/RTAlgorithms.x
+++ b/javatools_bridge/src/main/x/_native/crypto/RTAlgorithms.x
@@ -124,8 +124,8 @@ service RTAlgorithms {
      */
     static Tuple<String, KeySize>[] macs = [
         ("HmacSHA1",    16),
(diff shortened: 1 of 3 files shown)


javatools_bridge/src/main/x/_native/crypto/RTAlgorithms.x
```diff
@@ -124,8 +124,8 @@ service RTAlgorithms {
      */
     static Tuple<String, KeySize>[] macs = [
         ("HmacSHA1",    16),
-        ("HmacSHA256", [16, 32]),
```
ggleyzer, line 127: I think it's better to simply add the desired sizes, don't you?
```
        ("HmacSHA256", [16, 32, 128]),
```

ggleyzer (changes requested):

javatools_bridge/src/main/x/_native/crypto/RTAlgorithms.x
ggleyzer: And BTW, while there maybe add "128" to the SHA256?

mrcompator: oh! My bad, I thought those were min/max sizes. Shame on me for not testing it :man_shrugging:  On the other hand I have checked and there are multiple different sites requiring different lengths:

1. Shopify Webhooks 38 bytes
2. AWS SigV4 44 bytes
3. Enterprise JWT min 32 bytes up 128 bytes
4. Github Webhook secrets: user defined length 

Perhaps let's leave this PR open for now and have a think what the best approach is?

ggleyzer: I say: add what you need right now and I'll merge it )

mrcompator: Added 44 to HmacSHA256

ggleyzer (approved):
Looks good

PR #466, opened 2026-05-22
Improve GenericTypeResolver API

ggleyzer:
An important part of the compiler type validation and runtime is resolution of generic/formal types. The GenericTypeResolver used to have two paths: one for resolving the FormalConstant and the other for resolving the generic type names, while the first one had a default implementation immediately turning the formal constant into its name. The problem with this approach was that some times unrelated formal constants of the same name would resolve the same way.

This change eliminates the name-based resolution in favor of the FormalConstant-based one.

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/ClassStructure.java
+++ b/javatools/src/main/java/org/xvm/asm/ClassStructure.java
@@ -3771,8 +3771,16 @@ public class ClassStructure
         }
 
         @Override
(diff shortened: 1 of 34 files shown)


jnorthrup:
https://github.com/xtclang/xvm/pull/473 works fine needs no special adaptation to work with this PR, there are maybe one or generics checks that were dissimilar that typedef only runs into by bringing generics into the type data and checks.

PR #483, opened 2026-06-17
Bundle SLF4J no-op provider in javatools

ggleyzer:
When we start the platform that uses non-self-certifying CA, such as "Let's Encrypt" the following messages get logged:

```
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details

```
The reason for that is that `javatools.jar` includes acme4j for ACME/certbot certificate support, which uses SLF4J internally, and the fat jar ends up containing `slf4j-api`, but no SLF4J provider.

This change adds `slf4j-nop` to the `javatools` runtime so the existing effective behavior is preserved while avoiding the startup warning. This keeps ACME logging disabled, which matches the previous runtime behavior after SLF4J fell back to NOP.

Disclaimer: the fix was created by Codex

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -27,6 +27,7 @@ jakarta-xml-bind-api = "4.0.5"
 jaxb-runtime = "4.0.9"
 jetbrains-annotations = "26.1.0"
 jline = "4.0.15"
+slf4j = "2.0.18"
 
 # =============================================================================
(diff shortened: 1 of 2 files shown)


lagergren (approved):
Looks great.

PR #491, opened 2026-07-01
Fix for duplicate and incorrect exploded property methods.

cpurdy:
Eliminated duplicates by capping the necessary method.


ggleyzer (approved):
looks good

PR #501, opened 2026-07-08
Address TODOs and fix a regression

cpurdy:


ggleyzer (approved):
I'm not sure what is so special about AccessTypeConstant that it deserves both `getValueString()` and `toString()`

PR #502, opened 2026-07-10
Avoid XDK recompiles when rebuild is disabled

ggleyzer:
When I change some classes in the javatools Java library and I know for sure it doesn't impact the .x file compilation I run this command:

    ./gradlew installDist --no-scan -PxtcDefaultRebuild=false

However I noticed that sometimes it still re-compiles some or all of xdk libraries.

I asked Codex to fix that - this is what it came up with.

Here's the explanation of what it found:

> The issue was that -PxtcDefaultRebuild=false only changed the flag passed to the XTC compiler. Gradle was still treating javatools.jar and launcher runtime jars as inputs to every compileXtc task, so a Java-only javatools change could make Gradle mark the XDK library compile tasks out-of-date before the compiler ever saw rebuild=false.
> This change makes the Gradle task model match the intent of rebuild=false. When rebuild is disabled, XTC compile tasks no longer fingerprint the launcher runtime jars as inputs, but they still depend on the xdkJavaTools configuration so the Java tools are built before compilation if compilation is actually needed. Extracted XDK contents are also narrowed to track .xtc module files instead of all files in that directory.

### DIFF
--- a/plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
+++ b/plugin/src/main/java/org/xtclang/plugin/XtcProjectDelegate.java
@@ -460,6 +460,7 @@ public class XtcProjectDelegate {
         compileTask.configure(task -> {
             task.setDescription("Compile an XTC source set, similar to the JavaCompile task for Java.");
             task.dependsOn(XDK_EXTRACT_TASK_NAME);
+            task.dependsOn(configs.getByName(XDK_CONFIG_NAME_JAVATOOLS_INCOMING));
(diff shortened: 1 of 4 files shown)


lagergren (approved):
This is a path I have trod before and have to spend an hour or so picking at this to verify the configuration dependency is sane. The proper fix would to never use the rebuild flag at all in the plugin, since dependencies are handled by gradle and the plugin. The hard part in this is that you want to recompile javatools and not have to rebuile the xdk, to which it is a full dependency, but you can't really model that.

PR #510, opened 2026-07-31
Jit

cpurdy:
Merge stable JIT work into mainline


cpurdy:
Ran into a second issue from an earlier commit, so I'll resubmit this shortly with everything re-tested.

PR #511, opened 2026-07-31
Add support for @Lazy static properties

ggleyzer:
This issue arose from [the question from Michael on Discord](https://discord.com/channels/837372142576205885/1406009322307190834/1532566968752931007)

As it is, we don't allow `@Lazy` static properties. 

This change adds the necessary support in the compiler and the interpreter.


cpurdy (approved):
Reviewed with Gene, approved pending two last changes.

PR #512, opened 2026-07-31
JIT Milestone: More core classes are compiling now.

cpurdy:
JIT Milestone: More core classes are compiling now.


ggleyzer (approved):
Reviewed multiple times

PR #516, opened 2026-08-06
Avoid mutating pooled property signatures

mrcompator:
## Summary
- set the property-signature flag before registering a resolved signature
- retain canonical constant-pool identity for property signatures and ordinary method signatures
- add focused coverage for canonical reuse and method/property separation

## Why
`SignatureConstant` comparison includes the transient property-signature flag. `resolveGenericTypes()` previously obtained a pooled method signature and then set `m_fProperty`, changing the semantic identity of a key after insertion into the constant pool's `ConcurrentHashMap`.

A subsequent lookup for the equivalent method signature could no longer match that mutated key. We observed the resulting null lookup as an NPE while preparing a JsonDB multi-map transaction.

The fix creates the resolved signature, sets `m_fProperty`, and only then registers it. This avoids mutating a registered key while preserving the constant pool's one-canonical-instance behavior and state such as `m_sJitName`.

## Reproducer
`SignatureConstantTest.resolvingGenericPropertySignaturePreservesCanonicalInstances()` creates and resolves a generic property signature twice. It verifies that both resolutions return the same pooled property signature and that the equivalent method signature remains a distinct non-property instance.

## Test plan
- [x] `./gradlew :javatools:test --tests org.xvm.asm.constants.SignatureConstantTest`
- [x] `./gradlew :javatools:test`

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/constants/SignatureConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/SignatureConstant.java
@@ -285,10 +285,18 @@ public class SignatureConstant
         }
 
         if (fDiff) {
-            SignatureConstant that = pool.ensureSignatureConstant(
+            // Property signatures are transient (never serialized). Do not register one into the
+            // ConstantPool and then flip m_fProperty on a pooled method signature — SignatureConstant
+            // is used as a ConcurrentHashMap key, so mutating m_fProperty after registration corrupts
+            // the pool.
+            if (m_fProperty) {
+                SignatureConstant that = new SignatureConstant(
+                        pool, getName(), aconstParamResolved, aconstReturnResolved);
+                that.m_fProperty = true;
+                return that;
+            }
+            return pool.ensureSignatureConstant(
                     getName(), aconstParamResolved, aconstReturnResolved);
-            that.m_fProperty = this.m_fProperty;
-            return that;
         }
 
         return this;
--- a/javatools/src/main/java/org/xvm/asm/constants/SignatureConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/SignatureConstant.java
@@ -285,15 +285,12 @@ public class SignatureConstant
         }
 
         if (fDiff) {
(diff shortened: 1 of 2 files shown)


ggleyzer:
I agree, it's a bug, but the fix is too specific. While it's true that the property signature is transient and never serializes into the file structure, it doesn't mean we should not put it into the pool. We commonly use the ConstantPool to ensure that we have one only one instance for a given constant. That allows us to keep some state (such as m_sJitName) on that one-and-only-one instance.

Therefore, the correct fix is to create a new SC, change its m_fProperty field and only then register it.

javatools/src/main/java/org/xvm/asm/constants/SignatureConstant.java
```diff
@@ -285,10 +285,18 @@ public SignatureConstant resolveGenericTypes(ConstantPool pool, GenericTypeResol
         }
 
         if (fDiff) {
-            SignatureConstant that = pool.ensureSignatureConstant(
+            // Property signatures are transient (never serialized). Do not register one into the
```
ggleyzer: Trivial: we usually don't capitalize in-method comments and separate them by semicolon. Also, I would keep it short and specific - avoiding reference to a bug that this change is fixing. For  example:

    // don't register the constant until its state is changed 
    if (m_fProperty) {
                SignatureConstant that = new SignatureConstant(
                        pool, getName(), aconstParamResolved, aconstReturnResolved);
                that.m_fProperty = true;
                return getConstantPool().register(that);
            }

ggleyzer (changes requested):

javatools/src/main/java/org/xvm/asm/constants/SignatureConstant.java
mrcompator: Updated in 960287eed: the property flag is now set before pool.register(), preserving canonical pooling without mutating a registered key. I also shortened the comment and changed the regression test to verify canonical instance reuse.

ggleyzer (approved):

PR #517, opened 2026-08-06
Support raw AES secret keys

mrcompator:
## Summary
- construct raw AES keys with `SecretKeySpec` instead of requesting an AES `SecretKeyFactory`
- use the base `AES` key algorithm for all currently registered AES cipher transformations
- add a standalone Ecstasy reproducer that exercises the public crypto API end to end

## Why
The crypto API accepts a raw `PrivateKey` for AES and successfully creates a decryptor, but the first `encrypt()` call fails with:

```
RTError: AES SecretKeyFactory not available
```

`SecretKeyFactory` is intended for algorithms with a provider-specific key specification; standard Java providers do not normally expose one for AES. Raw AES bytes are already the encoded secret key and should be represented by `new SecretKeySpec(bytes, "AES")`.

For transformations such as `AES/CBC/PKCS5Padding`, the full transformation belongs in `Cipher.getInstance(...)`, while the key algorithm must remain `AES`.

This change only fixes raw AES key import. It does not add IV persistence, authentication, or AES-GCM support.

## Reproducer
`manualTests/src/main/x/AesRawKeyRepro.x` constructs a 256-bit AES `PrivateKey`, obtains a decryptor through the public Ecstasy API, encrypts `hello`, and decrypts it again. This exercises the Ecstasy-to-native boundary where the failure occurs.

The original runtime fails at `encrypt()` with the error above; the patched runtime completes successfully.

## Test plan
- [x] `./gradlew :javatools:test`
- [x] `./gradlew :manualTests:compileXtc`
- [x] Run `AesRawKeyRepro` using the locally installed patched XDK


javatools/src/test/java/org/xvm/runtime/template/_native/crypto/xRTAlgorithmsTest.java
ggleyzer, line 1: I don't see any reason to have a java test if we have a natural (Ecstasy) test

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTAlgorithms.java
```diff
@@ -208,63 +208,81 @@ public static Key extractKey(Frame frame, ObjectHandle hKey, String sAlgorithm,
...
-            default:
-                // unlike specific cases above, default implementation uses generic "SecretKeySpec"
-                return switch (keyForm) {
-                    case PublicOrSecret, PrivateOrSecret ->
-                        SecretKeyFactory.getInstance(sAlgorithm).
-                            generateSecret(new SecretKeySpec(abRaw, sAlgorithm));
+    /**
+     * Extract a key from its raw encoding.
+     */
+    static Key extractKey(byte[] abRaw, String sAlgorithm, KeyForm keyForm)
```
ggleyzer: this method was created with a sole purpose to be called from a unit test. If we decide that the natural (.x) test sufficient, which I strongly prefer, there is no need for that in could be inlined back

ggleyzer (changes requested):

ggleyzer:
Good catch!

javatools/src/test/java/org/xvm/runtime/template/_native/crypto/xRTAlgorithmsTest.java
mrcompator: Updated in 35e165260: removed the Java test and retained the natural Ecstasy reproducer, which exercises raw AES encryption through the public API and native boundary.

javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTAlgorithms.java
mrcompator: Updated in 35e165260: removed the test-only overload and inlined the AES handling back into the original extractKey method.

ggleyzer (approved):

PR #518, opened 2026-08-06
Add authenticated AES encryption and key wrapping

mrcompator:
## Summary
- add an authenticated-encryption API backed by AES-GCM with generated nonces, AAD, and fail-closed authentication
- add AES-KW and AES-KWP key-wrapping APIs for visible and opaque secret keys
- document safe AES usage and cover the implementations with NIST/RFC vectors, tamper tests, and smoke modules

## Test plan
- [x] `./gradlew :xdk:lib-crypto:testXtc`
- [x] installed-XDK `AesGcmTest` smoke module
- [x] installed-XDK `AesKeyWrappingTest` smoke module
- [x] `./gradlew check`
- [x] `git diff --check`

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTAlgorithms.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/crypto/xRTAlgorithms.java
@@ -148,7 +148,9 @@ public class xRTAlgorithms
             }
 
             case 1:   // SymmetricalCipher
-            case 2: { // AsymmetricalCipher
+            case 2:   // AsymmetricalCipher
+            case 6:   // AuthenticatedCipher
+            case 7: { // KeyWrappingCipher
                 Cipher cipher = Cipher.getInstance(sName);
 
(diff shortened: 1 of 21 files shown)


mrcompator:
CTR, CFB, OFB, and CTS are intentionally excluded from this change. They are unauthenticated AES modes: ciphertext can be modified without detection, and their IV/nonce requirements are easy to misuse. They remain useful for compatibility with specific legacy protocols or storage formats, but AES-GCM should be the application-facing default. We can add an individual low-level mode later when an interoperability requirement justifies its API and tests.

lagergren (approved):

ggleyzer:
I'm not convinced we need the seal/open operations, KeyWrapper etc. Most of the time the CryptoKey you get from a store is an impenetrable shell that can be used for encryption/decryption purposes, but cannot be used to retrieve the raw cryptographic material anyway.

Granted, I may not fully understand the intent, so a face-to-face review/discussion seems necessary

lagergren (approved):

PR #519, opened 2026-08-07
Cache directory repository module metadata

ggleyzer:
## Intent

Avoid deserializing every XTC module merely to discover its qualified name when a DirRepository is reconstructed in a new JVM. This preserves repository ordering and support for modules stored under nonstandard filenames.

DirRepository now stores a small transient index under the system temporary directory. Each entry records the filename, timestamp, size, module name, and prior load error state. On startup, unchanged entries are reused and only new or modified modules are deserialized. Cache files are path-bound, versioned, validated against their physical size, written atomically, and treated strictly as an optional performance aid; malformed or unavailable caches fall back to the existing scan.

## Performance

The first run still pays the existing scan cost and creates the index. Subsequent runs benefit while unchanged modules remain valid.

- full manual JIT suite run improved from 6.5s to 4.7s
- manualTests:runParallel improved from 17s to 13s, approximately 24%

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/DirRepository.java
+++ b/javatools/src/main/java/org/xvm/asm/DirRepository.java
@@ -9,8 +9,6 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 
-import java.nio.charset.StandardCharsets;
-
 import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -21,7 +19,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.UUID;
 
 
 /**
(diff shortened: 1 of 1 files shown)


javatools/src/main/java/org/xvm/asm/DirRepository.java
```diff
@@ -149,6 +175,140 @@ protected void ensureCache() {
...
+     */
+    private File getCacheFile() {
+        String tempDir = System.getProperty("java.io.tmpdir");
+        if (tempDir == null) {
+            return null;
+        }
+
+        try {
+            String path = m_dir.getCanonicalPath();
+            String key  = UUID.nameUUIDFromBytes(path.getBytes(StandardCharsets.UTF_8)).toString();
```
cpurdy: wouldn't we always want to use the same name?

cpurdy (approved):
Looks fine in general.

javatools/src/main/java/org/xvm/asm/DirRepository.java
ggleyzer: The `UUID.nameUUIDFromBytes()` just produces an MD5 digest of the path name. The idea was suggested by Codex and I didn't think much about it. After the reflection, however, I agree that it's unnecessarily heavy and obscure. I'll change it something simpler, e.g.
    
    `String key = Integer.toUnsignedString(path.hashCode(), 16);`

PR #525, opened 2026-08-13
Add 'xtc bundle': merge compiled modules into a single multi-module .xtc

lagergren:
## Summary

Adds a `bundle` verb to the unified launcher that merges compiled per-module `.xtc` files into a **single multi-module `.xtc` container** ("bundle") that every tool — runner, compiler, linker, reflection — can resolve modules from by name. One file replaces a `lib/` directory.

The `.xtc` format always defined a `FileStructure` as a container of one or more modules, and the runtime already fuses `ecstasy`+`mack`+`_native` into one in-memory structure on every startup; a bundle simply persists that shape to disk.

```bash
xtc bundle -L build/install/myapp/lib          # bundle an install dir, main module inferred
xec -L myapp.bundle.xtc myapp.bundle.xtc       # run straight from the bundle
```

## What's in the change

**The `bundle` verb** (`Bundler`, `BundlerOptions`, `Launcher` wiring): selection by *declared* module name from the module path (or explicit names/files), main-module inference (`--main` to override), `-o` output, `--include-system` to lift the default exclusion of `xtclang.org` modules (app bundles stay lean; system modules remain run-time-resolved fingerprints), and a warning when a statically-imported module on the path is left out of the bundle.

**Format/assembly layer**:
- `FileStructure.merge` now supersedes a same-id fingerprint with the real module (previously a silent no-op — the "merge order is critical" trap) and synthesizes `Required` fingerprints for dependencies of a merged module that are real siblings in a multi-module source, making merged/detached modules self-describing.
- `FileStructure.linkModules` (runtime): module ids absorbed from a multi-module file that the app never imports are skipped instead of failing the link.
- `ModuleStructure.markEmbedded()` — bundled non-main modules persist as `ModuleType.Embedded` (previously unreachable; `Primary` non-main members trip an assert under `-ea`).
- `FileStructure.hasMultipleChildren()`, `validateModuleConstants()` (pool validation usable offline, without the runtime-only NakedRef precondition).

**Repository layer**: `FileRepository` generalized to expose every real module of a container (`getModuleNames`/`loadModule`, plus a mandatory version-aware `loadModule` override), serving non-main members as **memoized detached copies**; `xCoreRepository.getModule` does the same at the Ecstasy reflection boundary. This preserves the pervasive assumption that a module's file structure is its own single-module file (`getResolvedModule(...).mainModule`, runtime-compiler fingerprint hoisting).

**Bundle metadata and reflection support added after review**:
- Persistent `.xtc` headers now carry constant-pool-free `FileTemplate.Kind` metadata: `Single`, `Library`, and `Executable`. Multi-module bundle output is written as `Library`; `Executable` is represented for the future fully-linked workflow.
- The metadata records real module names, excluding fingerprints, and optional version strings before the constant pool. `FileStructure.readMetadata(...)` can read that block without deserializing the full file.
- `FileRepository` uses the new metadata path to list bundled modules and available versions cheaply, while still loading/detaching module bodies only when a caller asks for a specific module.
- Ecstasy reflection can now tell whether an uploaded file is a bundle via `FileTemplate.kind`, `moduleNames`, and `bundle`; `ModuleTemplate.fingerprint` is native-backed so embedded real modules are distinguishable from dependency fingerprints.
- Round-trip tests cover `Single` and `Library` metadata, unknown metadata block skipping, repository version discovery, and the bundle file kind written by `xtc bundle`.

**Single-file XDK**: opt-in `./gradlew xdk:bundleXdk` (JavaExec, incremental, configuration-cache compatible) produces `xdk/build/bundle/xdk.xtc` — all 22 XDK modules (ecstasy as main, mack, `_native`, every lib) in one fully self-contained 4.36 MB file with zero remaining fingerprints.

**Docs**: `doc/xtc-bundle.md` (concepts, pipeline, design notes, implementation map, validation record, limitations, future work) and the `xtc bundle` section in `doc/xtc-cli.md`.

## Validation

- **xqiz.it platform** (11 modules): bundles to one 2.48 MB `platform.xtc` (separate files ~4.9 MB — shared constant pool) and boots with full parity: identical log sequence and HTTPS behavior, dynamic by-name resolution (`host`/`platformUI`/`proxy_manager`), and the runtime jsondb module generation-and-compile path all work from the single file.
- **Single-file XDK**: applications compile and run with `-L xdk.xtc` as the only system library reference.
- **Two-bundle deployment**: the platform boots and serves from `java -jar javatools.jar run -L xdk.xtc -L platform.xtc platform.xtc <pw>` — two bundle files and a jar, zero lib directories.
- Footprint of the bundled XDK on a trivial app: +25% max RSS, +11% startup vs the lib dir (why the single-file XDK stays opt-in for now).
- 14 new tests (`BundlerTest`, JSON round-trip, dispatch coverage); `javatools:test` and `xdk:test` green.
- Review-response focused verification: `./gradlew javatools:test --rerun-tasks --tests org.xvm.asm.FileStructureTest --tests org.xvm.tool.BundlerTest`.

## Follow-ups (not in this PR)

Switching the generated launcher scripts / docker image / distribution artifacts to `xdk.xtc` and making the single file the default XDK form; a Gradle plugin `xtcBundle` task; fully self-contained application blobs; platform-side upload handling that uses the new `FileTemplate.bundle`/metadata surface.


javatools/src/main/java/org/xvm/asm/FileStructure.java
```diff
@@ -828,6 +871,14 @@ public boolean hasChildren() {
         return getChildrenCount() > 0;
     }
 
+    /**
+     * @return true iff this FileStructure contains more than one module, i.e. it is a
+     *         multi-module container rather than a plain single-module file
+     */
+    public boolean hasMultipleChildren() {
+        return getChildrenCount() > 1;
+    }
```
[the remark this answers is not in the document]
lagergren: Addressed in ea7569082. Copilot was right: FileStructure children include fingerprint dependency stubs, so hasMultipleChildren() now counts only real non-fingerprint modules. Added FileStructureTest.testHasMultipleChildrenIgnoresFingerprints and reran the focused FileStructure/Bundler tests.

ggleyzer:
I haven't looked deeply into this PR yet, but what seems to completely missing is any natural (ecstasy.mgmt and ecstasy.reflect) support for the bundle. We definitely need to make sure the bundle works with the platform. Do we need some addition to the `ModuleRepository` API? Any changes to the `FileTemplate`?

ggleyzer:
To make it a bit more clear, the platform's "upload" endpoint stores a file and uses `Container.loadFileTemplate()` to create a `FileTemplate`. Maybe all we need is an ability for the `FileTemplate` to tell that it's a bundle...

But then there is definitely work to be done on the platform side to process that scenario

ggleyzer (dismissed):
See my comments above please

cpurdy (changes requested):
All in all, this is aligned with what we had planned, so it's a good start. There are a number of issues we need to address before bringing it into the master codebase, so let me enumerate those:

1. When we do this, it will break the platform project, at least temporarily. We need to be prepared for that, including having some additional API support in the Ecstasy libraries for accessing information about the underlying FileTemplate and ModuleTemplate information.
2. Specifically, on ModuleTemplate.x, I am suggesting a change to the API and documentation as follows:
```
    /**
     * Indicates whether this ModuleTemplate is a module fingerprint.
     *
     * A module fingerprint is a ModuleTemplate containing only the subset of the identities within
     * that module that is depended upon by other modules within a FileTemplate, and thus required
     * for linking to. This structure is called a "module fingerprint", because it represents only
     * the outline of the module, and multiple versions of that module may be able to match
     * (fulfill the requirements of) that same fingerprint. As part of the linking process, a module
     * fingerprint that is depended upon is replaced with an actual, complete module definition that
     * matches the fingerprint.
     *
     * A module with `fingerprint == True` must be `resolved == False`.
     */
    @RO Boolean fingerprint.get() = !resolved && this != parent.mainModule;

    /**
     * Indicates whether this ModuleTemplate has been linked, for example by [FileTemplate.resolve].
     *
     * An unresolved module may be a [fingerprint] module. A `resolved` module is never a
     * [fingerprint] module.
     */
    @RO Boolean resolved;
```
 3. We would be using the bundling feature for two dimensions. The first dimension is what you have implemented in this PR. We have decided to name the points on this dimension as follows: (i) "single module", (ii) "library", and (iii) "xecable". The first (`Single`) represent the compiler output: a single module of a single (default CI) version (and any number of  additional fingerprints). The second (`Library`) represents more than one module and/or more than one version of module(s) (and any number of  additional fingerprints). The third (`Xecable`) represents a fully linked module and all the modules its depends upon (transitively closed), ready to be run by `xec`;  it would not contain any fingerprints. We do not need to fully implement all of these and the supporting tool workflows for them today, but we should agree on the terms and goals.
 4. We should modify the header of the persistent FileStructure (see `assemble()` and `disassemble()`) to add this meta information in a length-encoded format at the front of the file (`assemble()` would omit it under a new major/minor version combo, and `disassemble()` would either verify its correctness or skip over it). It would contain the enum `Single/Library/Xecable`, a list of module names (not including any fingerprints) in the file, and an optional (?) version tree for each. None of the meta information would contain any references into the ConstantPool, to avoid the requirement for deserialization.
 5. We would introduce helpers to get this meta data without disassembling the FileTemplate, and take advantage of those new helpers, which should dramatically improve performance of some of our tools.
6. Before merging the branch into master, we'd modify platform to use it, and use that as a test case.

I've gone over the gist of this approach with Gene to validate my assumptions and the feasibility, and while this isn't the a blocker or priority for the JIT project (which is consuming most of our cycles), it is an important step in the overall Ecstasy project.

lagergren:
@ggleyzer addressed in b94e1bf46.

The pushed changes add the natural `ecstasy.reflect` support that the platform upload path needs:

- `FileTemplate.Kind` with `Single`, `Library`, and `Xecable`
- `FileTemplate.kind`, `moduleNames`, and `bundle`
- native-backed `ModuleTemplate.fingerprint`, so embedded real modules and dependency fingerprints are no longer inferred by the old heuristic

On the repository side, I did not add a new `ModuleRepository` API in this pass. Instead, `FileRepository` now reads the constant-pool-free metadata header and can expose bundled module names/versions cheaply through the existing repository surface. The platform can use `Container.loadFileTemplate()` and then branch on `FileTemplate.bundle` / `kind` / `moduleNames` for the upload scenario you called out.

I also updated the PR description with the significant review-response changes and left platform-side upload processing as an explicit follow-up.

lagergren:
@cpurdy addressed in b94e1bf46.

Summary against the numbered review items:

1. Added the Ecstasy reflection surface needed by platform upload handling: `FileTemplate.kind`, `moduleNames`, `bundle`, and native-backed `ModuleTemplate.fingerprint`.
2. Updated `ModuleTemplate.fingerprint` to match the proposed model semantically, but made it native-backed so the Java structure tells Ecstasy whether a module is actually a fingerprint.
3. Added the `Single` / `Library` / `Xecable` terminology as `FileTemplate.Kind`. `xtc bundle` writes multi-module output as `Library`; `Xecable` is represented but the producing workflow remains future work.
4. Added a length-prefixed metadata block immediately after the file magic/version and before the constant pool. It records the kind, real module names excluding fingerprints, and version strings, with no constant-pool references. Disassembly verifies known metadata against the module contents and skips unknown metadata formats.
5. Added `FileStructure.readMetadata(...)` overloads so tools/repositories can inspect the metadata without deserializing the file body. `FileRepository` now uses that path to populate module names and available versions cheaply.
6. Platform-side upload behavior is still called out as a follow-up, but the PR now provides the metadata/reflection API that platform needs to detect and branch on bundles.

I updated the PR description with the significant review-response changes and added focused tests for metadata round-trips, unknown metadata skipping, repository version discovery, and bundle kind tagging.

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTComponentTemplate.java
```diff
@@ -277,7 +277,7 @@ public static TypeComposition getMultiMethodTemplateComposition(Container contai
      * @return the handle to the appropriate Ecstasy {@code ComponentTemplate.Format} enum value
      */
     protected static EnumHandle makeFormatHandle(Frame frame, Component.Format format) {
-        xEnum enumForm = (xEnum) INSTANCE.f_container.getTemplate("reflect.ComponentTemplate.Format");
+        var enumForm = INSTANCE.f_container.getTemplate("reflect.ComponentTemplate.Format", xEnum.class);
```
ggleyzer: I completely disagree with this refactoring. We are writing code for a human reader, not LLMs or compilers. When a person reads the code, we should make it easier, not harder to understand. In this instance, while I like your refactoring of `getTemplate()` API and getting rid of the casts, I strongly believe that 

    xEmum enumForm = INSTANCE.f_container.getTemplate("reflect.ComponentTemplate.Format", xEnum.class);

would read better than "var". What are we trying to accomplish here by using `var` anyway?

lagergren: Not repeating the type twice on the same line, when it is in one place - readable and explicit only. Easier to read, recommended for Java coding standards, and also removing a cast that also makes it harder to factor out the info. Imho this is a well used idiom, and casts in Java should be avoided. Ideally, we should have an API that explicitly requested the right types by function name, and never have any "object" or erasure returns, since that is a bit of a code smell. I find this form increases readability significantly, not having to use my brain to sort out three explicit type names in a statement of code, and I find that most people agree here. This is a human readability enhancement. I don't care about cognitive load on LLMs

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTModuleTemplate.java
```diff
@@ -99,8 +103,11 @@ public int invokeNativeGet(Frame frame, String sPropName, ObjectHandle hTarget,
      * Implements property: modulesByPath.get()
      */
     public int getPropertyModulesByPath(Frame frame, ComponentTemplateHandle hTemplate, int iReturn) {
```
ggleyzer, line 105: What is the point of this splitting one method in two? I know Claude/Codex like this, but why should we...

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTComponentTemplate.java
ggleyzer: I respectfully disagree:
1) It's not easier to read at all; the reader's eyes shouldn't need to go all the way to the right to see what the type of `enumForm` is. 
2) Using `var` is recommended **only** when it improves readability, but it is not meant to replace explicit types everywhere.
3) I wrote earlier, that I liked removing the cast; not arguing that at all

One other point:
There is such a thing as code style. I would not ever go and change a "style" of a class that you had created. But you are changing an existing code to match your personal style. Besides; you're doing it randomly. For example there are many places that say:

    ConstantPool pool = getConstantPool();
    
why not replace it with

    var pool = getConstantPool();
    
On top that, instead of creating a targeted and as succinct as possible a change that is concentrated on the issue at hand, we are having a large change that is 50% refactoring.

ggleyzer: BTW, I realize that you actually didn't write almost any of that by hand. You just have this style in your skill.md file and Claude follows it. It should be very easy to tell Claude not to do it if an existing class doesn't have that style

ggleyzer:
I know that Cam has been reviewing the `asm` part of the change, so I only reviewed the `runtime` part. Outside of the stylistic issues I outlined above, it looks good.

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTComponentTemplate.java
lagergren: Agreed on minimizing non-payload style churn in this PR. I pushed a89dd234e to restore explicit locals in the touched runtime classes while keeping the typed `getTemplate(..., xEnum.class)` accessor and removing the cast.

I still think the typed accessor is worth keeping because it moves the cast to the accessor boundary and makes the expected runtime type part of the call contract. That reduces repeated low-value syntax on a line where the business intent is otherwise apparent. The broader cleanup principle belongs in a separate modernization branch, not mixed randomly into this feature PR.

lagergren:
I agree with minimizing non-payload style churn in this PR, so I pushed a small follow-up that restores explicit locals in the touched runtime classes and removes the unnecessary helper split.

I still think the additive typed accessors are worth keeping. They are not just style: they move the cast to the accessor boundary and make the expected runtime type part of the call contract. That reduces repeated low-value syntax like “declare type, cast to same type, then use it” on a line where the business intent is otherwise apparent.

More broadly, I do think we should keep moving this code toward readable business logic instead of scaffolding-heavy Java idioms where the mechanics dominate the intent. In places where a stream/lambda or a small typed helper makes the operation read directly, I consider that an improvement. For this PR, though, I’m keeping that principle constrained to payload-bearing changes and leaving broader modernization for a separate branch.

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTModuleTemplate.java
lagergren: Agreed; I removed this extra helper split in a89dd234e. It was not carrying enough weight for this PR. The method is back to one implementation, while the typed component accessor remains because it keeps the cast at the accessor boundary instead of scattering it through the business logic.

lagergren:
I would strongly argue that we should postpone @cpurdy point 6. I believe (at least I hope) I have tested this extensively now and I do not think I am breaking compatibility.  It would be, IMHO, far better to get this into master, then update the platform to try to use it, and fix any issues or any errors in thinking about apis/abstraction that we find out are there.  This is a semantic change and a bytecode level change (version), so I would definitely put it in master first and then incrementally improve it if anything is lacking. 

I have, however, done a few platform repo tests and I can put up a POC branch later.  But as usual, if this is backwards compatible and you think it goes in the right direction it should get merged. Gitflow tends to fit better working incrementally, and there is also the time spent having to keep something scheduled to merge rebased and tested.

lagergren:
Follow-up for the remaining executable-bundle work is tracked in #537: https://github.com/xtclang/xvm/issues/537

That issue covers producing a real `FileTemplate.Kind.Executable` artifact, validating that `xec` can run it, and checking whether the platform can run from that artifact shape. I plan to take that on next; this PR should stay focused on landing the library-bundle workflow, metadata format, repository support, and reflection API now.

ggleyzer:
> I would strongly argue that we should postpone @cpurdy point 6. I believe (at least I hope) I have tested this extensively now and I do not think I am breaking compatibility. It would be, IMHO, far better to get this into master, then update the platform to try to use it, and fix any issues or any errors in thinking about apis/abstraction that we find out are there. This is a semantic change and a bytecode level change (version), so I would definitely put it in master first and then incrementally improve it if anything is lacking.
> 
> I have, however, done a few platform repo tests and I can put up a POC branch later. But as usual, if this is backwards compatible and you think it goes in the right direction it should get merged. Gitflow tends to fit better working incrementally, and there is also the time spent having to keep something scheduled to merge rebased and tested.

Agree; as long as the existing bahevior is preserved I see no problem with adding the library support later

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTModuleTemplate.java
```diff
@@ -100,7 +104,7 @@ public int invokeNativeGet(Frame frame, String sPropName, ObjectHandle hTarget,
...
```
ggleyzer, line 105: I don't see a point in splitting this method in two

lagergren:
@cpurdy Your extra commit looks good and it clears up the Version ambiguities well. I have no problems with it.  If everything else is in order, and you think we are fine to go in, please approve the PR and we can merge it. 

I can think of some immediate followups for the API and for enabling "Executable" xtc files, and anything that may fall out when doing more testing than I have done on stuff like single xdk.xtc files and platform.xtc files.  As soon as I have approval, I can merge this to master! 

/M

lagergren:
I audited the two review commits (b77199333, 8005132e0). Most of it is genuinely good — the `VersionTree.contains()` fix for valueless intermediate nodes, the `getEldestSibling()` clone-safety fix, and replacing the assert-only `Collections.unmodifiable*` wrapping (which only protected under `-ea`) are all real improvements, and I've kept every bit of that. But two behaviors regressed, and both have to go back. I've restored them in 8be552ddd and 0f89b4bdd, with regression tests, rather than reverting anything:

**1. CI versions became unconstructable from int parts (8be552ddd).** The guard in `Version(int[], String)` was tightened from `-PREFIX.length` (−6) to `-5` — which matches the *stale comment* above it, not the parser, which still encodes `"ci"` as −6. That's not a latent nit: `VersionTree`'s own iterator rebuilds each `Version` from int parts, so any tree containing a `-ci` version threw `IllegalStateException` on iteration — reachable from `writeFileInfo` (bundling a CI-versioned module), every `getAvailableVersions` consumer, and `withoutBuild()`. Nothing caught it because no test used a CI version; there is one now. The bound is also no longer a magic number — it's derived from the `TEXT` table as `CATEGORY_CI`/`CATEGORY_NONE`, so the guard and the encoding can't drift apart again, and the `NONE` sentinel stays unconstructable as intended.

**2. The module listing went non-deterministic (0f89b4bdd).** `FileInfo`'s canonicalizing constructor collected the module map with `Collectors.toUnmodifiableMap`, which throws away the main-module-first `LinkedHashMap` ordering that both `buildFileInfo` and `readFileInfo` deliberately construct — ordering that bb5029d75 explicitly established on this branch. The scrambling is user-visible: `FileRepository.getModuleNames()` and the Ecstasy reflection module list served through `xRTFileTemplate` varied from run to run (JDK immutable maps salt their iteration order per JVM). And because `Map.equals` is order-insensitive, changing the test assertions from `List.of("App", "Lib")` to `Set.of(...)` didn't just miss the regression — it deleted the only assertion that could have caught it. Deterministic output ordering is a hard requirement for me: golden tests, reproducible bundles, and stable tooling output all depend on it. The constructor now copies into a `LinkedHashMap` wrapped unmodifiable — same read-only guarantee (`ensureReadOnly` on each tree is kept, and the raw `VersionTree` cast is gone), plus the order contract, and the round-trip test asserts the order explicitly again.

Both fixes are additive on top of the review commits — no behavior from them is reverted beyond these two points. `VersionTest`, `FileStructureTest`, and `BundlerTest` pass locally.

cpurdy (approved):

cpurdy:
... and of course the build fails *after* I approve it! 🤣

lagergren:
> ... and of course the build fails _after_ I approve it! 🤣

Dont' worry. Those were racy runParallel tests. I should disable those until the gradle plugin uses the Tool/Embedding API.

PR #529, opened 2026-08-14
Implement blocking TCP client sockets

mrcompator:
## Summary
- Implement `Network.connect` via `java.net.Socket` (`xRTSocket`), replacing the stub that always returned `False`.
- Wire blocking `Socket.in` / `Socket.out` (`nativeReadBytes` / `nativeWriteBytes`). `nativeListen` and async `channel` stay TODO.
- Keep stock `RTSocket` scaffolding; `rawChannel` is `Channel? = Null` so construct succeeds before async I/O exists.
- Review follow-up: `connect` / `read` / `write` use `scheduleIO` + `waitForIO` so they do not pin the service fiber. Failed bind/options/connect closes the Java socket before a handle owns it. Write `offset`/`count` are checked as Ecstasy `Int` and raise XVM `OutOfBounds` instead of a host `IndexOutOfBoundsException`.

## Test plan

Listen/accept is still a stub, so the Ecstasy client cannot echo to itself. Start a tiny listener, then connect.

### Java (`xRTSocketLoopbackTest`)

Calls the same helpers `xRTSocket.connect` / read / write use (`openConnectedSocket`, `readBytes`, `writeBytes`, `invalidWriteRange`). Covers loopback echo, refused connect, bind-failure close, and bad write ranges. Does not boot a full XVM `Frame`; that path is `TestTcpClient`.

```bash
./gradlew :javatools:test --tests org.xvm.runtime.template._native.net.xRTSocketLoopbackTest
```

### Ecstasy (`manualTests/TestTcpClient.x`)

This is the real `Network.connect` + `in` / `out` path.

Terminal 1:

```bash
python3 -c 'import socket;s=socket.socket();s.setsockopt(socket.SOL_SOCKET,socket.SO_REUSEADDR,1);s.bind(("127.0.0.1",9999));s.listen(1);c,_=s.accept();c.sendall(c.recv(64));c.close();s.close()'
```

Terminal 2 — Gradle (uses this branch’s runtime):

```bash
./gradlew :manualTests:runOne -PtestName=TestTcpClient -PtestArgs=127.0.0.1,9999
```

Or `xec` after installing this tree’s XDK (do not use a stock/Homebrew `xec`; that still stubs connect):

```bash
./gradlew :xdk:installDist
./xdk/build/install/xdk/bin/xec -L manualTests/build/xtc/main/lib TestTcpClient 127.0.0.1 9999
```

Expect `ok (127.0.0.1, <local>) -> (127.0.0.1, 9999)`.

- [x] Java helper JUnit (echo, refuse, bind-fail close, write range)
- [x] `TestTcpClient` echo round-trip via this branch’s `xec`
- [ ] Confirm `socket.channel` still TODOs; sync `in`/`out` still work
- [ ] Confirm listen/accept unchanged (still stub)

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
@@ -3,12 +3,15 @@ package org.xvm.runtime.template._native.net;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.SocketException;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 import org.xvm.asm.ClassStructure;
 import org.xvm.asm.ConstantPool;
@@ -111,9 +114,6 @@ public class xRTSocket
         switch (method.getName()) {
         case "nativeReadBytes":
             return nativeReadBytes(frame, hSocket, (int) ((JavaLong) hArg).getValue(), iReturn);
-
-        case "nativeAvailable":
-            return nativeAvailable(frame, hSocket, iReturn);
         }
 
         return super.invokeNative1(frame, method, hTarget, hArg, iReturn);
@@ -164,9 +164,41 @@ public class xRTSocket
      */
     public static int connect(Frame frame, byte[] abRemoteIP, int nRemotePort,
                               byte[] abLocalIP, int nLocalPort, int[] aiReturn) {
-        Socket sock;
+        Callable<Socket> task = () ->
+                openConnectedSocket(abRemoteIP, nRemotePort, abLocalIP, nLocalPort);
+
+        CompletableFuture<Socket> cf = frame.f_context.f_container.scheduleIO(task);
+        Frame.Continuation continuation = frameCaller -> {
+            try {
+                Socket      sock    = cf.get();
+                InetAddress local   = sock.getLocalAddress();
+                byte[]      abLocal = local == null ? new byte[0] : local.getAddress();
+                int         nLocal  = sock.getLocalPort();
+                return INSTANCE.constructSocket(frameCaller, sock, abLocal, nLocal,
+                        abRemoteIP, nRemotePort, aiReturn);
+            } catch (Throwable e) {
+                Throwable cause = unwrap(e);
+                if (cause instanceof IOException) {
+                    return frameCaller.assignValue(aiReturn[0], xBoolean.FALSE);
+                }
+                return frameCaller.raiseException(
+                        xException.makeHandle(frameCaller, cause.getMessage()));
+            }
+        };
+
+        return frame.waitForIO(cf, continuation);
+    }
+
+    /**
+     * Open a client TCP socket with the same options {@link #connect} uses.
+     * On bind, option, or connect failure the Java socket is closed before this returns.
+     */
+    static Socket openConnectedSocket(byte[] abRemoteIP, int nRemotePort,
+                                      byte[] abLocalIP, int nLocalPort)
+            throws IOException {
+        Socket  sock  = new Socket();
+        boolean owned = false;
         try {
-            sock = new Socket();
             sock.setTcpNoDelay(true);
             sock.setKeepAlive(true);
             if (abLocalIP != null && abLocalIP.length > 0) {
@@ -176,17 +208,13 @@ public class xRTSocket
             }
             sock.connect(new InetSocketAddress(InetAddress.getByAddress(abRemoteIP), nRemotePort),
                     CONNECT_TIMEOUT_MS);
-        } catch (IOException e) {
-            return frame.assignValue(aiReturn[0], xBoolean.FALSE);
-        } catch (Exception e) {
-            return frame.raiseException(xException.makeHandle(frame, e.getMessage()));
+            owned = true;
+            return sock;
+        } finally {
+            if (!owned) {
+                closeQuietly(sock);
+            }
         }
-
-        InetAddress local = sock.getLocalAddress();
-        byte[] abLocal = local == null ? new byte[0] : local.getAddress();
-        int    nLocal  = sock.getLocalPort();
-
-        return INSTANCE.constructSocket(frame, sock, abLocal, nLocal, abRemoteIP, nRemotePort, aiReturn);
     }
 
     protected int constructSocket(Frame frame, Socket sock, byte[] abLocal, int nLocalPort,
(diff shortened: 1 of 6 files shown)


ggleyzer:
I just glanced over - looks pretty good. I think a couple of tests would be quite useful; can be something in the manualTests directory

mrcompator:
See PR description for how to execute Java and XTC tests

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
```diff
@@ -0,0 +1,377 @@
...
+        try {
+            sock = new Socket();
+            sock.setTcpNoDelay(true);
+            sock.setKeepAlive(true);
+            if (abLocalIP != null && abLocalIP.length > 0) {
+                sock.bind(new InetSocketAddress(InetAddress.getByAddress(abLocalIP), nLocalPort));
+            } else if (nLocalPort != 0) {
+                sock.bind(new InetSocketAddress(nLocalPort));
+            }
+            sock.connect(new InetSocketAddress(InetAddress.getByAddress(abRemoteIP), nRemotePort),
```
lagergren, line 209: This `connect` is a blocking host call executed on the service context. Existing native I/O paths such as `xRTNameService` and `xOSFile` use `frame.f_context.f_container.scheduleIO(...)` with `frame.waitForIO(...)` so slow DNS/file/network operations do not pin an XVM service fiber. MAybe move connect onto the scheduled I/O path and complete socket construction from the continuation? Deferring this to some review who knows what he is doing rather than me. I could be wrong.

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
```diff
@@ -0,0 +1,377 @@
...
+                    CONNECT_TIMEOUT_MS);
+        } catch (IOException e) {
```
lagergren: After `new Socket()`, any `setTcpNoDelay`, `setKeepAlive`, `bind`, or `connect` failure returns `False` without closing the socket. I would suggest closing the socket in all failure paths until ownership has been transferred to the `SocketHandle`, for example with a `Socket sock = new Socket(); boolean transferred = false; try { ... transferred = true; } finally { if (!transferred) closeQuietly(sock); }` pattern.

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+        if (cBytes <= 0) {
+            return frame.assignValue(iReturn, xArray.makeByteArrayHandle(new byte[0], Mutability.Constant));
+        }
+        try {
+            InputStream in  = sock.getInputStream();
+            byte[]      buf = new byte[cBytes];
+            int         off = 0;
+            while (off < cBytes) {
+                int n = in.read(buf, off, cBytes - off);
```
lagergren: `InputStream.read(...)` can block indefinitely here, and the loop keeps the service fiber occupied until either all requested bytes arrive or EOF occurs. This should follow the same scheduled I/O pattern as the other native blocking operations, then assign the resulting byte array from the continuation.

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+        byte[] ab = xByteArray.getBytes((ArrayHandle) ahArg[0]);
+        int    of = (int) ((JavaLong) ahArg[1]).getValue();
+        int    n  = (int) ((JavaLong) ahArg[2]).getValue();
+        if (n <= 0) {
+            return Op.R_NEXT;
+        }
+        try {
+            OutputStream out = sock.getOutputStream();
+            out.write(ab, of, n);
```
lagergren: This write/flush path is both blocking and range-sensitive. `offset` and `count` come from Ecstasy `Int` values that were narrowed to Java `int` without checking, and invalid ranges will throw a host `IndexOutOfBoundsException` rather than an XVM bounds/argument exception. Please validate `0 <= offset`, `0 <= count`, and `offset + count <= ab.length`, reject values above `Integer.MAX_VALUE`, and schedule the actual write/flush through `scheduleIO`.

javatools/src/test/java/org/xvm/runtime/template/_native/net/xRTSocketLoopbackTest.java
```diff
@@ -0,0 +1,70 @@
...
+                    peer.getOutputStream().write(buf);
+                    peer.getOutputStream().flush();
+                    return buf;
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            });
+
+            byte[] ping = {'p', 'i', 'n', 'g'};
+            try (Socket sock = openLikeNative(new byte[] {127, 0, 0, 1}, port)) {
```
lagergren: This test opens a plain `java.net.Socket` through `openLikeNative`, so it does not exercise `xRTSocket.connect`, native dispatch, socket handle construction, or `Network.connect`. It would still pass if the XVM bridge were broken. Please add an integration test that runs the Ecstasy `Network.connect` path against the loopback server, or refactor this test so it invokes the native template path directly.

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
```diff
@@ -0,0 +1,377 @@
...
+        if (frame.f_context != hSocket.f_context) {
+            return xRTFunction.makeAsyncNativeHandle(method).
+                    call1(frame, hTarget, new ObjectHandle[] {hArg}, iReturn);
+        }
+
+        switch (method.getName()) {
+        case "nativeReadBytes":
+            return nativeReadBytes(frame, hSocket, (int) ((JavaLong) hArg).getValue(), iReturn);
+
+        case "nativeAvailable":
```
lagergren: `nativeAvailable` is declared with zero parameters in `RTSocket.x`, so this `invokeNative1` branch should maybe not be used. I may be wrong. So I guess either remove it or add a short explanation if there is a runtime arity reason for keeping it that I am not seeing right now. Otherwise it makes the native dispatch table harder to audit. Also bit of a nitpick.

javatools_bridge/src/main/x/_native/net/RTSocket.x
```diff
@@ -131,26 +139,77 @@ service RTSocket(SocketAddress localAddress, SocketAddress remoteAddress)
...
+            Byte[] got = nativeReadBytes(count);
+            if (got.size < count) {
+                reachedEof = True;
+            }
+            return got.freeze(inPlace=True);
+        }
     }
 
 
     // ----- SocketInput class ---------------------------------------------------------------------
```
lagergren: Minor cleanup: this section header still says `SocketInput class`, but the section defines `SocketOutput`. Worth fixing while touching the file.

lagergren (changes requested):
The socket implementation is a good start, but I think this needs another pass before merge. The main issues are native blocking I/O running directly on service fibers, socket cleanup on failed connection setup, unchecked `Int` to `int` narrowing/range handling, and a Java test that exercises `java.net.Socket` rather than the new XVM native path.

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
lagergren: This is a bit of a nitpick, but this seems like the ideal place to use var instead of OutputStream in Java. Quite clear what it does already.

lagergren (approved):

mrcompator:
@lagergren @ggleyzer thanks for the review.

Addressed in `c38de5b`:

- `connect` / `read` / `write` now go through `scheduleIO` + `waitForIO`, same as `xRTNameService` / `xOSFile`
- Java `Socket` is closed on bind / options / connect failure before a handle owns it
- Write `offset` / `count` are checked as Ecstasy `Int`; bad ranges raise XVM `OutOfBounds` instead of a host `IndexOutOfBoundsException`
- Dropped the unused `nativeAvailable` branch from `invokeNative1`; fixed the `SocketOutput` section header

Tests:

- JUnit `xRTSocketLoopbackTest` now calls the helpers `xRTSocket.connect` / read / write actually use (not a copy of `new Socket()`). Also covers bind-failure close and bad write ranges
- `manualTests/TestTcpClient` is still the real `Network.connect` + `in` / `out` path (needs a one-shot echo on 9999; commands are in the PR description)

Listen / async `channel` still out of scope.

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
ggleyzer: > This is a bit of a nitpick, but this seems like the ideal place to use var instead of OutputStream in Java. Quite clear what it does already.

I disagree on this one)

javatools/src/main/java/org/xvm/runtime/template/_native/net/xRTSocket.java
ggleyzer: 100%

ggleyzer:
Merged into master manually

PR #534, opened 2026-08-16
Fix runtime enum singleton initialization race

lagergren:
Problem
- `manualTests:runParallel` starts manual test modules concurrently inside one runtime/container, so lazy runtime initialization paths can run at the same time.
- Natural enum construction was publishing the enum constant handle before construction completed. At that point the handle can still be the mutable `:struct`, so another fiber could observe and use it as the enum's final value.
- `SingletonConstant` also allowed the initialization owner/future state to be checked and updated without one atomic critical section, so two service contexts could race around the same singleton.

Original failure
The motivating failure was from `manualTests:runParallel` in https://github.com/xtclang/xvm/actions/runs/31867080587/job/94969815227?pr=525. The important part is that a public reflected enum property expected `ParameterTemplate.Category`, but received the natural enum's construction struct:

```text
Exception during execution of module "TestGenerics" (#6): ecstasy:TypeMismatch: Expected "ecstasy:reflect.ParameterTemplate.Category", actual "immutable ecstasy:reflect.ParameterTemplate.Category.TypeParameter:struct"
    at reflect.RTParameterTemplate.construct(String?, Int, ecstasy:reflect.TypeTemplate, immutable Object?, ecstasy:reflect.ParameterTemplate.Category) (L_SET _native:reflect.RTParameterTemplate.category, #4)
    at reflect.RTMethodTemplate.parameters.get().->(Int) (RTMethodTemplate.x:35)
    at reflect.RTMethodTemplate.parameters.get() (RTMethodTemplate.x:27)
    at reflect.Type.createHasher().isImplemented(String, Int) (Type.x:961)
    at reflect.Type.createHasher() (Type.x:973)
    at reflect.RTType.hasher.calc() (RTType.x:71)
    at annotations.Lazy.get() (Lazy.x:39)
    at reflect.RTType.hashed() (RTType.x:62)
    at maps.HashMap.construct(Int) (HashMap.x:20)
    at collections.HashSet.construct(Int) (HashSet.x:18)
    at reflect.TypeSystem.construct(Array<Module>, Array<Boolean>, Set<String>) (TypeSystem.x:29)
    at reflect.Class.displayName.get() (Class.x:150)
    at reflect.Type.estimateStringLength() (Type.x:673)
    at reflect.Type.estimateStringLength() (Type.x:681)
    at text.Stringable.toString() (Stringable.x:35)
    at testVirtualChild() (generics.x:26)
    at run() (generics.x:8)
    at ^TestGenerics (run())
        =========
    at ^Runner (CallLaterRequest: native)
```

That shape is the race: `ParameterTemplate.Category.TypeParameter` had escaped as a `:struct` while another fiber was reflecting method parameters during `TestGenerics` startup.

Why these call sites are dangerous
- The fixed native/public paths cross from internal template state into user-visible values or native constructor arguments during lazy startup.
- Injection, reflect enum helper values, reflect contribution calls, and array mutability construction can all run while another fiber is initializing the enum singleton.
- Those paths therefore must either wait for enum initialization or return a deferred handle. Returning the raw handle from the template's enum list is not safe for natural enums because that list may temporarily contain construction structs.

What is intentionally left alone
- There are still ugly static `INSTANCE` template fields elsewhere. Most of those are assigned during native container/bootstrap before service fibers are running, and are then used as stable template caches or dispatch shortcuts rather than as lazy singleton value publication.
- That pattern is still not ideal: it is JVM-global, container-coupled, and makes initialization ordering harder to reason about. This PR avoids broad churn there because those sites are not the same race class as publishing partially constructed enum values or non-atomically claiming singleton initialization ownership.
- For paths touched by this fix, reflect enum helpers now resolve through the current frame container instead of relying on the static instance.

Fix
- Publish enum singleton handles only after construction finalizes.
- Guard `SingletonConstant` initialization owner/future transitions with a monitor while keeping the completed handle volatile for cheap reads.
- Route public/native enum lookups through initialized/deferred handles, including injection, reflect access/form/format helpers, array mutability forwarding, and reflect contribution calls.
- Resolve deferred enum arguments before passing them into native calls that expect finalized handles.
- Add checked typed `getTemplate(..., Class<T>)` helpers and convert only the already-touched `xEnum` template lookups to those helpers for readability. These helpers use `Class.cast(...)`, so they do not require unchecked-cast suppression.
- Add a focused `SingletonConstant` concurrency regression test.

Verification
- `./gradlew :javatools:compileJava --console=plain --warning-mode=all --no-daemon`
- `./gradlew :javatools:test --tests org.xvm.runtime.SingletonConstantTest --console=plain --warning-mode=all --no-daemon`
- `./gradlew :javatools:test --console=plain --warning-mode=all --no-daemon`
- `./gradlew :manualTests:runParallel --console=plain --warning-mode=all --no-daemon` (3 passes for the race fix; 1 final pass after the typed helper cleanup)

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/constants/SingletonConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/SingletonConstant.java
@@ -123,10 +123,13 @@ public class SingletonConstant
         // INITIALIZING to anything or from a struct to an immutable value
         assert handle != null;
 
-        CompletableFuture<ObjectHandle> cfInitialized = m_cfInitialized;
-        m_handle            = handle;
-        m_fiberInitializing = null;
-        m_cfInitialized     = null;
+        CompletableFuture<ObjectHandle> cfInitialized;
+        synchronized (this) {
+            cfInitialized       = m_cfInitialized;
+            m_handle            = handle;
+            m_fiberInitializing = null;
+            m_cfInitialized     = null;
+        }
 
         if (cfInitialized != null) {
             cfInitialized.complete(handle);
@@ -145,12 +148,14 @@ public class SingletonConstant
 
         // initialization is entered from the main context; record which fiber owns the attempt, so
         // other fibers would wait without being mistaken for recursion
-        if (m_fiberInitializing != null) {
-            return false;
-        }
+        synchronized (this) {
+            if (m_handle != null || m_fiberInitializing != null) {
+                return false;
+            }
 
-        m_fiberInitializing = fiber;
-        return true;
+            m_fiberInitializing = fiber;
+            return true;
+        }
     }
 
     /**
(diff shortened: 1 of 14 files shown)


lagergren:
This is a bigger symptomatic problem trying to spawn multiple containers in the same runtime that I have to fix for LSP response times too, so I am closing that and will attempt to put up a wider (but also shallower) PR that pipes types and templates that now race through s specific "INSTANCE" handling module. My current draft of this works and solves my testing problem, and it will also remove the sporadic startup crashes from the parallel run in manualTests. Discussed it a bit with Cam yesterday and I will try to bring up the "complete" change for this today.

PR #538, opened 2026-08-21
Make compiler counters atomic (trivial change)

lagergren:
## Why?

The reentrancy audit found four compiler-wide static `int` counters used to allocate synthetic labels and code-container ids. They are independent from the runtime container ownership work, but they are still process-global mutable state and can lose updates when incremental or parallel same-JVM compilation exercises AST code generation concurrently.

## How?

This replaces the counters with `private static final AtomicInteger` fields.

Sequential behavior matches master:

- former `++counter` sites now use `incrementAndGet()`
- former `counter++` sites now use `getAndIncrement()`

So ordinary single-threaded label/id numbering is preserved.

## Tests

- `./gradlew :javatools:test --tests org.xvm.compiler.ast.CompilerCounterAtomicTest --console=plain --warning-mode=all`
- `./gradlew :javatools:test --console=plain --warning-mode=all`

`CompilerCounterAtomicTest` verifies the field shape, preserved sequential values, stable per-node memoization, and bounded concurrent uniqueness.


cpurdy (approved):
A couple of small notes:

- In this case, I'd probably have retained the old names for the counters (because in my mind an atomic wrapper around a volatile int is still functionally an int type), but opinions on this can differ
- In this case, the test seems a bit overboard

lagergren:
> A couple of small notes:
> 
> * In this case, I'd probably have retained the old names for the counters (because in my mind an atomic wrapper around a volatile int is still functionally an int type), but opinions on this can differ
> * In this case, the test seems a bit overboard

We really don't have time to go into naming, when trying to fix any process global state that blocks LSP and incremental reuse, but since this turned from a dangerous mutable static into a static final, the naming convention seems to imply that it should be UPPER_SNAKE_CASE as it is now a constant. 

You also said we should move away from the confusing Hungaring notation last week.

PR #539, opened 2026-08-21
Fix concrete utility this-escape hazards

lagergren:
## Why?

This started as a fix for two concrete `javac -Xlint:this-escape` hazards in
`javatools_utils`, found while auditing runtime reentrancy / ownership around
container-local state and unsafe constructor publication. The broader failure
mode: when a constructor publishes `this`, or calls subclass-overridable
behavior before construction has completed, another code path can observe a
partially initialized object. That hides as a rare race for a long time, then
shows up as stale owner state, wrong container data, or state leaks between
otherwise independent runtime executions — exactly what makes same-JVM runtime
reuse brittle.

Review feedback reshaped both fixes:

- `AbstractConverterMap` called overridable view factory methods from its base
  constructor — the classic constructor virtual-dispatch hazard: a subclass
  override can run while the subclass is still only half-built.
- `CooperativelyCleanableReference` registered `this` in a static `KEEP_ALIVE`
  set from its constructor. Review concluded this particular escape was benign
  in practice, and since the class has no production callers anywhere in the
  repository, it is now simply deleted rather than redesigned.

## How?

- `AbstractConverterMap` no longer creates its views during base construction.
  `keySet()`, `values()`, and `entrySet()` compute their view at most once on
  first access and cache it in a **final** `Lazy.Bound` holder field:

  ```java
  private final Lazy.Bound<AbstractConverterMap<K, V, SK, SV>, Set<K>> keys =
          Lazy.ofBound(AbstractConverterMap::newKeySet);

  @Override
  public Set<K> keySet() {
      return keys.get(this);
  }
  ```

  The unbound method reference captures no instance, so nothing touches
  `this` during construction; the owner is supplied at access time. The holder
  publishes the computed view through release/acquire ordering, so a
  concurrent reader always observes a fully constructed view no matter how a
  subclass implements its views, computes at most once (no duplicate views
  under a racy first access), and leaves no mutable cache field in the class.

  To be precise about the memory-model claim: this is the *same* publication
  guarantee that volatile fields would provide — release/acquire is not a
  stronger fence than volatile (formally slightly weaker, and fully sufficient
  for safe publication). Relative to the original plain-field code it
  introduces a real happens-before edge where there was none. What is strictly
  stronger than a volatile cache field is structural, not the fence:
  compute-at-most-once makes duplicate views impossible, the `final` holder
  field leaves nothing mutable to reason about and stays visible per JLS 17.5
  final-field semantics even if the map itself is racily published, and the
  unbound method reference makes constructor capture of `this` impossible by
  construction.
- `CooperativelyCleanableReference` and its test are deleted. If the class is
  ever needed, it should be designed together with its first real use case.
- New API: `Lazy.Bound` / `Lazy.ofBound(Function<O, T>)` — a thread-safe
  lazily computed value that receives its owner at access time and binds to
  the first owner it computes from, so a `final` lazy field never needs a
  supplier that captures `this` during construction. Per review it was renamed
  from `Owner`/`ofOwner`, uses a `VarHandle` (obtained via the existing
  `VarHandles` helper) instead of an `AtomicReference`, and the unused
  reflective `get(Object, Class)` overload was dropped. The converter map view
  caches above are its first users; it is the intended shape for future
  conversions of the widespread `if (field == null) field = compute()` getter
  pattern.

## Behavior and Compatibility

- Map views remain live and map-backed; only the point of creation moved from
  the constructor to first access. Repeated accessor calls return the same
  cached view instance, now guaranteed unique even under concurrent first
  access.
- `Container`'s `Collections.newSetFromMap(new ConcurrentWeakHasherMap<>())`
  path captures `keySet()` once during set construction and is unaffected.
- Normal map operations (`get`, `put`, `remove`, `containsKey`) are untouched;
  the added cost is one acquire-mode read per view accessor call and three
  small eagerly allocated holder objects per map instance.
- Removing `CooperativelyCleanableReference` deletes only unused code.

## Tests

- `AbstractConverterMapTest`: regression coverage that view creation does not
  happen during base construction; view caching and liveness checks; and a
  concurrency test proving the overridable view factory runs exactly once and
  all racing threads observe the identical view under concurrent first access.
- `WeakHasherMapTest`: views from the affected utility-map path remain live
  against the backing map.
- `LazyTest`: coverage for `Lazy.Bound` — compute-at-most-once, owner binding,
  null values, and concurrent access.

Manual verification:

```bash
./gradlew :javatools_utils:test
```

passes locally.


cpurdy (dismissed):
I'm concerned about setting LLMs loose on a codebase without specific goals that are aligned with what we need to focus on. In this case, it's changing code that isn't even being used (CooperativelyCleanableReference) and making some relatively stupid changes to code that is being used (AbstractConverterMap). I think you definitely need Mark Falco to review and discuss with you the CooperativelyCleanableReference changes, so there's no way I would be able to approve these changes myself. Furthermore, on AbstractConverterMap, making some of those fields private and lazily instantiating and caching them in a thread unsafe manner would probably be fine, but this change just seems to follow a trend of "repeatedly allocating and initializing unnecessarily even in a hot loop should be fine, right?" which isn't based on the reality that I'm accustomed to, unless Hotspot has introduced magic since the last time I tested things like this. All that to say: There's room for improvement in the code that you had Copilot chomp on, but I don't think that Copilot is demonstrating any wisdom in its changes here.

lagergren:
Let me restate the case for this change more clearly.

**What this fixes.** These are real `this-escape` violations, not stylistic ones. The goal is simple: no code path should ever be able to observe a half-initialized object. Constructor escapes are exactly how that happens — ideally they would never have compiled in the first place.

**Performance.** I benchmarked the change and verified the implications: no measurable difference. This code runs during Container setup, not on a hot path. The cost today is zero, and what we buy is the guarantee that a partially constructed object cannot leak.

**Severity.** Taken alone this is a "should fix", not a "must fix". But it is part of a foundational effort: master currently has abundant data leakage between containers, and I have reproducers for several places that are unsafe even in fully single-threaded use. I will keep putting those up as small, bite-sized PRs — the changes are not deep, but they are wide, so they have to land incrementally.

**Why it matters right now.** The single-threaded execution model will not hold up performance-wise for the LSP, so this class of problem has to be eliminated step by step. The follow-on reentrancy work is blocked until pieces like this can land.

**The ask.** Unless there is a concrete "this breaks something that works today" objection, I would like to get this merged so the incremental cleanup can proceed.

lagergren:
There are also not LLM changes - these are my changes, based upon this-escape violations. There are so many issues with this in the runtime and compiler right now that even a change to the single threaded model can change it, pick the wrong owner for a resource and other things. I will try to incrementally put up PRs and an action plan. It is not going to be a deep and complicated change, but sadly, it's going to be very wide. However - it can be done incrementally, and I desperately need some reentrancy stability to be able to continue at all with language support. 

Some of the patterns that I have run into in this rabbit hole are very problematic, and would not have cost any extra effort to write in a manner that would make the code automatically reentrant, but may take some cycles to figure out how to get rid of at this stage. However - I am confident that most of them can be eliminated in discrete PRs, but we seriously need to get them through as they appear, so unless you have any obvious "this breaks anything that works" issues with this PR, it should be merged, or I cannot begin to move that way.

cpurdy:
So for the map one, let's just lazy instantiate the "child" data structures on demand, and cache them as private fields (I'm not worried about accidentally creating two of the same). And I agree that the current code is not ideal.

cpurdy:
For the other one, to the best of my knowledge that class isn't used **anywhere**, which means you can discuss it at length with Mark (and the rest of us, too) so that we at least have a good plan in place by the time we think about using it. I can't recall why Mark did what he did, and maybe he can't either, but I'd feel a lot better having him review that part of the change compared with ignorant me.

javatools_utils/src/test/java/org/xvm/util/converter/AbstractConverterMapTest.java
```diff
@@ -0,0 +1,93 @@
...
+        ConstructorSensitiveMap map = new ConstructorSensitiveMap();
+
+        Set<String> keys = map.keySet();
+        Collection<String> values = map.values();
+        Set<Entry<String, String>> entries = map.entrySet();
+
+        assertNotSame(keys, map.keySet());
+        assertNotSame(values, map.values());
+        assertNotSame(entries, map.entrySet());
+
```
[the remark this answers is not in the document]
lagergren: Fixed in 97bed1dc1: the test now asserts the caching contract (assertSame plus liveness) instead of constraining view identity.

javatools_utils/src/test/java/org/xvm/util/WeakHasherMapTest.java
```diff
@@ -37,6 +37,24 @@ void shouldRemoveFromKeys() {
...
+    void shouldReturnLiveFreshViews() {
+        Map<Integer, String> map = new WeakHasherMap<>(Hasher.natural());
+
+        var keys = map.keySet();
+        var values = map.values();
+        var entries = map.entrySet();
+
+        assertNotSame(keys, map.keySet());
+        assertNotSame(values, map.values());
+        assertNotSame(entries, map.entrySet());
```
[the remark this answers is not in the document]
lagergren: Identity assertions were already dropped in 34542326d; 97bed1dc1 renames the test to shouldReturnLiveViews so the name stops implying fresh instances. Only liveness is asserted.

javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java
```diff
@@ -15,31 +15,13 @@ public abstract class AbstractConverterMap<K, V, SK, SV> implements Map<K, V> {
...
      * Construct a {@link AbstractConverterMap}.
      *
      * @param storage the backing store
      */
     protected AbstractConverterMap(Map<SK, SV> storage) {
         this.storage = storage;
-        keys = newKeySet();
-        values = newValues();
-        entries = newEntrySet();
     }
```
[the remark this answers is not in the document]
lagergren: 97bed1dc1 reintroduces the fields as private lazy caches per review. The old protected fields were an implementation detail; subclasses interact through the accessors and the overridable factories. javatools_utils does not promise binary compatibility to external subclasses, so keeping them exposed-but-deprecated is not warranted.

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
```diff
@@ -68,17 +68,33 @@ public class CooperativelyCleanableReference<V> extends WeakReference<V> {
...
+     * Create and register a {@link CooperativelyCleanableReference} for a given referent.
+     *
+     * @param referent the referent to manage
+     * @param cleaner  the function to run once the referent is unreachable, this function must not reference the referent
+     *
+     * @return the registered reference
+     */
+    public static <V> CooperativelyCleanableReference<V> create(V referent, AutoCloseable cleaner) {
+        CooperativelyCleanableReference<V> ref = new CooperativelyCleanableReference<>(referent, cleaner);
+
```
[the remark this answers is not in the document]
lagergren: Addressed in b7f61e151.

lagergren:
Implemented the lazy caching you suggested in 97bed1dc1: keySet()/values()/entrySet() now create their view on first access and cache it in a private field — the same benign-race idiom as java.util.AbstractMap, and safe under the race because the view types keep all state in final fields. No per-call allocation, and the overridable view factories still never run during base construction, which was the point of the this-escape fix (the constructor-sensitivity regression test still guards that).

On CooperativelyCleanableReference: since it is not used anywhere today, the realistic options are (a) delete it, or (b) take this trivial, human-verifiable transform — the constructor stops publishing `this`, and a static factory registers the reference only after construction completes. I suggest we keep the transform and revisit later whether the class is needed at all; there is nothing to design-review until something actually uses it.

lagergren:
Added `Lazy.ofOwner` in afb6d43c3 as a purely additive API — nothing on master touches `Lazy`, so it lands silently. The converter map deliberately keeps the plain private lazy fields you asked for; no behavior change there.

Why `ofOwner` would be the better shape *if* we later convert those fields: the plain-field idiom is correct today, but its correctness rests on three facts someone has to re-verify on every future change — the views must stay final-field-only (that is what makes racy publication safe), duplicate creation must stay harmless, and creation must stay out of the constructor. `Lazy.ofOwner` turns all three into type-level guarantees: the holder field itself becomes `final` (no mutable field left to reason about), compute-at-most-once removes the duplicate-view caveat entirely, and the owner is passed at access time (`views.get(this)`), so no supplier captures `this` during construction — plain `Lazy.of(this::newKeySet)` would quietly reintroduce the very capture this PR removes. The trade is one eagerly allocated holder per field per map instance, which is exactly why the conversion is not folded in here: if we decide it is worth it, it is a separate three-line, trivially reviewable diff.

lagergren:
I did add the Lazy.ofOwner API thing, because that can be used for pretty much any field that acts as a cache, and it solves several real problems where I am trying to trivially modify XTC to run concurrent or sequential containers in the real VM. I have not applied it to the cached fields to keep the change at bare minimum, but this is the correct "we are sure this is computed after X and seen by everyone" to use without synchronization if you want to implement a cached field. 

There are literally hundreds of mutable fields in XTC where a getter checks if (field == null) and then writes a mutable cached value. This is not a good idiom, since it is not memory model safe at all, should the visibility of such fields escape. Using the Lazy class with final field assignments would improve code quality a lot, and something that is meant to be final except that it is lazily computed/cached CAN be final. This is a huge readability and safety thing, and it also trivially fixes several cached fields computed in getters that actually DO cause problems if you want to run sequential or parallel containers in the same process. The cost is nothing, and it can be gradually migrated into.

cpurdy:
Mark's on vacation be he's going to look at the other changes later today, I hope.

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
```diff
@@ -68,17 +68,32 @@ public class CooperativelyCleanableReference<V> extends WeakReference<V> {
...
      * Construct a {@link CooperativelyCleanableReference} for a given referent.
      *
      * @param referent the referent to manage
      * @param cleaner  the function to run once the referent is unreachable, this function must not reference the referent
      */
-    public CooperativelyCleanableReference(V referent, AutoCloseable cleaner) {
+    private CooperativelyCleanableReference(V referent, AutoCloseable cleaner) {
         super(referent, clean(QUEUE[ThreadLocalRandom.current().nextInt(QUEUE.length)]));
         this.cleaner = Objects.requireNonNull(cleaner, "null cleaner");
-        KEEP_ALIVE.add(this);
```
markandrewfalco, line 80: I get why the general pattern has issues. But that doesn’t make all usages of “premature this” unsafe. This usage is safe, so why change it?

javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java
```diff
@@ -16,19 +16,24 @@ public abstract class AbstractConverterMap<K, V, SK, SV> implements Map<K, V> {
     private final Map<SK, SV> storage;
 
     /**
-     * The converter keySet.
+     * Lazily created and cached key set view. The view is created on first access rather than
+     * during base construction, because the view factories are overridable and must not run while
+     * a subclass is still initializing. The plain private field is the same benign-race caching
```
markandrewfalco: Nit, I'm not a fan of code comments which describe the PR change, I.e. the reason for the change he. Those should be PR comments and code comments should describe the current impl.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
```diff
@@ -275,6 +297,93 @@ public boolean isComputed() {
...
 
+    /**
+     * Thread-safe lazy implementation for values computed from an explicit owner passed at access
+     * time. This preserves final-field lazy caching without constructing a supplier that captures
+     * the owner from its constructor.
+     */
+    public static final class Owner<O, T> {
+        private static final Object UNSET = new Object();
+
+        private final AtomicReference<Object> valueRef = new AtomicReference<>(UNSET);
```
markandrewfalco: Personally I'd avoid the AtomicRefence overhead of an extra object and just use VarHandle to achieve the atomic access. We have a VarHandles helper class to make obtaining the VarHandle a trivial one liner.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
```diff
@@ -275,6 +297,93 @@ public boolean isComputed() {
         }
     }
 
+    /**
+     * Thread-safe lazy implementation for values computed from an explicit owner passed at access
+     * time. This preserves final-field lazy caching without constructing a supplier that captures
+     * the owner from its constructor.
+     */
+    public static final class Owner<O, T> {
```
markandrewfalco: Should this implement Lazy?

javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java
```diff
@@ -172,17 +174,20 @@ public void clear() {
 
     @Override
     public Set<K> keySet() {
-        return keys;
+        var keys = this.keys;
+        return keys == null ? this.keys = newKeySet() : keys;
```
markandrewfalco: Unless I'm missing something you're introducing a thread safety issue here. If the container has reached what appears to the user to be a read-only stated, this lazy publication could result in one reader seeing a partially initialized "view". Also I believe these abstract converter maps get (or can be) used as the base to some concurrent containers, so more reason to publish through either a final, volatile, or Lazy.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
markandrewfalco: - I'm struggling a bit with the "Owned" meaning here. I'd suggest considering "Bound"

javatools_utils/src/main/java/org/xvm/util/Lazy.java
```diff
@@ -275,6 +297,93 @@ public boolean isComputed() {
...
+         * for reflective or wildcard callers that naturally hold a {@code Lazy.Owner<?, ?>}; normal
+         * typed code should prefer {@link #get(Object)}.
+         *
+         * @param owner the owner used to compute the value on first access
+         * @param clz   the expected result type
+         * @param <R>   the requested result type
+         *
+         * @return the value as the requested type
+         */
+        public <R> R get(Object owner, Class<R> clz) {
```
markandrewfalco: Seems it should be Class\<R extends T\>

javatools_utils/src/main/java/org/xvm/util/Lazy.java
```diff
@@ -275,6 +297,93 @@ public boolean isComputed() {
...
+        private void checkOwner(O owner) {
+            O ownerExisting = this.owner;
+            if (ownerExisting != owner) {
+                throw new IllegalArgumentException(
+                        "Owner lazy value already belongs to a different owner");
+            }
+        }
+
+        @SuppressWarnings("unchecked")
+        private T getOwner(Object owner) {
```
markandrewfalco: I'm not sure why this variant exists, and would suggest dropping it, but getOwner certainly makes it sound like it would return the Owner. Maybe it should be "getByOwner", or better still "getUnchecked"

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
```diff
@@ -68,17 +68,32 @@ public class CooperativelyCleanableReference<V> extends WeakReference<V> {
...
 
+    /**
+     * Create and register a {@link CooperativelyCleanableReference} for a given referent.
+     *
+     * @param referent the referent to manage
+     * @param cleaner  the function to run once the referent is unreachable, this function must not reference the referent
+     *
+     * @return the registered reference
+     */
+    public static <V> CooperativelyCleanableReference<V> create(V referent, AutoCloseable cleaner) {
```
markandrewfalco: Nit, I prefer "of" for such factory methods.

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
```diff
@@ -68,17 +68,32 @@ public class CooperativelyCleanableReference<V> extends WeakReference<V> {
      */
     private final AutoCloseable cleaner;
 
+    /**
+     * Create and register a {@link CooperativelyCleanableReference} for a given referent.
```
markandrewfalco: I wouldn't mention "register" here is is sort of an implementation detail and nothing that the user needs to be aware of.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
lagergren: This has nothing to do with the change. This is a mechanism I have added to get rid of the mutable getter pattern where cached things are loaded in getters with a mutable field null check, and this is sugar to replace it so we can get final fields but still du cache fields and that is going to be useful soon. Nothing in this branch is using it. The naming doesn't really matter at this point, I can rename it later.

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
lagergren: The alternative would be to add a suppress warning after verifying it is safe. I want the linter to fail the build if we introduce new this-escapes since we have some really unsafe ones, which will make the developer careful consider whether it is safe, whether it should be refactored to just not escape this, which is easy when writing it but not later, and then if it is safe force the developer to suppress it with an explicit suppression and a reason.  The existing usage this escapes are causing all kinds of major havoc for me doing stuff in parallel.

lagergren:
As per Mark's, comment, @cpurdy - If you insist of mutable cache fields for something that at the moment doesn't need mutable cache fields, they do need to be volatile.  This shows exactly the trouble of using mutable fields for caching in Ecstasy, and there are hundreds of them, where some of them would have real thread safety implementations a non trivial reentrancy issues. I Would strongly favor not having these cached fields as all, since trivially again, from the code, there is currently nothing that exercises them. Should we end up with runtime regressions if these entry set getters are being used more widely, then it is a VERY simple this to infer why with JFR and then introduce the caching mechanism back. 

This completely proves my point that you have to provide every explicit seen before/known seen after edge you absolutely can to ensure that we the memory model gives us the guarantees it can. This is exactly what most of the brittle things in XTC that fail reentrantly when I am doing LSP stuff in an existing context/VM stem from, so I rest my case.

javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java
lagergren: Agreed, and fixed in 65e6fcd2d: the three cache fields are now volatile. Strictly speaking the plain-field version was safe today only because every cached view keeps all of its state in final fields (that includes the HasherMap/WeakHasherMap overrides and javac's final synthetic outer refs), but that invariant is unenforceable across overridable view factories — which is exactly your point. Volatile makes the publication guarantee unconditional, and the worst case under a racy first access is a harmless duplicate view, since the views are stateless facades over the same backing map.

As per my comment in the main thread: I'd still prefer to have no cache fields at all until something measurable needs them, but as long as we cache, volatile is the floor.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
lagergren: Done in 65e6fcd2d — the holder now uses VarHandles.of(Bound.class, "value") with getAcquire/setRelease instead of an AtomicReference, so no extra object per field.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
lagergren: Renamed in 65e6fcd2d: Owner → Bound, ofOwner → ofBound (keeps the of* factory family consistent). The docs now read as "a lazy value bound to the first owner it computes from".

javatools_utils/src/main/java/org/xvm/util/Lazy.java
lagergren: Dropped the overload (and the private getOwner helper) entirely in 65e6fcd2d, per your suggestion on the other thread. For the record, Class<R extends T> would have made it uncallable from the wildcard contexts it existed for — a caller holding a Lazy.Bound<?, ?> can't name the captured T — but with no callers yet, deleting beats debating. Trivial to re-add alongside a real caller.

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
lagergren: Moot as of 65e6fcd2d — the class and its test are deleted, since nothing in the repo uses it. If it comes back it gets designed together with its first real use case (and an of() factory).

javatools_utils/src/main/java/org/xvm/util/CooperativelyCleanableReference.java
lagergren: Moot as of 65e6fcd2d — the class is deleted.

lagergren:
Pushed 0e863fedb, which replaces the volatile view cache fields with **final `Lazy.Bound` holders** — partly to illustrate the pattern I keep advocating, since this is exactly the shape I want the hundreds of `if (field == null) field = compute()` getters in XTC migrated towards.

**TL;DR on the guarantees, stated precisely:** the publication guarantee is the *same* as the volatile version — release/acquire is not a stronger fence than volatile, and both are fully sufficient; the real new happens-before edge is relative to the original plain-field code, which had none. What is strictly stronger than a volatile cache field is *structural*: compute-at-most-once (duplicate views impossible), a `final` field (nothing mutable left, visible per JLS 17.5 even under racy publication of the map), and constructor-capture immunity. Details below.

```java
private final Lazy.Bound<AbstractConverterMap<K, V, SK, SV>, Set<K>> keys =
        Lazy.ofBound(AbstractConverterMap::newKeySet);

@Override
public Set<K> keySet() {
    return keys.get(this);
}
```

The method reference is *unbound*: the field initializer stores only a `Function` and captures no instance, so nothing can touch `this` during base construction — the owner is supplied at access time. Note that `Lazy.of(this::newKeySet)` would silently reintroduce the very constructor capture this PR removes; `ofBound` structurally cannot.

**Why the guarantee is now iron-clad — being precise about the JMM claims:**

1. *The publication edge.* First access takes a synchronized compute path: exactly one thread runs the factory, writes the owner, then publishes the view with `setRelease`. Every consumer either observes that store via the `getAcquire` fast path — release/acquire ordering makes all writes program-ordered before the release (the fully constructed view, and the owner) visible — or misses it and takes the monitor, which is a happens-before edge by itself. Relative to the **original plain-field code**, which had *no* inter-thread edge at all and leaned entirely on final-field reasoning about the view internals, this introduces a real, explicit before→after edge where previously there was none.

2. *Versus the volatile fields, honestly:* the release/acquire pair is **not a stronger fence than volatile** — volatile is formally sequentially consistent, acquire/release is slightly weaker, and both are fully sufficient for safe publication. The reader-side guarantee ("any view you can observe is fully constructed") is identical. What `Bound` adds is not a stronger fence but guarantees a volatile field *cannot express*:
   - **Compute-at-most-once.** With a volatile (or plain) cache field, two racing first accesses can each construct a view; the field settles on one, but a thread can keep using the loser. That was benign only because the views are stateless — an invariant someone must re-verify on every future change to every subclass view. With `Bound`, duplicates are structurally impossible.
   - **The field is final.** There is no mutable cache field left in the class to reason about, nothing can ever null it out or swap the view later, and JLS 17.5 final-field semantics make the holder itself visible even to a thread that obtained the map through a data race.
   - **Constructor-capture immunity** by construction, per above.

3. *Proof in tests:* `shouldComputeEachViewAtMostOnceUnderConcurrentFirstAccess` gate-releases 16 threads onto a cold `keySet()` and asserts that every thread observed the *identical* view instance and the overridable factory ran *exactly once*; the existing constructor-sensitivity test still proves nothing overridable runs during base construction.

So: the same publication guarantee as the volatile version, strictly stronger structural guarantees, at the price of three small eagerly allocated holders per map instance. The correctness argument stops being a comment paragraph ("safe under the race because the views keep all state in final fields…") and becomes the API contract of the holder. That is the whole point of `Lazy`/`Bound` as the standard replacement for the mutable-getter caching idiom.

javatools_utils/src/main/java/org/xvm/util/Lazy.java
```diff
@@ -275,6 +300,83 @@ public boolean isComputed() {
...
+         *
+         * @param owner the owner used to compute the value on first access
+         *
+         * @return the value
+         */
+        @SuppressWarnings("unchecked")
+        public T get(O owner) {
+            requireNonNull(owner, "owner");
+
+            Object value = VALUE.getAcquire(this);
```
markandrewfalco: I'd avoid acquire/release and go with full violations since you expose the resulting object and acquire/release are just a little bit weaker than users may naturally expect. Oddly they actually perform worse then volatile, unless JDK has been fixed since Java 20 when I last tested.

javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java
```diff
@@ -16,19 +18,29 @@ public abstract class AbstractConverterMap<K, V, SK, SV> implements Map<K, V> {
...
-    protected final Set<K> keys;
+    private final Lazy.Bound<AbstractConverterMap<K, V, SK, SV>, Set<K>> keys =
+            Lazy.ofBound(AbstractConverterMap::newKeySet);
 
     /**
-     * The converter values.
+     * Lazily computed and cached values view; see {@link #keys} for the caching contract.
      */
-    protected final Collection<V> values;
+    private final Lazy.Bound<AbstractConverterMap<K, V, SK, SV>, Collection<V>> values =
```
markandrewfalco, line 36: I'd have just got with volatiles and a null check, no need for the extra wrapper or indirections.

markandrewfalco (approved):

javatools_utils/src/main/java/org/xvm/util/Lazy.java
lagergren: Agreed on using volatile-strength operations here. Done in de491dbe68e: `Lazy.Bound` now uses `VALUE.getVolatile` / `VALUE.setVolatile`, and the docs say volatile ordering instead of release/acquire. I kept the VarHandle helper, so this still avoids the `AtomicReference` wrapper overhead.

javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java
lagergren: I am keeping `Lazy.Bound` rather than spelling this as three volatile cache fields. The reason is the pattern we want to establish: final cache-holder fields on the owner, no supplier or bound method reference capturing `this` during construction, and compute-at-most-once instead of relying on duplicate first views staying harmless. With de491dbe68e the holder now uses full volatile VarHandle operations internally, so the publication guarantee is the same strength as raw volatile fields while the call site keeps the final-field lazy-cache shape.

Issue #541, opened 2026-08-24
Intermittent permanent deadlock in shared per-container Injector under heavy concurrent load


mrcompator:
## Summary

Under sustained heavy concurrent load, the runtime can deadlock permanently (0% CPU, no thread activity, never recovers) inside the shared per-container `Injector` service chain (`PassThruInjector` → `SimpleInjector`/`TestInjector`). The trigger is unrelated code paths performing a **fresh, uncached `@Inject`** on every call from deeply-nested async service call chains; the injector processes requests strictly FIFO with no cycle detection, so an unlucky interleaving creates a genuine circular wait.

This was found while investigating an intermittent hang in `lib_jsondb`'s test suite, but the mechanism is general to the interpreter's dependency-injection machinery and not specific to jsondb, the `Medium` storage model, or xunit — those just happen to combine to reliably reproduce it.

## Reproduction

- Test: `storage.JsonMapStoreTest.probeConcurrentMultiMapTransactionsAgainstMediumModelDoNotDesyncSize` (currently only in a local working copy, not yet committed — happy to attach it as a patch if useful).
- Command (from `manualTests/`, with `xdk/build/install/xdk/bin` on `PATH`):
  ```
  xtc test -L build/xtc/main/lib -o build/xtc/main/lib \
    -t storage.JsonMapStoreTest.probeConcurrentMultiMapTransactionsAgainstMediumModelDoNotDesyncSize \
    src/main/x/jsondb/jsondb_test.x
  ```
  Note: `-t` does not skip the other tests in the class, so each attempt costs ~20-30 minutes of wall-clock time before reaching the target test.
- The test spins up 12 concurrent workers, each running 1500 transactions against two `Medium`-model `JsonMapStore`s, then reads `mapData.keys.toArray()`.
- Reproduced 5 times out of ~8 attempts. Each hang was confirmed via `top` showing sustained **0.0% CPU** for 700+ seconds (a genuine deadlock, not a slow computation), then killed manually.

## Diagnosis process (ruled out, in order)

- **TxManager's commit/backlog scheduling** — a polling watchdog on `currentlyPreparing`/`pendingPrepare.size` showed it fully idle (`NO_TX`, empty) for the entire duration of every hang.
- **Virtual-thread I/O carrier starvation** — `-Djdk.tracePinnedThreads=full` produced zero pinning traces; `jcmd Thread.dump_to_file` showed zero live `IOWorker` virtual threads at hang time.
[...]

PR #542, opened 2026-08-24
Fix keysAt() undercount in JsonMapStore when the last pending mod sorts before the final history key

mrcompator:
## Summary

`JsonMapStore.keysAt()` merges two sorted streams — committed `history` entries and the current transaction's own pending `mods` — to build the live key list, while `size` is computed independently (and correctly) by `sizeAt()`.

In the `Greater` case (the current pending mod's key sorts before the just-pulled `histEntry`), when that mod is the **last** pending mod, the code took the "mods exhausted" branch and broke out of the merge loop without ever classifying `histEntry`. It had already been consumed from the `history` iterator that round, so it was neither added by the `Greater` branch (which only adds the mod's own key) nor recoverable by the trailing "drain the remainder of history" loop (the iterator had already moved past it). That key silently disappeared from the result, leaving `keys.size` exactly one short of `size`.

## Fix

Classify the pulled `histEntry` (check `floor(readId)`/liveness, same as the trailing drain loop does) before breaking out of the merge loop in the "mods exhausted" branch, instead of dropping it silently.

## Test

Added `shouldNotLoseHistoryKeyWhenLastPendingModSortsBeforeIt`, a deterministic (not key-order-dependent by luck) reproducer: seed two committed keys ("B", "C") and then, within a single open transaction, insert one new key ("A") that sorts before both — "A" is the only (and therefore last) pending mod, and "B" is the `histEntry` that gets dropped without the fix.

Verified: fails on `master` (`keys.size` one short of `size`), passes with the fix. Full `JsonMapStoreTest` class: 17/17 passing.

## Note

This is a narrower, self-contained fix extracted from a broader investigation into intermittent `keys.size == size` desyncs under heavy concurrent load. That broader investigation turned up a separate, unrelated deadlock in the interpreter's shared per-container `Injector` service, filed separately as #541 — not addressed here.

### DIFF
--- a/lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
+++ b/lib_jsondb/src/main/x/jsondb/storage/JsonMapStore.x
@@ -398,10 +398,29 @@ service JsonMapStore<Key extends immutable Const, Value extends immutable Const>
         switch (model) {
         case Empty:
         case Small..Medium:
+
+            /**
+             * Determine whether the history contains a live value at the specified read point.
+             *
+             * This check applies only when the current transaction has no modification for the
+             * corresponding key; a current modification shadows the historical value.
+             *
+             * @return True iff the history contains a non-deleted value at readId
+             */
+            Boolean isHistoryVisible(History valueHistory, Int readId) {
+                if (Int txFloor := valueHistory.floor(readId),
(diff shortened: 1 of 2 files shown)


ggleyzer (approved):
Good find! I updated the comments and extracted a common routine

PR #544, opened 2026-08-27
JIT Milestone: Tupple support; improved Ctx handling

ggleyzer:


cpurdy (approved):
Reviewed latest with Gene. All tests pass.

PR #545, opened 2026-08-28
EmbeddingSupport API + lib_runner for Embedding/LSP/tool support

lagergren:
**A review vehicle, not a merge request.** Opened on @cpurdy's branch so the design has somewhere to be commented on. @cpurdy / @ggleyzer: retarget, take it over, or close it as you like.

## What this reviews

`2568d6be4` — `LspSupport` (compile ×2, run ×2, a `Control` handle, the `TC-xx` diagnostic vocabulary), `lib_runner` as the Container-0 supervisor, and per-run consoles as named native resources.

I rewrote this description because the original one was written against `4fb0c8505` and had gone stale: it described `ToolConnector.java` (renamed since), called `lib_runner` a stub, and listed five `// TODO GG` sites. None of those remain — the branch works.

The findings below came from wiring my `XtcEngine` embedding to this branch and running existing tests through it, not from reading.

## One worth answering before this lands

**A standard XDK module cannot run under the runner.** `runner.x:161` uses `BasicResourceProvider`, which has no case for `Directory` or `FileStore`, so `TestFiles` (`manualTests/src/main/x/files.x`, in `testModuleNames` at `manualTests/build.gradle.kts:467`) fails with `Invalid resource: Key: storage, FileStore`. Switching to `PassThroughResourceProvider` fixes availability but gives up per-run isolation, since every run then resolves to container zero's instances. So neither stock provider is right, and the underlying gap is that `runTask(template, repository, consoleId)` has no way to say "this run gets its own resources" — the same gap as the `UnsupportedOperationException` at `LspSupport.java:475`. Detail [in the diff](https://github.com/xtclang/xvm/pull/545#discussion_r3933436017).

## Two things here I'd keep

The `ErrorListener` threaded through every compile/run entry point, and the `TC-xx` code vocabulary with documented `%1`/`%2` params. A long-running host already owns a diagnostic sink, wants messages as they're produced and correlated with its own request, and wants something to switch on other than message text.

Everything else is inline on the diff, and most of it is small.


javatools/src/main/java/org/xvm/api/ToolConnector.java
```diff
@@ -0,0 +1,348 @@
...
+    /**
+     * Internal constructor.
+     */
+    ToolConnector() {}
+
+    /**
+     * Internal singleton implementation.
+     */
+    private static class Singleton {
+        static ToolConnector instance = new ToolConnector();
```
lagergren: `static ToolConnector instance` is not `final`. The holder idiom relies on `static final` for its safe-publication guarantee - without it, a reader can in principle observe a partially constructed instance.

Written exactly once, at class init: is there a reason it needs to be re-assignable?

javatools/src/main/java/org/xvm/api/ToolConnector.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    private static Object LOCK = new Object();
```
lagergren: `LOCK` is not `final`, and this one has teeth: a re-assignable lock cannot guarantee mutual exclusion, because two threads can end up synchronizing on different objects. Everything `synchronized (LOCK)` protects rests on this.

Written exactly once - `private static final Object LOCK = new Object();`?

javatools/src/main/java/org/xvm/api/LspSupport.java
```diff
@@ -0,0 +1,348 @@
...
+    /**
+     * Internal singleton implementation.
+     */
+    private static class Singleton {
+        static ToolConnector instance = new ToolConnector();
+    }
+
+    private static Object LOCK = new Object();
+
+    private boolean          configured;
```
lagergren, line 83: This field and the two below are written exactly once, inside `configure()`, which explicitly refuses to change them afterwards ("configuration has been performed, and cannot be modified"). So they are write-once by construction, but not declared that way.

They are also the fields read WITHOUT holding `LOCK` (see `isConfigured()` and `getConfiguredRepository()`), while the class javadoc promises "thread-safe and concurrent". Those two facts are the same problem: mutable-but-logically-final state published without a happens-before edge.

Collapsing the three into one immutable value fixes both at once:

```java
private record Config(ModuleRepository repo, String injector) {}
private volatile Config config;   // null until configured; ONE atomic publication
```

`configured` then becomes derived (`config != null`) rather than stored, so the two can never disagree, and the unsynchronized reads become safe.

javatools/src/main/java/org/xvm/api/LspSupport.java
```diff
@@ -0,0 +1,348 @@
...
+    private ModuleRepository cfgRepo;
+    private String           cfgInjector;
+    private Connector        connector;
```
lagergren, line 86: `connector` is the one field here with a genuine reason not to be final - it is lazily created in `ensureConnector()`.

Worth confirming it is only ever read inside `synchronized (LOCK)`. It is today; if a future caller reads it outside the lock, it needs to be `volatile`.

javatools/src/main/java/org/xvm/api/ToolConnector.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    /**
+     * @return the Connector instance
+     */
+    private Connector ensureConnector() {
+        synchronized (LOCK) {
+            if (connector == null) {
+                Connector starting = useJit() ? new JitConnector(cfgRepo) : new InterpreterConnector(cfgRepo);
+                starting.loadModule("org.ecstasy.runner");
```
lagergren: Possible bug: this loads `"org.ecstasy.runner"`, but `lib_runner/src/main/x/runner.x` declares `module runner.xtclang.org`. Those names do not match, so this lookup would fail to find the module.

Which one is intended to be the canonical name?

javatools/src/main/java/org/xvm/api/LspSupport.java
```diff
@@ -0,0 +1,348 @@
...
+                configured  = true;
+            }
+        }
+        return this;
+    }
+
+    /**
+     * @return true iff the TooolConnector has been configured
+     */
+    public boolean isConfigured() {
```
lagergren, line 169: `configured` is read here without holding `LOCK`, and it is a non-volatile field written by `configure()` on another thread. There is no happens-before edge, so this can return a stale `false` (or, worse, `true` while `cfgRepo` is not yet visible).

Same for `getConfiguredRepository()` below, and for `verifyConfigured()`. The `volatile Config` shape suggested on the field declaration makes all three correct without taking the lock.

lagergren (commented):
Field-by-field notes, anchored to the code. The short version: **all six non-final fields are written exactly once**, so the class has no genuinely re-assignable state - and the three that are read without the lock are exactly the three that are logically write-once. Making them final/immutable is the same fix as the thread-safety gap, not a separate style preference.

Happy to push a commit for any of this, or leave it entirely to you - it's your branch.

javatools/src/main/java/org/xvm/api/LspSupport.java
lagergren: Retracting the first half of this. `connector` being lazily created is not a reason for it to be non-final. `AbstractConverterMap` already has three `Lazy.ofBound` fields (`javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java:30-43`) that are lazy *and* final, so the idiom is already in the tree and applies here directly. Once `connector` is a final `Lazy.Bound`, the static `LOCK` can be deleted rather than worked around.

The second half asks the wrong question. Confirming the field is only read under `LOCK` is true today (`:122`, and the one other use at `:480` goes through `ensureConnector`), but publication is not the exposure. `ensureConnector()` is public and never calls `verifyConfigured()`, so calling it before `configure(...)` builds a connector on a null `cfgRepo` at `:125-126`. Latent rather than reached: `LspTest` configures in `main` at `:49` before it reaches `ensureConnector` at `:184`.

Both points fold into the thread on the field block above, which is the same lifecycle question. Closing this one.

lib_runner/src/main/x/runner.x
```diff
@@ -0,0 +1,225 @@
...
+
+        void start() {
+            assert container == Null;
+
+            ResourceProvider injector;
+            if (Int consoleId ?= this.consoleId) {
+                @Inject(resourceName=$"console_{consoleId}") Console console;
+                injector = new TaskResourceProvider(console);
+            } else {
+                injector = new BasicResourceProvider();
```
lagergren: `BasicResourceProvider` doesn't supply a file system, so a run that touches one dies here.

Its `getResource` switch handles `Console`, `Clock`, `Timer`, `Random`/`rnd`, `String`, `List<String>`, and enum/`Destringable` values built from string injections. There's no case for `Directory` or `FileStore`, so `curDir`, `rootDir`, `homeDir` and `storage` fall through to `default:` and come back as a deferred exception. (The `HashCollector` and `Linker` imports at the top of that method have no matching cases.)

Wiring this up and running `TestFiles` fails with `Invalid resource: Key: storage, FileStore`. That isn't a module I wrote for the occasion — `manualTests/src/main/x/files.x` injects `@Inject("storage") FileStore` at `:5` and `@Inject Directory curDir` at `:68`, and it's listed in `testModuleNames` at `manualTests/build.gradle.kts:467`, in neither exclusion set. Any module that touches the file system is excluded from the runner as written, which is a large fraction of anything real.

The runner it replaces made the other choice: `manualTests/src/main/x/runner.x:100` declares `service RunnerResourceProvider(Console console) extends PassThroughResourceProvider`. That's the same shape as `TaskResourceProvider` here — a console-supplying subclass — with a different base, and the base is the entire difference.

Pass-through isn't the fix either, though. I tried it: the missing resources come back, and two ownership tests then fail the other way round, because pass-through delegates to the parent, so every run resolves to container zero's instances instead of its own. Availability or isolation: the two stock providers sit at opposite ends, and a host needs both.

So the gap is in the signature rather than in the choice of provider. `runTask(template, repository, consoleId)` has nowhere for a caller to say "this run gets its own file system, rooted here" — no injector, no `rootDir`. It's the same gap seen from the Java side at `LspSupport.java:475`, where `rootDir` and a custom injector throw `UnsupportedOperationException`.

Worth closing for sequential runs on its own merits, and it happens to be one of the things concurrent runs would need anyway. Happy to take it as a separate branch if you'd like — it's an API change, so it wants to be accepted or rejected as a unit rather than patched inline.

lib_runner/src/main/x/runner.x
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * @return a human-readable task status
+     */
+    String taskStatus(Int id) = TaskRegistry.taskStatus(id);
+
+    /**
+     * Mutable task registry.
+     */
+    static service TaskRegistry {
+        private Map<Int, Task> tasks = new HashMap();
```
lagergren, line 102: `tasks` is never emptied. Four references in the module — this declaration, `tasks[id] = task` at `:101`, `contains` at `:125`, `get` at `:130` — and none of them removes.

`Task.container` is never cleared either: `kill()` at `:182` sets `running = False` but leaves the field, and on normal completion `kill()` isn't called at all. So the registry grows by one entry per run for the life of container zero, and each entry pins a `Container`, and through it that container's composition and template caches. That's precisely the case this module exists to serve — a long-lived VM doing many consecutive runs.

Eviction has to be explicit rather than automatic on completion, because the Java side reads `taskResult` and `taskFailure` after `taskRunning` has gone false. So the caller needs a way to say it's finished:

```ecstasy
// TaskRegistry
void forgetTask(Int id) {
    if (Task task := tasks.get(id)) {
        task.release();
        tasks.remove(id);
    }
}

// module surface, alongside taskRunning/taskResult/taskFailure
void forgetTask(Int id) = TaskRegistry.forgetTask(id);

// Task — idempotent, safe to call more than once
void release() {
    if (Container container ?= this.container) {
        if (running) {
            container.kill();
            running = False;
        }
        this.container = Null;
    }
}
```

plus `this.container = Null` on both completion paths in `start()`'s `whenComplete` (`:178`, `:185`) and in `kill()`, so a finished task stops pinning its container even before anyone gets round to forgetting it. The `this.` is needed in `whenComplete` — the local from `start()` shadows the property there.

Fix: [`e886707`](https://github.com/xtclang/xvm/commit/e8867078ce232647ac4a03601f32e5d47fb82a77) on `lagergren/lspapi-review-fixes`, branched off `2568d6be4`. I'm collecting the small fixes from this review there, one commit each, so you can cherry-pick whichever you want rather than take them as a lump. I didn't commit onto this branch directly because you've force-pushed it five times since the 31st and a commit of mine would just get dropped.

There's no test for the registry at all today. "N runs, then assert the registry is empty" would close it.

javatools/src/main/java/org/xvm/runtime/NativeContainer.java
```diff
@@ -398,13 +399,23 @@ private void initResources(ConstantPool pool) {
      * @param key       the injection key
      * @param supplier  the resource supplier
      */
-    private void addResourceSupplier(InjectionKey key, InjectionSupplier supplier) {
+    public void addResourceSupplier(InjectionKey key, InjectionSupplier supplier) {
         assert !f_mapResources.containsKey(key);
```
lagergren, line 403: `f_mapResources` is a `ConcurrentMap` (`:912`), which it is because more than one thread reaches it — and that's what makes this pair of lines a problem. `containsKey` then `put` is two operations, so two threads registering the same key can both pass the assert, and the second silently replaces the first's supplier.

With `-ea` off, which is how this normally runs, the assert isn't even there and the `put` clobbers unconditionally.

`putIfAbsent` does the check and the insert as one, and lets the assert report a real collision rather than hide it:

```java
InjectionSupplier prev = f_mapResources.putIfAbsent(key, supplier);
assert prev == null;

f_mapResourceNames.put(key.f_sName, key);
```

Fix: [`9f2e7f6`](https://github.com/xtclang/xvm/commit/9f2e7f6ccd2d7535109f9c1dfde31609869b4727) on `lagergren/lspapi-review-fixes`.

(The two-map update below stays non-atomic, which I take to be deliberate — the names map is a lookup index, and `removeResourceSupplier` already uses the two-argument `remove` to avoid dropping someone else's entry.)

lib_runner/src/main/x/runner.x
```diff
@@ -0,0 +1,225 @@
...
+            assert Task task := tasks.get(id) as $"Unknown task {id}";
+            return task;
+        }
+    }
+
+    /**
+     * State and control for one application container.
+     */
+    service Task(Int id, ModuleTemplate template, ModuleRepository repository, Int? consoleId) {
+        Boolean running;
```
lagergren: These three are outputs: `Task` writes them, and everything else — `TaskRegistry`, and the Java side through it — only reads. As declared they're writable by anything holding the service, so an outcome can be reported that never happened.

`public/private` says the intent and enforces it, and it's the idiom already used throughout `lib_ecstasy` (47 occurrences, e.g. `HasherMap.x:108`, `AsyncSection.x:54`):

```ecstasy
public/private Boolean running;

public/private Int? result;

public/private String? failure;
```

`status` just below already has this shape — a computed getter with no setter at all — so this makes the other three consistent with it.

Fix: [`b561e3b`](https://github.com/xtclang/xvm/commit/b561e3b0561a51992cb5ed360a59d90049a95f99) on `lagergren/lspapi-review-fixes`.

javatools/src/main/java/org/xvm/api/LspSupport.java
```diff
@@ -0,0 +1,544 @@
...
+    /**
+     * Internal constructor.
+     */
+    LspSupport() {}
+
+    /**
+     * Internal singleton implementation.
+     */
+    private static class Singleton {
+        static LspSupport instance = new LspSupport();
```
lagergren, line 66: The holder idiom's guarantee is that a reader either sees a fully constructed instance or doesn't see the class initialized at all — and that rests on the field being `static final`. Without `final`, the JVM isn't required to give a reader that ordering, so the one property this pattern exists to provide isn't actually in force.

```java
static final LspSupport INSTANCE = new LspSupport();
```

And the constructor on `:60` is package-private with exactly one caller — this holder. `private` states that and keeps it true; nothing extends the class.

Also then the naming should be UPPER_SNAKE_CASE.

Fix: [`6b92b36`](https://github.com/xtclang/xvm/commit/6b92b3603819bc882089b1673d296826f4d3117a) on `lagergren/lspapi-review-fixes`.

This is the narrow finality point only — not the singleton itself, which I don't think should change: `xExternalConsole.register` reads a mutable static `INSTANCE`, and there are 144 templates shaped that way, so the runtime genuinely can't host two connectors today. The singleton here is a symptom of that, not the cause.

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
```diff
@@ -0,0 +1,153 @@
...
+        ctx.setService(hConsole);
+        return hConsole;
+    }
+
+    @Override
+    public int invokeNativeN(Frame frame, MethodStructure method, ObjectHandle hTarget,
+                             ObjectHandle[] ahArg, int iReturn) {
+        ConsoleHandle hConsole = (ConsoleHandle) hTarget;
+        switch (method.getName()) {
+        case "print": { // Object o = "", Boolean suppressNewline = False
```
lagergren, line 83: Worth recording what this override buys, because it's the kind of thing a later cleanup undoes by accident.

Writing to `hConsole.f_out` here means a redirected run never touches `CONSOLE_LOG`. The parent's `PRINT`/`PRINTLN` continuations (`xTerminalConsole.java:239-252`) call `CONSOLE_LOG.log(ach, …)` on every print, and `CONSOLE_LOG` is a `public static final ConsoleLog` — one instance for the whole JVM — backed by:

```java
private final String[] m_asLine = new String[1024];
private int            m_cLines = 0;
private int            m_iLine  = 0;
```

with **zero** occurrences of `synchronized`, `volatile`, `Atomic` or any lock in the entire file (`javatools_utils/src/main/java/org/xvm/util/ConsoleLog.java`). Two threads printing concurrently interleave read-modify-write on those cursors: lost lines, one line's text stored under another's index, a torn buffer in `get`/`render`. Nothing fails loudly, so the scrollback is just silently wrong.

That's a master defect, not this PR's — neither file is touched by this branch, and I'm raising it separately. Two things about it are this PR's business though:

1. **This override is the mitigation.** Anyone later "simplifying" it to delegate to `super` reintroduces the exposure for every hosted run at once. A line of comment here would stop that.

2. **The fallback doesn't have it.** A run with no console id gets `BasicResourceProvider`'s `case (Console, "console")`, which resolves to the parent's terminal console — so it prints through `CONSOLE_LOG` after all. That's the path `runTask(…, Null)` takes today, which is every run the Java side currently starts. So concurrent runs without console ids still interleave on the unsynchronized buffer.

Fixing `ConsoleLog` itself is one class and no API change — synchronize `log`/`size`/`get`/`render`, or swap the ring buffer for a concurrent structure. Worth doing before a model that runs modules concurrently lands, rather than after.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,252 @@
...
+                consoleId = xExternalConsole.register(connector.getNativeContainer(), console);
+            }
+
+            ObjectHandle hModule     = xRTModuleTemplate.makeHandle(main, file.getModule());
+            ObjectHandle hRepository = xCoreRepository.INSTANCE.makeHandle(repository);
+            ObjectHandle hConsoleId  = consoleId == null
+                    ? xNullable.NULL
+                    : xInt64.makeHandle(consoleId);
+            ObjectHandle hTaskId = postRequest("runTask", hModule, hRepository, hConsoleId).join();
+            taskId = ((JavaLong) hTaskId).getValue();
```
lagergren, line 99: A suggestion about the shape of the new API rather than about this line, which is correct as written.

`MainContainer.invokeAsync` returns `CompletableFuture<ObjectHandle>` (`MainContainer.java:250`), so every caller has to re-establish the type by hand. Three sites here, and all three are right:

```java
taskId = ((JavaLong) hTaskId).getValue();                       // :96
        : ((JavaLong) result).getValue());                      // :168
        : ((StringHandle) result).getStringValue());            // :175
```

What's slightly awkward is that the type is known on both sides and dropped only in between. `runner.x` declares `Int runTask(...)`, `Int? taskResult(...)`, `String? taskFailure(...)`; this file knows exactly what it expects. The boundary is the one place that knowledge doesn't exist, so it gets re-asserted with a cast the compiler can't check — and if a signature in `runner.x` changes later, the failure is a `ClassCastException` here rather than a compile error there.

I want to be clear this isn't an just LSP specific thing to fix: it's how the whole runtime is written, and there are ~1,200 casts to a `*Handle` type across `javatools` (`grep -rE '\(\s*[A-Za-z_][A-Za-z0-9_.]*Handle\s*\)'`). Nobody is going to unpick that here, and this PR adds three correct ones to it. We could at least stop it from growing with invokeAsync... 

Something like:

```java
<T extends ObjectHandle> CompletableFuture<T> invokeAsync(String name, Class<T> type, ObjectHandle... args)
```

and/or a couple of small helpers that unwrap `Int` and `String?` once, would put the knowledge in one place and mean the count stops growing at exactly the point where new code is being written. Much cheaper now, while the surface is this small, than once there are twenty call sites.

So not API issues per se, but somewhat offensive to a person working hard to on the Java Language Team building a strongly typed language that isn't used...

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+        return file;
+    }
+
+    private CompletableFuture<ObjectHandle> postRequest(
+            String methodName, ObjectHandle... arguments) {
+        return connector.getMainContainer().invokeAsync(methodName, arguments);
+    }
+
+    private void watch() {
```
lagergren: This poll rediscovers something the Ecstasy side already has.

At 25ms, a run costs ~40 `taskRunning` requests per second into container zero for its whole duration, plus two more at the end (`taskResult`, `taskFailure`). Every one is a cross-service call that exists only to ask "are you done yet". It also bounds latency from below: a run finishing in 1ms is still reported ~25ms later.

Meanwhile `Task.start` already holds the completion (`runner.x:170`):

```ecstasy
@Future Tuple outcome = container.invoke("run", ());
&outcome.whenComplete((tuple, exception) -> { … });
```

So the information the poll is reconstructing is right there. A `waitForTask(id)` on the module surface would give the Java side one future instead of N requests, and `postRequest("waitForTask", …).whenComplete(…)` would replace `watch()` entirely.

I had a go at it to check it's not fantasy. `Ref.x:70` says `@Future` references are allowed to be unassigned, so a `Task` can hold one as state rather than as a local, and

```ecstasy
private @Future Tuple outcome;

Tuple waitForCompletion() = outcome;
```

compiles fine (`:xdk:lib-runner:compileXtc` on a forced rebuild).

**But I'm not proposing a patch, because there's an ordering question I can't answer from the outside.** `Task` registers its own `whenComplete` to record `result`/`failure` and clear `running`. If a caller is also waiting on that same future, I don't know that the task's continuation is guaranteed to run before the waiter is resumed — and if it isn't, the Java side wakes up and reads `taskResult`/`taskFailure` before they've been set, which is a worse bug than the poll. Making the wait complete *after* the recording, rather than handing out the raw `outcome`, is presumably the answer, but that's a call for someone who knows the fiber scheduling guarantees rather than someone reading the source.

Worth doing while `runner.x` is open anyway. Happy to write it if you tell me which shape is safe.

javatools/src/test/java/org/xvm/runtime/LspTest.java
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * This test is a modification of the reproducer from
+ * <a href="https://github.com/xtclang/xvm/issues/543">issue 534</a>
+ */
+public class LspTest {
+    private static File dirLib;
+    private static File dirJavatools;
+    private static Path dirOut;
+
+    static void main(String[] asArg) throws Exception {
```
lagergren, line 45: The scenarios in here are the right ones — compile success and failure, run output, an exception propagating out of a run, latency, pool growth. My only concern is that nothing re-runs them.

The cheapest useful version is nearly free: **annotate the five scenarios `@Test`**. They already assert by throwing `IllegalStateException`, and a throwing test fails, so that alone buys CI coverage without rewriting a single check.

The one thing genuinely tied to `main` is the two path arguments:

```java
dirLib       = new File(asArg[0]);
dirJavatools = new File(asArg[1]);
```

and there's an established answer to that a few files away: `DirRepositoryConcurrentScanTest.java:44-50` resolves `xdk/build/install/xdk/lib` with a `../` fallback and then `assumeTrue(...)`, so it skips cleanly when the XDK hasn't been built rather than failing noisily.

One scenario needs an actual assertion rather than just the annotation. `testPoolGrows` computes `nativePoolSize(...)` on each of its 13 iterations (`i <= 12`) and only prints it. Whatever the intended invariant is — that the pool stops growing once shapes repeat, or that each novel shape costs a bounded number of constants — a printed number can't check it, and it's the one measurement in this file that nothing looks at.

Not urgent for the design conversation, but it's what would make the rest of this checkable rather than a matter of opinion — mine included.

javatools/src/main/java/org/xvm/api/LspSupport.java
lagergren: Coming back to this with the other half, and with a correction to my own comment above.

**The correction.** I named `isConfigured()` and `getConfiguredRepository()` as the unsynchronized readers. It's eight sites, not two: `:93`, `:103`, `:170`, `:177`, `:185`, `:211`, `:459`, `:474`. Only `configure` (`:152`) and `ensureConnector` (`:122`) take `LOCK`.

**The `record Config` half** stands as written: one immutable value behind one `volatile`, so the publication is atomic and "half configured" stops being representable — the flag and the values can't disagree because there is no flag. `configure` then compares whole values, which is what its error message already claims to do.

**The half I hadn't worked out in August** is `connector`, and it reverses what I said on the sibling thread about it having a genuine reason not to be final. It doesn't. As `private final Lazy.Bound<LspSupport, Connector>` it's computed at most once and published with volatile ordering by the holder itself (`Lazy.java:432-457`, double-checked through a `VarHandle`), so `ensureConnector` needs no lock. And once neither `configure`'s readers nor `ensureConnector` need it, the static `LOCK` can be **deleted** rather than worked around — which is the part worth having, since it's a `static` lock guarding instance state, correct only while exactly one instance exists.

`Lazy` isn't a new dependency: `AbstractConverterMap` already uses `Lazy.ofBound` for three fields (`javatools_utils/src/main/java/org/xvm/util/converter/AbstractConverterMap.java:30-43`).

**One deliberate behaviour change.** `createConnector` now states its precondition. `ensureConnector` is public and never called `verifyConfigured`, so calling it before `configure` built a `Connector` over a null `cfgRepo` and failed later, somewhere else. Latent rather than reached today — `LspTest.main` configures at `:49` before reaching `ensureConnector` at `:184` — but it's reachable through a public method.

Written up as [`618e1f9`](https://github.com/xtclang/xvm/commit/618e1f90bc52a39390f1bf8ebdce93400a61d300) on `lagergren/lspapi-config-lifecycle` — its own branch rather than with the small fixes, because it's a design decision you should be able to take or reject as a unit. `:javatools:compileJava` is clean. No test, for the reason in the `LspTest` thread.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Create and start a control for the specified module.
+     */
+    static LspSupport.Control create(Connector connector, ModuleStructure module,
+                                     ModuleRepository repository, PrintStream console,
+                                     ErrorListener errs) {
+        if (!(connector instanceof InterpreterConnector interpreter)) {
+            throw new IllegalArgumentException("An InterpreterConnector is required");
+        }
+        return new InterpreterControl(interpreter, module, repository, console, errs).start();
```
lagergren, line 67: Construct-then-start is why three fields here can't be final — the object exists before the values do:

```java
private volatile Instant started;     // write-once
private          long    taskId;      // write-once
private          Long    consoleId;   // write-once, then nulled to stay idempotent
```

Moving `start()`'s work into the constructor would be the wrong repair: a constructor that boots things fails at construction time for reasons unrelated to what the caller asked for. But the right place already exists — `create` is the factory, and it's the only caller. Give it the work, and hand the constructor finished values.

Then the class has no "constructed but not yet valid" state at all, seven fields are final, and only `running`/`stopped`/`result` stay mutable, which they honestly are: they're the outcome.

Two things fell out of doing it that I didn't expect:

- **`repository` stops being a field.** It's only read while starting — `prepareModule` and the `xCoreRepository` handle — so it becomes a parameter of a now-static `prepareModule` rather than state the control carries for its whole life.
- **`unregisterConsole` no longer needs to null `consoleId`.** That null-write exists to make it idempotent, because it's reachable from both `start`'s catch and `finish`. Once the failure path lives in the factory — before any object exists — the only caller is `finish`, which is `synchronized` and returns early once `running` is false, so it happens exactly once by construction.

There's also a smaller point about `taskId`: today its safe publication rides on the `watch()` submission happening after the write. That's correct, but by accident of ordering rather than by design. `final` makes it correct by construction.

[`980d1a5`](https://github.com/xtclang/xvm/commit/980d1a5b96f8d5d5cfd4e66783a3502383759c1e) on `lagergren/lspapi-control-finality`, +43/−38 in this one file, `:javatools:compileJava` clean. Separate branch from the config one so they can be taken independently.

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
```diff
@@ -0,0 +1,153 @@
...
+import org.xvm.runtime.template.xException;
+
+import org.xvm.runtime.template.text.xString.StringHandle;
+
+/**
+ * The injectable "Console" that prints to the specified PrintStream.
+ */
+public class xExternalConsole
+        extends xTerminalConsole {
+    public static xExternalConsole INSTANCE;
```
lagergren, line 31: Module output now has two mechanisms for one concern, and neither can do the other's job. The CLI prints through `xTerminalConsole`'s `PRINT`/`PRINTLN` continuations, which are `private static final` and close over the static `CONSOLE_OUT`, so they can't be redirected. A hosted run prints through a per-run `PrintStream` in this second template, which can't reach the terminal's `ConsoleLog`.

`xTerminalConsole`'s sink is static only because nothing had ever needed otherwise — it caches a single `m_hConsole`. Giving that template a per-handle sink, defaulting to the terminal writer, would leave the CLI byte-for-byte unchanged and remove the need for this class, its Ecstasy declaration (which adds only a `toString`), and this mutable `public static INSTANCE`.

Two caveats on that, though. It's a change to `xTerminalConsole` — master's code, not yours — so I don't think it belongs in this PR; I'll raise it separately. And it isn't "delete `CONSOLE_OUT`": `DebugConsole` writes to it at `:232`, `:302` and `:2133`, and a terminal debugger writing to the terminal is correct. What would change is the console template's *dependence* on it.

**The part that is this PR's is `PrintStream`.** Three things, measured rather than asserted:

1. **A failed write can't be reported.** Writing to a stream whose `write` throws `IOException` doesn't propagate — `PrintStream` swallows it and sets a flag reachable only through `checkError()`, which nothing here polls. A hosted run writing to a closed pipe just goes quiet.
2. **`println` depends on the caller's auto-flush setting.** `print` flushes explicitly here; `println` doesn't. With `new PrintStream(out)` — `autoFlush` defaults to `false` — over a *buffered* sink, `println` leaves **0 bytes** at the sink until something flushes; with `autoFlush=true` it arrives immediately. `LspTest` passes a `ByteArrayOutputStream`, which isn't buffered, which is why this hasn't bitten yet. It's a requirement the signature can't state.
3. **The charset is whatever the caller happened to construct.** The parent's `PrintWriter` takes the `char[]` directly; `PrintStream` re-encodes it using the caller's implicit default.

`PrintWriter` fixes 2 and 3 and matches what the parent already writes through. It doesn't fix 1 — nothing in the `Print*` family reports a failure — but it stops adding a second encoding decision. `org.xvm.tool.Console` isn't the answer either: it's line-oriented (`out(Object)`, `err(Object)`), so it can't express `suppressNewline`.

Unrelated and trivial: the Ecstasy file is spelled `ExtermalConsole.x` while it declares `ExternalConsole`. Renamed in [`045404b`](https://github.com/xtclang/xvm/commit/045404b665d501c8fc1a70fad1d78ea0344e1923) on `lagergren/lspapi-review-fixes`.

lib_runner/src/main/x/runner.x
lagergren: Or do I have to inject these kinds if things explicitly, but only in this configuration?

ggleyzer: You are 100% correct; this is on my list of TODOs. The plan to to create a dedicated "root" directory for every newly created container - very similar to the platform's HostInjector

lib_runner/src/main/x/runner.x
ggleyzer: Agree; this is extactly what I'm working on right now; expecting to push the corresponding change in a couple of hours

javatools/src/main/java/org/xvm/runtime/NativeContainer.java
ggleyzer: Prior to this work this method was private and the responsibility of ensuring uniqueness was on the main container itself. I needed to open up the access, but the contract stays the same - the main container must ensure the uniqueness; the assert is basically a documentation of that contract

lib_runner/src/main/x/runner.x
ggleyzer: I'm changing all this rn

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
ggleyzer: Yes, "a redirected run never touches CONSOLE_LOG". By design.

My assumption was that the concurrency responsibility lies on the provided PrintStream. If you think it's incorrect or too tolling for the user I'd be willing to reconsider

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: MainContainer.invokeAsync() was introduced in this project. There are just three call sites, so we don't save much. But I don't disagree in prinicipal

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: I do have a plan and I'm working on it

javatools/src/test/java/org/xvm/runtime/LspTest.java
ggleyzer: Good point. I'm a bit hesitant to include long running tests in a CI run, but would take your advice on this one. You may have noticed - this is the reproducer you originally wrote)

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: let's revisit this one after my current change is in place

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
ggleyzer: Agree; we need to document the assumptions and expectations regarding the passed in PrintStream. If you think it would be better to accept PrintWriter, I'd be glad to change that

lib_runner/src/main/x/runner.x
ggleyzer: implemented; please review

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: implemented; please review

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: please re-review

ggleyzer:
I think it's ready for a formal review now; un-drafting it

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
lagergren: A PrintStream is a mechanism that has no way of reporting an error, and it does things with flush behind the scenes for newlines and stuff. I don't think it is a great vehicle. If we do have consoles, or anything it is integrated with that we can reuse for that output, it's probably a good idea. Same in the debugger console that basically parses emitted stdout (if I understand it correctly).  I can get by with the PrintStream here but I think it is a very limiting design choice - the main use case is probably a user invisible output status - log - but there can also be failures and so on.  Leave it for now but I forsee very much that I will have to change it to something else later.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
lagergren: Exactly what I said - since it is new, let's type it properly so the Object downcasts don't grow at least. This is exactly how I would expect it to look and I would like that shape to incrementally be applied to other invokes and similar places. It also provides abstraction should we want to change something in e.g. ObjectHandle and it would turn runtime class cast exceptions today into something that simply won't compile, if we miss a case somewhere.

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
lagergren: Go with a PrintWriter for now, but I may have to change this later, depending on what I discover I can and can't do from the xtc embedding engine.

lib_runner/src/main/x/runner.x
lagergren: Confirmed, very nice and better than I expected:   `DirectoryFileStore(taskDir)` with `storage`, `rootDir`, `homeDir`, `curDir` and `tmpDir` all answered from it is exactly the shape this thread said was missing: fabricate what has to be per-run, delegate the rest.

I should correct myself on one point. I had concluded the platform had no `FileStore` that could be rooted below `/` — I read the native `xOSFileStore` with its `static final File ROOT = new File("/")` and stopped there, without looking in `lib_ecstasy/fs`, where `DirectoryFileStore` has been all along, documented for precisely this ("the Container will not be able to see 'above' the level of the injected FileStore's 'root' directory"). You used the right tool and I was wrong about it not existing.

I've taken your runner into my embedding branch as-is and `TestFiles` runs under it.

One piece of this thread is still open, and it's the injections rather than the file system: `TaskResourceProvider` extends `BasicResourceProvider`, whose `String` case forwards to the parent, so two runs asking for the same name both resolve against container zero and see one value. `LspSupport.run` still throws for `injections` too. I have that working in my calling branch on my side as a pair of parallel `String[]` arrays on `registerTask`, with a test that runs one module twice with different values — happy to put it up separately if you want it.

lib_runner/src/main/x/runner.x
lagergren: Reviewed, and this is more complete than what I had.

`TaskRegistry.unregisterTask^(id)` sits after the success/failure branch in `whenComplete`, so it runs on both paths. My attempt to figure this out only forgot a task when the run completed normally, so a failed run leaked its container and everything the container held. I have now lifted in this into the embedded branch and it seems to work nicely.

Dropping `container = Null` in the same continuation is the other half, and that turns out to matter more than I realized, though. With the container released promptly it disappears from the runtime's (weak) registry, and an ownership sweep that used to report cross-run references after a second run now comes back clean. So this closed something I had filed separately as still-open.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
lagergren: Reviewed — the poll is gone and the shape is right.

Worth saying explicitly why, because it's a subtlety I flagged as an open question and didn't want to guess at: you don't hand out the raw `outcome` future. `completion` is a *separate* `@Future` assigned inside the same `whenComplete`, after `result`/`failure`/`running` have been recorded. That's what makes it safe. Handing out `outcome` directly would have let a waiter be resumed before the task recorded its own state, and the caller would read the outcome fields before they were set — worse than the poll it replaced.

`Control.join()` on top of it is the API I wanted. To use this API, my embedding engine now does `registerTask` → `startTask` and unpacks the `(result, failure)` tuple, and the whole 40-requests-per-second-per-run machinery is deleted.

javatools/src/test/java/org/xvm/runtime/LspTest.java
lagergren: It doesn't have to be all-or-nothing, and the cheap version is nearly free.

The five scenarios already assert by throwing `IllegalStateException`, and a throwing test fails, so **annotating them `@Test` alone** buys CI coverage without rewriting a single check. If `testPoolGrows` and `testRunLatency` are the ones you're worried about, tag just those, or gate them on a property, and let the three fast ones run every commit.

The only thing genuinely tied to `main` is the two path arguments, and there's an in-tree pattern for that a few files away: `DirRepositoryConcurrentScanTest.java:44-50` resolves `xdk/build/install/xdk/lib` with a `../` fallback and then `assumeTrue(...)`, so it skips cleanly when the XDK hasn't been built rather than failing.

`testPoolGrows` is the one that needs an actual assertion rather than just the annotation — it computes `nativePoolSize(...)` on each iteration and only prints it, so whatever the intended invariant is, nothing checks it.

(And yes — noted that it started as my reproducer. All the more reason it would be nice to keep it running.)

Test caching also works well, locally, and can be added to the CI. I hope that I can use the embedding api to substantially also improve plugin performance and JVM forking time which is now one of the performance issues in the tests for cold runs with this.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,236 @@
...
+            ObjectHandle hModule     = xRTModuleTemplate.makeHandle(main, file.getModule());
+            ObjectHandle hRepository = xCoreRepository.INSTANCE.makeHandle(repository);
+            ObjectHandle hConsoleId  = consoleId == null
+                    ? xNullable.NULL
+                    : xInt64.makeHandle(consoleId);
+
+            ObjectHandle hTaskId = postRequest("registerTask", hModule, hRepository, hConsoleId).join();
+            taskId = ((JavaLong) hTaskId).getValue();
+
+            CLEANER.register(this, new TaskCleanup(connector, module.getSimpleName(), taskId));
```
lagergren: This is the one thing in the rework I'd push back on.

A task's file-system root is deleted when its `Control` becomes phantom-reachable, and `retainStore` defaults to `True` on the native path, so this is the only thing that deletes it. Four concerns, in increasing order:

1. **It's non-deterministic.** Roots accumulate until a GC happens to notice. A long-lived LSP server does a lot of runs between collections.
2. **Cleaners aren't run at JVM exit.** A process that exits normally can leave every root behind, and there's no deterministic path that removes them.
3. **It blocks a cleaner thread on the runtime.** `invokeAsync(...).join()` runs on the Cleaner's thread and calls into container zero.
4. **It can fire after the runtime is gone.** A host that closes its connector — which is what my embedding engine's `close()` does — can have a cleanup fire afterwards, calling `invokeAsync` on a shut-down runtime from a daemon thread nobody is watching. Best case a swallowed exception; worst case a `join()` that never returns.

The alternative is already in your own code: `submitTask` passes `retainStore=False` and the task deletes its own root in `whenComplete` — deterministic, in Ecstasy, no Cleaner involved. We've taken that route in our engine (a `registerTransientTask` variant) and register no cleaner at all.

Suggestion: make `Control` `AutoCloseable` and delete on `close()`, and keep the Cleaner purely as a backstop for callers who leak the Control rather than as the primary mechanism.

Separately, and I'll raise it properly once this is settled: `@Inject Directory curDir` in container zero is the *process* working directory, so runs create `{moduleName}_{id}` wherever the host happened to be started. For an editor plug-in that's the user's project directory.

lib_runner/src/main/x/runner.x
lagergren: This one didn't make it into the rework — `Boolean running;` on `Task` is still writable by anything holding the service, though everything around it changed.

`public/private Boolean running;` is the whole fix, and `status` just below already has that shape by being a computed getter.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
lagergren: Re-reviewed. The surrounding code improved a lot — `completion` replacing the poll, the failure path simplified — but this specific point is unchanged: it's still `new InterpreterControl(interpreter, ...).start()`, so `started`, `taskId` and `consoleId` still can't be `final`, and the object still exists before its values do.

Not urgent, and I'm not asking you to move the work into the constructor — that would be the wrong fix. The factory is already there and is already the only caller, so it can do the work and hand the constructor finished values. When I did that on our side, two things fell out that I hadn't predicted: `repository` stopped being a field at all (it's only read while starting), and `unregisterConsole` no longer needed to null `consoleId` to stay idempotent, because with the failure path in the factory its only caller is `finish`, which is synchronized and returns early once `running` is false.

lagergren: Concretely, since prose about a refactor is easy to disagree with. Applied to your current head (`96a1473`), compiles:

```diff
-        return new InterpreterControl(interpreter, module, repository, console, errs).start();
-    }
+        Instant started   = Instant.now();
+        Long    consoleId = console == null
+                ? null
+                : xExternalConsole.register(interpreter.getNativeContainer(), console);
+        try {
+            FileStructure file = prepareModule(interpreter, module, repository);
+            MainContainer main = interpreter.getMainContainer();
+            ...
+            ObjectHandle hTaskId = main.invokeAsync(
+                    "registerTask", hModule, hRepository, hConsoleId).join();
+            long taskId = ((JavaLong) hTaskId).getValue();
+
+            // the future exists before the task does, so the control can hold it as final
+            var completion = new CompletableFuture<ObjectHandle>();
+
+            InterpreterControl control = new InterpreterControl(interpreter, module, console, errs,
+                    started, consoleId, taskId, completion);
+
+            CLEANER.register(control, new TaskCleanup(interpreter, module.getSimpleName(), taskId));
+
+            main.invokeAsync("startTask", hTaskId).whenComplete((r, e) -> {
+                if (e == null) {
+                    ...
+                    control.finish(result, failure);
+                    completion.complete(r);
+                } else {
+                    control.finish(-1, e.toString());
+                    completion.completeExceptionally(e);
+                }
+            });
+            return control;
+        } catch (RuntimeException e) {
+            if (consoleId != null) {
+                xExternalConsole.unregister(interpreter.getNativeContainer(), consoleId);
+            }
+            throw e;
+        }
+    }
```

and the fields become:

```java
    private final InterpreterConnector connector;
    private final ModuleStructure      module;
    private final PrintStream          console;
    private final ErrorListener        errs;
    private final Instant              started;
    private final Long                 consoleId;
    private final long                 taskId;
    private final CompletableFuture<ObjectHandle> completion;

    private volatile boolean running = true;
    private volatile Instant stopped;
    private volatile Long    result;
```

Three things worth pointing at:

- **`completion` ends up final too**, which I couldn't manage when I described this in prose. Create the empty future *before* constructing the control and complete it from the `startTask` continuation, instead of assigning a future onto the object afterwards.
- **`repository` stops being a field.** It's only read while starting, so it becomes a parameter of a now-static `prepareModule` rather than state the control carries for its whole life.
- **`unregisterConsole` no longer nulls `consoleId`.** That write only existed to make it idempotent across two callers; with the failure path in the factory, the only caller is `finish`, which is synchronized and returns early once `running` is false — idempotent by construction.

I deliberately left the `CLEANER.register` call exactly where it is. That's the separate conversation and I didn't want to bundle it in here.

Whole thing is `d18f351` on `lagergren/lspapi-h12-on-new-head` if it's easier to read as a commit — it's 54 insertions, 42 deletions, one file.

lagergren:
Looks good.  I have some comments - most importantly please use PrintWriter as suggested instead of the problematic Java 1.0 PrintStream with its exception swallowing and buffer issues, and I may have to extend further depending on what I discover I need to do + some pushback on the cleaner.

Furthermore the InterpreterControl init and my with for some more finality for the fields and not unnecessarily introducing and mutable state.    The I have left open remaining simplifications and best practice comments that haven't been addressed but which would be nice, but are not life or death important.

lagergren:
I would call it "EmbeddingSupport" instead of "LspSupport". LspSupport is a very specific use case for everything that can be done with this.

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
ggleyzer: Done in [this commit](https://github.com/xtclang/xvm/pull/545/changes/c43f29eec8e9f084beb3cbe99cb88471e286f32d)

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: The reason I used that approach was that task completion is too early to delete it: the caller may still need to inspect the generated files. However, I agree that phantom reachability is not the best mechanism.

I’ve changed Control to implement AutoCloseable. It explicitly cleans up the task directory on close().

lagergren: Agreed — and removing it outright is better than the backstop I suggested.

I'd proposed keeping the Cleaner as a safety net for a leaked `Control`. That was wrong, for the reason you gave. If completion is too early to delete because the caller may still be inspecting the files, then a Cleaner firing at an arbitrary later point has exactly the same problem in a less predictable form: it deletes while someone may still be reading, just non-deterministically. A backstop that can be wrong makes the lifetime unknowable rather than merely making it the caller's responsibility, which is worse.

`AutoCloseable` puts the decision where the knowledge is — only the caller knows when it is done with the files. And all four of the original concerns go with the Cleaner: no non-determinism, nothing to miss at JVM exit, no cleaner thread blocked on the runtime, and no callback into a connector that has already shut down.

You were also right that completion is too early, which corrects the alternative I argued for. I pointed at `submitTask`'s shape — `retainStore=False`, delete in `whenComplete` — and that would have deleted output the caller had asked for. It is the right shape only when nothing can name the root afterwards (our embedding engine hands out no `Control` and exposes no task directory, so it takes that path), and it is not the right default for an API whose whole point is to hand the caller a handle.

Resolving this. One observation about `close()` itself I'll raise separately, since it is about the new code rather than this thread.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,229 @@
...
+            if (closed) {
+                return;
+            }
+            closed = true;
+        }
+
+        if (running) {
+            postRequest("killTask", xInt64.makeHandle(taskId)).join();
+        }
+        completion.join();
```
lagergren, line 215: One thing about `close()` itself, separate from the Cleaner question that thread settled.

```java
if (running) {
    postRequest("killTask", xInt64.makeHandle(taskId)).join();
}
completion.join();
postRequest("deleteTaskDirectory", ...).join();
```

`completion.join()` is unbounded. The fourth concern I raised on the Cleaner — a `join()` that never returns — hasn't gone away; it has moved from a daemon thread nobody was watching to the caller's thread, which is a real improvement (it is visible, and it is attributable) but not a fix. A task that does not stop when killed now hangs whoever called `close()`, and in a try-with-resources that is the request thread of the host.

`killTask` first makes the common case fine. It is the uncommon one that worries me: a task wedged in a native call, or one the runtime cannot interrupt, leaves `close()` with nothing to time out against. And because `closed` is set before the joins, a second `close()` returns immediately rather than retrying — correct for idempotency, but it means there is no second chance either.

Two options, in preference order:

1. **Bound the wait.** `completion.get(timeout, unit)`, and on timeout still issue `deleteTaskDirectory` and return. The root is what `close()` is really for; waiting for completion is a means to it, so a task that will not die should not also cost the caller its directory.
2. **Do not wait at all when a kill was issued.** If `killTask` has been sent, the delete can be chained off `completion` asynchronously rather than joined — `close()` returns once the intent is recorded.

Either way the caller gets a `close()` that terminates. Happy to be told this is theoretical if `killTask` is guaranteed to complete the task's future in bounded time — that is the part I cannot tell from here, and if it is guaranteed then a comment saying so on `completion.join()` would be enough.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: I added missing support for a "rootDir" that could be specified by the caller and pushed the task directory that is automatically created otherwise one level down - to ".runner" directory

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: This was a valid concern with the previous kill() behavior. I changed the runtime semantics so that Container.kill() now forcefully terminates every service in the nested container and completes exceptionally all existing fibers, including fibers waiting on futures.

As a result, killTask cannot return while any task fiber remains alive. Once postRequest("killTask", ...).join() returns, the subsequent completion.join() is only waiting for the runner-side completion callback to publish the result and unregister the task; it no longer depends on the operations in the task itself.

ggleyzer: Also, renamed LspSupport to EmbeddingSupport

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: Implemented the suggested refactoring; it allowed to remove no longer necessary synchronization

ggleyzer:
I think I implemented all the suggestions; please re-review

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,236 @@
...
+ * Interpreter-backed management and monitoring for one runner task.
+ */
+class InterpreterControl
+        implements EmbeddingSupport.Control {
+    /**
+     * Create and start the shared interpreter connector that hosts the runner module.
+     */
+    static Connector createConnector(ModuleRepository repository) {
+        InterpreterConnector connector = new InterpreterConnector(repository);
+        connector.loadModule("runner.xtclang.org");
```
lagergren, line 52: *small nitpick* Are you really sure that it wouldn't be readable with "var connector = new InterpreterConnector(repository)".   The line mentions the type twice, has a variable named "connector", which is a good variable name, and just does a new which absolutely will turn connector into an InterpretorConnector?  This is a great example of a var that DOES reduce cognitive load.

javatools/src/main/resources/errors.properties
```diff
@@ -354,3 +354,10 @@ VERIFY-90 = "{0}" incorporates "{1}", which is a virtual child of "{2}", but "{0
 VERIFY-91 = "{0}" is annotated by "{1}", which is a virtual child of "{2}", but "{0}" is not a virtual child of that class.
 VERIFY-92 = "{0}" is annotated by "{1}", which makes its method "{2}" unreachable.
 VERIFY-93 = "mixin {0} into {1}" extends "mixin {2} into {3}", but type {1} does not narrow type {3}.
+
+# LSP/runtime support errors
+LSP-1 = Unable to find application module "{0}".
+LSP-2 = Unable to find application module "{0}" version "{1}".
```
lagergren: Likely these should be called EMBEDDING-n or EBD-n or something that's not LSP.

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
```diff
@@ -0,0 +1,153 @@
...
+
+    public xExternalConsole(Container container, ClassStructure structure, boolean fInstance) {
+        super(container, structure, false);
+
+        if (fInstance) {
+            INSTANCE = this;
+        }
```
lagergren, line 38: I would give anything to remove this pattern, as in my PR, just collapsing it in a single place where we can add a single instrumentation, or debug dump printout that affects all instance pools and just something like   NativeTemplateCache.get(xExternalConsole.class) here. We now have 147 of these patterns.

javatools/src/main/resources/errors.properties
```diff
@@ -354,3 +354,10 @@ VERIFY-90 = "{0}" incorporates "{1}", which is a virtual child of "{2}", but "{0
...
+LSP-3 = Unable to create the application container: {1}. Cause: {0}
```
lagergren: Again these should likely be called EBD-x or EMBEDDED-x or something

javatools/src/main/java/org/xvm/tool/Compiler.java
```diff
@@ -567,14 +605,14 @@ protected void log(ErrorList errs) {
...
             // if there are any COMPILER errors, suppress all VERIFY errors except the first three
             boolean fSuppressVerify = false;
-            for (var err : listErrs) {
+            for (ErrorInfo err : listErrs) {
                 if (err.getCode().startsWith("COMPILER")) {
                     fSuppressVerify = true;
                     break;
                 }
             }
             int cVerify = 0;
```
lagergren, line 610: I mean, instead of clawing back to the comfort zone of types *EVERYWHERE* - I would recommend just doing: 

```
boolean fSuppressVerify = listErrs.stream()
        .anyMatch(err -> err.getCode().startsWith("COMPILER"));
```

VERY readable.

javatools/src/main/java/org/xvm/runtime/ServiceContext.java
```diff
@@ -864,6 +884,55 @@ public int shutdown(Frame frame) {
...
+    }
+
+    /**
+     * Request an asynchronous shutdown of this service.
+     *
+     * @return a future that completes when all of the service's fibers have terminated
+     */
+    public CompletableFuture<Void> requestShutdown() {
+        CompletableFuture<Void> future = m_futureShutdown = new CompletableFuture<>();
+        m_fShutdownRequested = true;
```
lagergren: CompletableFuture<Void> future = m_futureShutdown = new CompletableFuture<>();

A second call replaces it, and the first future is orphaned — nothing ever completes it, so its holder waits forever. That's the same failure class the whole path exists to remove.

Container.terminateServices() is safe today only because it memoizes, i.e. the caller is protecting the callee. Anything calling requestShutdown() on a service directly hits it. A CAS keeps the ordering that makes terminate()'s completion safe, and lets the field be final:

```
   public CompletableFuture<Void> requestShutdown() {
       CompletableFuture<Void> future = new CompletableFuture<>();
       if (!f_futureShutdown.compareAndSet(null, future)) {
           return f_futureShutdown.get();
       }
       m_fShutdownRequested = true;
       ensureScheduled(true);
       return future;
   }
```

Or you can synchronize, but this looks more like the rest of the CAS-locking and is nicer IMHO.

lib_runner/src/main/x/runner.x
```diff
@@ -0,0 +1,404 @@
...
+            return id;
+        }
+
+        /**
+         * Implementation of the `startTask` API. It's marked as `@Concurrent` to allow [killTask()]
+         * to enter the registry service while waiting for the task's completion.
+         */
+        @Concurrent
+        Tuple<Int, String> startTask(Int id) = taskFor(id).start();
+
```
lagergren: Nice. It is worth recording that this one is backed by an independent measurement. A JFR pass over the manualTests suite hit exactly the defect it fixes: TaskRegistry is a singleton service and startTask is a blocking call, so the registry is held for a whole run — one hung module stuck through every later run through the same engine, and killTask couldn't get in to stop the one that. @Concurrent is the right fix, and it's now load-bearing rather than defensive.

I highly recommend running JFR for shaking out concurrency issues. Also because there are some low hanging fruits.

javatools/src/main/java/org/xvm/api/InterpreterControl.java
```diff
@@ -0,0 +1,236 @@
...
+            postRequest(connector, "killTask", xInt64.makeHandle(taskId)).join();
+        }
+        completion.join();
+
+        if (rootDir == null) {
+            postRequest(connector, "deleteTaskDirectory", xInt64.makeHandle(taskId),
+                    xString.makeHandle(module.getSimpleName())).join();
+        }
+    }
+
```
lagergren, line 222: completion.join() now looks good - we drain the message queue thru ServiceContext.terminate(), and forces R_EXCEPTION on fibers, so a killed task genuinely completes now and completion.join()r returns. Very good. 

However you still have three unbounded closing joins: 

1.  killTask (:213)
2. completion (:215)
3. deleteTaskDirectory (:219) 

A fiber that can't be preempted likely still hangs the caller, and in try-with-resources that is the host's request thread. A timeout on the middle one would make the failure attributable instead of terminal, which would be preferred, right?

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: Yes, it would be readable. It's just my personal preference as a reader :)

javatools/src/main/resources/errors.properties
ggleyzer: Yes, my bad. Changing now

javatools/src/main/java/org/xvm/runtime/template/_native/io/xExternalConsole.java
ggleyzer: it's only used for native templates and only used during native container loading. I still have not seen any problem that has caused (within the assumptions of a single runtime instance). But if it bothers you much I will look into your PR again

lagergren:
Getting really good now. Summary of my latest comments: 

### Done — resolving

| thread | what settled it |
| --- | --- |
| `InterpreterControl` construct-then-start | 8 fields now `final`, the mutable trio `volatile`, and `synchronized` removed because it became unnecessary rather than merely deleted |
| `runner.x` — `BasicResourceProvider` file system | `DirectoryFileStore(taskDir)` with `storage`, `rootDir` and `homeDir` all served |
| `InterpreterControl` — polling | the poll is gone and the shape is right |

The locking one is worth calling out as the model for the rest: the fields became `final` because
the object is now constructed only once everything it needs is known, and dropping `finish()`'s
`if (!running) return;` guard is correct — it is now reached from exactly one `whenComplete`, and
the idempotence that guard was providing moved to where it belongs, `close()` gating on
`closed.compareAndSet(false, true)`.

### Substantially done, one thing left

**`close()`'s unbounded join.** `b1f1179da` addresses the *cause*: `terminateServices()` calls
`requestShutdown()` per service, and `terminate()` drains the message queue completing futures
exceptionally and forces `R_EXCEPTION` on fibers, so a killed task genuinely completes and
`completion.join()` returns. That is great!

What remains is the **bound**. `close()` still has three unbounded joins — `killTask` (`:213`),
`completion` (`:215`), `deleteTaskDirectory` (`:219`). A fiber that cannot be preempted still hangs
the caller, which in a try-with-resources is the host's request thread. A timeout on `:215` would
make that failure attributable rather than terminal. We can do this after merge too, as an add-on
fix, but it still seems unsafe here.

### New finding — `requestShutdown()` is not idempotent

```java
CompletableFuture<Void> future = m_futureShutdown = new CompletableFuture<>();
```

(which of course also could be written as the much more readable *nitpick*):

```java
var future = m_futureShutdown = new CompletableFuture<Void>();
```

A second call to `requestShutdown` replaces the field and **orphans the first future** — nothing ever completes it, so
its holder waits forever. That is the same failure class this whole path exists to remove.
`Container.terminateServices()` is safe today only because it memoizes, which is the caller
protecting the callee; anything calling `requestShutdown()` on a service directly hits it.

A compare-and-set preserves the ordering that makes `terminate()`'s completion safe — the CAS
publishes the future, then the volatile `m_fShutdownRequested` write, so a run-loop thread that
observes the flag is guaranteed to see it — and lets the field become `final`:

```java
public CompletableFuture<Void> requestShutdown() {
    CompletableFuture<Void> future = new CompletableFuture<>();
    if (!f_futureShutdown.compareAndSet(null, future)) {
        return f_futureShutdown.get();   // already requested; join the first sweep
    }
    m_fShutdownRequested = true;
    ensureScheduled(true);
    return future;
}
```

The completion site changes with it: `f_futureShutdown.get().complete(null)` in `terminate()`.

### Nice bonus, cause of my JFR crashes: `@Concurrent` on `startTask`

This one is backed by measurement rather than review. A JFR pass over the manualTests suite hit
exactly the defect it fixes: `TaskRegistry` is a singleton `service` and `startTask` is a blocking
call, so the registry is held for the whole duration of a run — one hung module wedged **every**
later run through the same engine, and `killTask` could not get in to stop the one that was stuck.
`@Concurrent` is the right fix.

### Still open

- `EmbeddingSupport` — `configured`, `cfgRepo`, `cfgInjector` and `connector` (`:83`–`:86`) are
  still non-final.
- `configured` is still non-volatile, read unlocked at `:93` and locked at `:103` — double-checked
  locking without `volatile`, which is a real race rather than a style point.
- The holder field at `:66` is still not `static final`. We need the JMM holder pattern for this to be
  a safe and proper singleton.
- `runner.x` — `Boolean running` on `Task` is still writable by anything holding the task instead of
public/private

javatools/src/main/resources/errors.properties
ggleyzer: done

javatools/src/main/java/org/xvm/tool/Compiler.java
ggleyzer: agree; changing

javatools/src/main/java/org/xvm/runtime/ServiceContext.java
ggleyzer: agree (even though it's a completely theoretical possibility); since it's a very cold path I'll use synchronization

javatools/src/main/java/org/xvm/api/InterpreterControl.java
ggleyzer: With the new shutdown semantics, close() should return almost immediately. Natural Ecstasy execution and suspended fibers are now terminated exceptionally, so neither should delay completion. A timeout would serve only as protection against a blocked native call or unknown runtime failure, which I think is a bit of an overkill.

ggleyzer:
Fields cfgRepo, cfgInjector and connector are intentionally mutable: configuration is installed once after construction, and connector is initialized lazily. They cannot be final without redesigning the lifecycle.

All other comments accepted and applied

lagergren:
Approved. This is definitely enough for me to take it from here now, merge whenever you feel like it.

PR #546, opened 2026-08-28
DRAFT POC: warm in-process XtcEngine on master (compile + run, no JVM fork)

lagergren:
**Draft POC.** Proves a warm, in-process compile+run engine works on `master`, so tooling (the Gradle plugin, an LSP server, a test runner) can stop forking a JVM and re-bootstrapping the runtime per invocation.

## Why this matters for the LSP: it fills the one stub

The LSP is **already built** in `lang/` — `lsp-server` (full lsp4j plumbing, `publishDiagnostics`, document services), `dap-server`, the IntelliJ plugin, the tree-sitter grammar. Language intelligence is served through a pluggable `Adapter`:

| Adapter | Backend | Status |
|---|---|---|
| `MockAdapter` | regex | testing/fallback |
| `TreeSitterAdapter` | tree-sitter | **works today** — "syntax-aware (~80% LSP features)" |
| `XdkAdapter` | **XDK compiler** | **STUB** — *"all methods log but return empty results"* |

So the editor, protocol and syntactic halves are done. The hole is the **compiler-backed adapter**: something that actually compiles a buffer into real diagnostics and can run a module. That is exactly what this supplies — `compile(...)` over in-memory buffers producing structured `Diagnostic`s (severity, **code**, message, source, line) streamed to the caller's `ErrorListener`, and `start(...)` returning a `RunControl`.

**`lang` owns the LSP; this owns compile-and-run.** Together they close the `XdkAdapter` stub. What remains beyond that is genuinely semantic — hover, type inference, cross-file navigation — plus the listed gaps. The `compile(Path...)` entry point for on-disk source trees - the thing an LSP workspace actually has - is now **implemented** here.

Deliberately minimal: **5 files touched + 1 new**, ~1050 lines including tests and notes. None of the deeper hardening work is lifted in.

## Naming the contract: `ToolApi`

The embedding contract was implicit in one class. `ToolApi` makes it explicit — `XtcEngine` implements it, and it is deliberately the shape `ToolConnector` targets, so the correspondence between the two is obvious rather than something a reader reconstructs. It states what the contract guarantees: the runtime is booted **once** and reused; diagnostics are never silently lost; runs are **observable** (a `RunControl`, not a bare future); and nothing handed in is aliased. `RunControl` expresses absence as `Optional`, so "still running", "failed" and "returned nothing" cannot be confused.

## What is proven — 11 green tests (`org.xvm.api.XtcEngineTest`)

1. **Compile Ecstasy source held in a `String`** — no files, no CLI.
2. **Compile from ON-DISK source** (`compile(Path...)`) — a `.x` file or a directory tree, which is what an LSP workspace and a build tool actually have. It runs identically to the in-memory path.
3. **Run the compiled module in a nested container in the same JVM**, awaiting event-driven completion.
4. **The module's `run()` return value reaches the host** — exact `Int` values, and a `void run()` reports no result rather than looking like a failure.
5. **A throwing run completes exceptionally** — the failure path: a run that blows up must not look like one that succeeded.
6. **One warm engine serves repeated compile+run cycles** — the runtime bootstrap is paid once, not per invocation.
7. **Several modules compile in one request**, and all of them run.
8. **Compiled modules sync to disk** (`writeTo`) and reload through a plain `DirRepository`.
9. Diagnostics **stream to the caller's own `ErrorListener`** as they are produced, and a failed compile is self-describing (the result always carries its diagnostics).
10. **A path that is not a module is a diagnostic, not a crash** — an LSP must not die because someone opened the wrong directory.
11. **The engine is usable through the `ToolApi` interface alone.**

Gate: `./gradlew xdk:installDist`, then `./gradlew :javatools:test :javatools_utils:test` — both green. Run them as separate invocations; in one invocation the test's module reads race `installDist`'s writes.

## The changes

| File | Change |
|---|---|
| `MainContainer` | capture the `callLater` future that was being **discarded**; expose `futureResult()` |
| `Container` | add `runModule(String, ObjectHandle...)` returning that completion future |
| `NestedContainer` | add `createForHost(...)`; add a parent injection fallback **only** for host containers (`f_hProvider == null`) — without it a host-run module gets no `Console` and dies with "Invalid resource". Guest containers keep the strict sandbox. |
| `FileStructure` | **bug fix**, see below |
| `DirRepository` | **bug fix** — synchronize the scan cache (also submitted standalone as #547) |
| `api/ToolApi.java` | new — the named CONTRACT a tool embeds; `XtcEngine` implements it, and it is the shape `ToolConnector` targets too |
| `api/XtcEngine.java` | new — the engine: compile pipeline (in-memory **and** on-disk), run, `RunControl`, JFR events |

## Error handling: no frivolous unchecked exceptions

Two deliberate choices, since this is a public API:

- `compile(...)` **propagates the `IOException` that `ModuleRepository.storeModule` declares**, rather than wrapping it in an `IllegalStateException`. A caller that can handle a failed write is given the chance to; one that cannot is not surprised by a `RuntimeException`. It is declared on the `ToolApi` contract too.
- A path that is not a module is ordinary **user** error and becomes a **diagnostic**. The catch is narrow — `IllegalArgumentException`, which is how `ModuleInfo` signals bad input — so it cannot swallow arbitrary `RuntimeException`s and hide real defects. Every failure path logs: an early cut returned null silently and produced "it failed" with zero diagnostics, violating this API's own guarantee that failures are self-describing. The test caught it.

## A master bug this found (deterministic, with a red proof)

`FileStructure.getErrorListener()` read the ambient `ConstantPool.getCurrentPool()` and dereferenced it **with no null check**. That thread-local is null on any thread with no pool bound — i.e. **every Java host thread driving the runtime** — so the embedding path NPE'd inside a *diagnostic accessor*:

```
NullPointerException: Cannot invoke "ConstantPool.getErrorListener()" because "poolCurrent" is null
  at FileStructure.getErrorListener(FileStructure.java:1397)
  at TypeConstant.ensureTypeInfo(TypeConstant.java:1659)
  at Container.findModuleMethod(Container.java:252)
```

Null-guarded here (one line) so the POC can run. **Now submitted standalone as #548**, where the repro turned out to be two lines (construct a `FileStructure`, ask it for its `ErrorListener`) — no engine required. That PR also sets out why expressing pool *ownership* as a thread-local rather than a parameter produces this family of bug repeatedly.

## Known gaps (tracked in `POC-NOTES.md`, not hidden)

1. ~~A module's `run()` return value does not reach the host.~~ **FIXED** — two separate defects, now covered by tests: `ServiceContext.callLater` hardcodes `cReturns = 0`, so the future was told to expect nothing and completed with an empty tuple (`runModule` now calls `postRequest(..., fReturn ? 1 : 0)`); and the native instantiate-and-run op ignored the caller-designated return slot in favour of `A_STACK`, so even once a return was requested the value landed where the service entry frame never reads (it completes from `f_ahVar[0]`).
2. `RunControl.kill()` cancels the caller's wait and releases the container, but does not force-unwind a running fiber.
3. `NativeFunctionHandle` binds to `xRTFunction.INSTANCE.f_container` (a process-static template instance) — fine for a single native plane.
4. ~~No `compile(Path...)` source-tree entry point.~~ **DONE** — `ModuleInfo` walks a module file or directory into the same `TypeCompositionStatement` the in-memory path produces (exactly what the CLI compiler does), so both entry points now share **one** pipeline and the load-bearing stage order lives in a single place.
5. Diagnostics carry a line number but no column/end position; LSP wants a `Range` to underline.
6. No incremental or cancellable compile — an LSP recompiles per keystroke, and today each call is a whole-module compile.
7. No pool-publication fence — sequential work is fine; parallel work needs review.

## What this unlocks, and what each mode still needs

| Mode | Works today? | Still needs |
|---|---|---|
| **Sequential compile** | ✅ proven | in-memory **and** from disk; nothing outstanding |
| **Sequential run** | ✅ proven | results and failures both surface |
| **Parallel compile** | ⚠️ unblocked here | `DirRepository`'s scan-cache race is **fixed** in this branch (and standalone in #547); the other known blocker, the static compiler counters, is **already fixed on master** by #538. Both closed, so this is now a matter of testing it under load. |
| **Parallel run** | ❌ | the shared-pool publication/interning work; this is the deep one |

### DIFF
--- a/POC-NOTES.md
+++ b/POC-NOTES.md
@@ -34,7 +34,7 @@ navigation) — the compiler-API surface the adapter would query — plus the sm
 
 ---
 
-## What is proven — 9 green tests (`org.xvm.api.XtcEngineTest`)
+## What is proven — 11 green tests (`org.xvm.api.XtcEngineTest`)
 
 | Test | Proves |
 |---|---|
@@ -47,6 +47,8 @@ navigation) — the compiler-API surface the adapter would query — plus the sm
 | `compiledModulesCanBeSyncedToDiskAndReloaded` | `writeTo(dir)` → reload via a plain `DirRepository` |
 | `reportsCompileDiagnosticsAndStreamsThemToTheCaller` | diagnostics stream to the caller's sink |
 | `theEngineIsUsableThroughTheToolApiContractAlone` | the `ToolApi` contract is sufficient by itself |
+| `compilesAModuleFromDisk` | **`compile(Path...)`** - on-disk source compiles and runs identically |
+| `aPathThatIsNotAModuleIsADiagnosticNotACrash` | a bad path is a diagnostic, not an exception |
 
 Gate: `./gradlew xdk:installDist`, then `./gradlew :javatools:test :javatools_utils:test` — green.
 Run them as **separate invocations**; in one invocation the test's module reads race
@@ -101,10 +103,13 @@ sealing, display purity. None is needed here.
 
 ## Known gaps
 
-1. **No `compile(Path...)` source-tree entry point** — only in-memory sources. An LSP workspace and
-   the Gradle plugin both want on-disk module directory trees. `org.xvm.tool.ModuleInfo` already
-   walks a module directory into a parsed node tree, so this is wiring, not new compiler work.
-   **Biggest remaining gap; low difficulty.**
+1. ~~No `compile(Path...)` source-tree entry point.~~ **DONE.** `ModuleInfo` walks a module file or
+   directory into the same `TypeCompositionStatement` the in-memory path produces, so both entry
+   points now share ONE pipeline - the load-bearing stage order exists in a single place. A path that
+   is not a module produces a diagnostic rather than an exception (`ModuleInfo` signals bad input by
+   throwing `IllegalArgumentException`; an LSP must not die because someone opened the wrong
+   directory). `compile(...)` propagates the `IOException` that `ModuleRepository.storeModule`
+   declares, rather than wrapping it in an unchecked exception.
 2. **Diagnostics carry `line` but no column/end position.** LSP wants a `Range` to underline. Needs
    checking whether `ErrorInfo` carries start/end offsets.
 3. **No incremental or cancellable compile.** An LSP recompiles per keystroke; today each call is a
--- a/POC-NOTES.md
+++ b/POC-NOTES.md
@@ -1,85 +1,117 @@
-# XtcEngine POC on master — status and known gaps
+# XtcEngine POC on master — status, and how it relates to the LSP
 
-Branch: `lagergren/xtc-engine-poc`, based on `origin/master` `82683bcd2`.
-Goal: prove a warm, in-process compile+run engine works on master, so a Gradle plugin (and an
-LSP server) can stop forking a JVM per invocation.
+Branch: `lagergren/xtc-engine-poc` (PR #546), based on `master`.
 
-## What is proven (3 green tests, `org.xvm.api.XtcEngineTest`)
+**Goal:** prove a warm, in-process compile+run engine works on `master`, so tooling stops forking a
+JVM and re-bootstrapping the runtime for every operation.
 
-1. **Compile Ecstasy source held in a String** — no files, no CLI.
-2. **Run the compiled module in a nested container**, in the same JVM, and await event-driven
-   completion.
-3. **One warm engine serves repeated compile+run cycles** — the runtime bootstrap is paid once.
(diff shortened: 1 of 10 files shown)


lagergren:
There is a rather large number of thread unsafe parallel code that triggers if you try anything but compiling and running in sequence, partially probably from my side, but in rather many non volatile mutable fields in XTC as well + constant pool leaks for repeated compiles with no eviction.

lagergren:
Have enough info after doing the review and a few adjustments to LSPAPI to continue without the XtcEngine raw sketch.

PR #547, opened 2026-08-28
Fix DirRepository scan-cache data race (blocks parallel compilation)

lagergren:
Fixes a data race in `DirRepository`'s directory-scan cache. Two files: the fix, and a test that is **red on `master` without it**.

## The bug

The scan cache is a plain `HashMap` (`modulesByFile` — a non-final field reassigned wholesale by `rebuildCache`) and a plain `TreeMap` (`modulesByName` — rebuilt with `clear()` followed by a `put()` loop). None of `loadModule`, `getModuleNames`, `ensureCache` or `rebuildCache` is synchronized.

A fresh repository has `lastScan == 0`, so `isCacheValid()` returns false and the **first** concurrent access has *every* thread run `rebuildCache` over the same maps simultaneously. On top of that, `getModuleNames()` returns `Collections.unmodifiableSet(modulesByName.keySet())` — a **live, fail-fast view** of the map being rebuilt.

Two parallel compiles resolving library modules from one repository hit this immediately:

```
java.util.ConcurrentModificationException
  at DirRepository.loadModule(DirRepository.java:91)
  at LinkedRepository.loadModule(LinkedRepository.java:101)
  at FileStructure.linkModules(FileStructure.java:825)
  at NativeContainer.loadNativeTemplates(NativeContainer.java:158)
```

and in principle a spinning corrupted `HashMap` or silently dropped entries ("module not found").

## The fix

Synchronize the three public entry points on the repository.

A coarse per-repository lock is the **right** granularity here rather than a pessimisation: this is a directory-scan cache consulted once per module lookup, not a hot inner loop.

Worth stating explicitly, because it looks like a place to reach for concurrent collections — **that would not be sufficient**. `ModuleInfo.ensureModule()` *also* lazily deserializes:

```java
if (module == null || module.isModified()) {
    module = tryLoad();
}
```

so with concurrent maps (or a copy-on-write volatile swap) two threads would still duplicate that deserialization and could publish a half-built module. The lock covers both the cache rebuild and the lazy materialization reached through it.

## The proof

`DirRepositoryConcurrentScanTest` — 8 threads × 80 iterations over a fresh repository. Verified **in this branch**:

- **without** the fix: fails with `ConcurrentModificationException`
- **with** the fix: passes

One note on the test: it feeds the repository modules the running JVM can actually parse (the freshly built XDK), because a module from a different `.xtc` format version fails to parse and never enters the cache — which would make the test vacuously green.

## Why now

This is the **last remaining blocker for parallel compilation**. The other blocker previously identified — the static compiler counters in `ConditionalStatement`, `ElseExpression`, `ElvisExpression` and `MethodDeclarationStatement` — is already fixed on master by #538 (all four are `AtomicInteger`). With both closed, two compiles of different modules are isolated: each clones library modules into its own build repository, so interning happens per-compile.

Gate: `./gradlew xdk:installDist`, then `./gradlew :javatools:test :javatools_utils:test` — green. (Separate invocations; in one invocation the test's module reads race `installDist`'s writes.)

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/DirRepository.java
+++ b/javatools/src/main/java/org/xvm/asm/DirRepository.java
@@ -25,6 +25,22 @@ import java.util.TreeMap;
 /**
  * A simple ModuleRepository that manages its contents in a directory.
  */
+/*
+ * THREAD SAFETY: the three public entry points below are synchronized on the repository.
+ *
+ * The scan cache is a plain HashMap (modulesByFile, a non-final field reassigned wholesale) and a
+ * plain TreeMap (modulesByName, rebuilt with clear() + a put() loop). Concurrent callers - two
+ * parallel compiles resolving library modules, say - otherwise race those rebuilds against get()
+ * and against the live keySet() view getModuleNames() hands out, which throws
+ * ConcurrentModificationException (and can spin or silently drop entries).
+ *
+ * A coarse per-repository lock is the right granularity here, not a pessimisation: this is a
+ * directory-scan cache consulted once per module lookup, not a hot inner loop. Note that swapping
+ * in concurrent maps, or copy-on-write with a volatile swap, would NOT be sufficient on its own,
+ * because ModuleInfo.ensureModule() ALSO lazily deserializes ("if (module == null) module =
+ * tryLoad()") - two threads would still duplicate that work and could publish a half-built module.
+ * The lock covers both the cache rebuild and the lazy materialization reached through it.
+ */
 public class DirRepository
         implements ModuleRepository {
     // ----- constructors  -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


cpurdy (approved):
I think we still need to give DirRepository a good once-over but this is a reasonable layer of protection in the meantime.

javatools/src/main/java/org/xvm/asm/DirRepository.java
```diff
@@ -25,6 +25,22 @@
 /**
  * A simple ModuleRepository that manages its contents in a directory.
  */
+/*
+ * THREAD SAFETY: the three public entry points below are synchronized on the repository.
+ *
+ * The scan cache is a plain HashMap (modulesByFile, a non-final field reassigned wholesale) and a
```
ggleyzer, line 31: I think the comment is a bit excessive

PR #550, opened 2026-08-28
Fix Version.isSameAs() indexing the wrong array (oneliner)

lagergren:
One-liner: `Version.isSameAs()` indexes `thatInts[i]` where it means `remaining[i]`.

```java
int[] remaining = cThis > cThat ? thisInts : thatInts;
for (int i = cShared, c = remaining.length; i < c; ++i) {
    if (thatInts[i] != 0) {          // <-- remaining[i]
```

`remaining` is computed and then never used. The loop bound comes from the longer
array while the indexed array is always `thatInts`, so when the receiver is the
longer of the two the read runs off the end.

The method compares the shared version-part prefix and then requires every part
beyond it to be zero, so `1.2.0` and `1.2` are meant to be the same version. Two
observed consequences:

```java
new Version("1.2.3").isSameAs(new Version("1.2"))
    // ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
    // (expected: false)

new Version("1.2.0").isSameAs(new Version("1.2"))
    // ArrayIndexOutOfBoundsException
    // (expected: true)
```

So the trailing-zeros semantic the method exists to provide never works in that
direction. The reversed direction works only by accident, because there
`remaining` and `thatInts` are the same array — which is why this survived.

**Fix:** index `remaining[i]`. No signature, semantic, or performance change; the
already-correct direction is untouched.

**Test:** `VersionTest.testIsSameAsAcrossDifferingPartCounts` covers both
directions, multiple trailing zeros, a non-zero part past the shared prefix, and
equal-length pairs as a control. Verified red on master before the fix
(`ArrayIndexOutOfBoundsException` at `Version.java:492`) and green after.

### DIFF
--- a/javatools/src/test/java/org/xvm/asm/VersionTest.java
+++ b/javatools/src/test/java/org/xvm/asm/VersionTest.java
@@ -309,6 +309,40 @@ public class VersionTest {
         assertThrows(IllegalStateException.class, () -> new Version(new int[] {1, -7}, null));
     }
 
+    /**
+     * {@code isSameAs} compares the shared version parts and then requires every part beyond the
+     * shared prefix to be zero, so "1.2.0" and "1.2" are the same version.
+     *
+     * <p>The remainder loop selected the longer array into {@code remaining}, iterated to ITS
+     * length, and then indexed {@code thatInts} anyway. Whenever the receiver was the longer of
+     * the two, that read ran past the end of {@code thatInts}: "1.2.3".isSameAs("1.2") threw
+     * {@link ArrayIndexOutOfBoundsException} rather than returning false, and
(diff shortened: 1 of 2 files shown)


cpurdy (approved):
Well, that was a painfully obvious bug. 🤦‍♂️

javatools/src/test/java/org/xvm/asm/VersionTest.java
ggleyzer, line 1: I don't believe this deserves a test

lagergren: Everything deserves a test. There were ZERO tests checking the equivalence between "v 1.2" and "v 1.2.0", so if there is one thing that should have had a test, and didn't have that test, which in turn led to this being undiscovered, this is it. I really don't follow this logic?

PR #551, opened 2026-08-28
Fix JsonMapStore Small→Medium transition resurrecting deleted keys

mrcompator:
## What was wrong

A `JsonMapStore` holds its data in memory in one of three "storage models" — Small, Medium, Large — and grows between them as the map gets bigger. When it grows **Small → Medium**, it moves the real values out of memory and leaves an `OffHeap` marker in the history in their place.

Both places that perform this conversion — `commit()` (for the batch that triggered the growth) and `retainTx()` (for the rest of the history, during maintenance) — rewrote **every** history entry to `OffHeap` without checking whether the entry was a `Deleted` tombstone.

Because `OffHeap != Deleted`, this made every key that had ever been `remove()`d look present again:

- `keysAt()` walks the history, so it counted those resurrected keys;
- `sizeAt()` reads a separate size index (`sizeByTx`), so it correctly did **not**;
- `keysAt()` asserts the two agree → it threw `"keys.size == size"`.

Because the in-memory history stays corrupted, every later read of that map fails too — the error is sticky and cascades.

Observed in the field as `"keys.size == size": keys.size=296, size=284` on a busy DBMap during a long accumulated test run. The durable transaction log and `sizeByTx` both agreed at 284; the in-memory history reported 296 — the 12-key gap was exactly the number of deleted keys resurrected by a Small→Medium transition.

## The fix

Skip `Deleted` entries when converting history values to `OffHeap`, in both `commit()` and `retainTx()`. Two small guards.

## Reproducer

`JsonMapStoreTest.shouldKeepSizeAndKeysConsistentUnderConcurrentTransactions` — **deterministic**, ~1m40s:

- `@TestInjectables` drops the model thresholds so the store crosses to Medium partway through the run;
- several `Client`s drive overlapping insert / delete / rollback transactions against one `mapData` store;
- a reader repeatedly checks, inside one read transaction, that `mapData.size` equals the iterated key count (`keysAt()`'s own internal assertion also surfaces here).

Fails as `keys.size=520, size=412` without the fix; passes with it. Full `jsondb_test` stays green (55 / 55).

### DIFF
--- a/manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
+++ b/manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
@@ -523,134 +523,73 @@ class JsonMapStoreTest {
     }
 
     /**
-     * Regression test for a `keys.size == size` desync in `JsonMapStore` -- distinct from the
-     * (fixed) `keysAt()` merge-ordering bug above.
-     *
-     * When the store grew from the Small to the Medium model, both the `commit()` and the
-     * `retainTx()` transition paths rewrote history entries to the `OffHeap` marker
-     * *unconditionally* -- including `Deleted` tombstones. Since `OffHeap != Deleted`, every
-     * previously-removed key became visible to `keysAt()` again, while `sizeAt()` (which
-     * tracks `sizeByTx`) still correctly excluded it, so `keysAt()`'s `assert keys.size ==
-     * size` failed and the store stayed desynced (`keysAt() > sizeAt()` by the number of
-     * resurrected tombstones).
-     *
-     * Seen in the field on the hottest DBMap as `"keys.size == size": keys.size=296,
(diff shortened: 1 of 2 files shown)


manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
```diff
@@ -521,4 +521,136 @@ class JsonMapStoreTest {
             }
         }
     }
+
+    /**
+     * Regression test for a `keys.size == size` desync in `JsonMapStore` -- distinct from the
```
ggleyzer: I would prefer to keep the doc for the regression test short, for example:

    Regression test for transition from Small to Medium model.

The rest of the explanation should be moved to the PR description (if it's not there already)

ggleyzer (approved):

PR #557, opened 2026-08-29
Fix IntersectionTypeConstant.mergeChildren guarding the wrong argument

lagergren:
`mergeChildren` null-checks `info1` twice and then dereferences `info2`:

```java
map1 = info1 == null ? ListMap.EMPTY : info1.getChildInfosByName();
map2 = info1 == null ? ListMap.EMPTY : info2.getChildInfosByName();   // info1, reads info2
```

**Both `TypeInfo`s are independently nullable**, and the caller says so itself — `RelationalTypeConstant.mergeTypeInfo` computes its `Progress` as `info1 == null || info2 == null ? Incomplete : …`.

So with `info1` present and `info2` absent — one half of an intersection resolved, the other not yet — the guard passes and `info2.getChildInfosByName()` throws `NullPointerException` while building a `TypeInfo`, far from anything that names the cause.

**Fix:** guard `map2` on `info2`.

**Scope check:** the whole `merge*` family was audited for the same copy-paste shape. This is the only instance, so it is a fix rather than a pattern.

**Test:** `IntersectionChildMergeTest` — the both-absent case as a control (it always worked, because the wrong guard is accidentally right when the two arguments agree) plus a pin on the guard itself. Verified against unfixed master.

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/constants/IntersectionTypeConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/IntersectionTypeConstant.java
@@ -574,7 +574,7 @@ public class IntersectionTypeConstant
     @Override
     protected ListMap<String, ChildInfo> mergeChildren(TypeInfo info1, TypeInfo info2, ErrorListener errs) {
         ListMap<String, ChildInfo> map1 = info1 == null ? ListMap.EMPTY : info1.getChildInfosByName();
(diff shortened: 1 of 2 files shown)


ggleyzer:
I don't think there is a reason to clutter the repo with such a trivial typo fix. I pushed it as a minimal change.

PR #558, opened 2026-08-29
Fix AstNode.fieldsForNames: dead type guard and null-hole returns

lagergren:
Two defects in the method that builds the reflective AST child-field model (65 call sites). It is supposed to fail loudly on a bad field name or an unsupported field type. **It did neither.**

### The type guard never fired

```java
if (!field.getType().isInstance(AstNode.class) && field.getType().isInstance(List.class)) {
```

`Class.isInstance(x)` asks whether the **object** `x` is an instance of the receiver. Passing `AstNode.class` asks whether the *`Class` object* is an instance of the field's type — false for every realistic field type. Both conjuncts were false, so any field type was silently accepted. Observed:

```
AstNode.class.isInstance(AstNode.class)       = false
AstNode.class.isAssignableFrom(AstNode.class) = true     <- what was meant
```

### The not-found path tested the wrong variable

```java
clzTry = clzTry.getSuperclass();
if (clz == null) {                       // clz is the PARAMETER, never null here
    throw new IllegalStateException(eOrig);
}
```

The intended variable was `clzTry`. When a named field exists nowhere in the hierarchy, `clzTry` walks to null, the loop exits, and `fields[i]` is left **null** — discarding the saved `NoSuchFieldException` that named the missing field. A renamed or deleted child field therefore surfaces later as an opaque failure where the null slot is dereferenced, instead of immediately as *"no such field &lt;name&gt;"*. Observed:

```
fieldsForNames(AstNode.class, "noSuchFieldAnywhere") -> fields[0]=null
```

### The obvious risk was measured, not assumed

Correcting a guard that never fired can start rejecting things it should have been rejecting all along. **That surfaced set is empty:** force-loading all 183 classes in `org.xvm.compiler.ast` — which runs the static initialisers where `fieldsForNames` is called — gives `loaded=183, guard violations=0`, on master.

**Test:** `AstNodeFieldModelTest` — a missing field must throw and name itself, an unsupported field type must be rejected (both red before), plus the two shapes the child model *is* made of (a direct `AstNode` field and a `List` field) and an inherited field, which are green before and after and pin what the fix must not break.

### DIFF
--- a/javatools/src/main/java/org/xvm/compiler/ast/AstNode.java
+++ b/javatools/src/main/java/org/xvm/compiler/ast/AstNode.java
@@ -1886,8 +1886,6 @@ public abstract class AstNode
                 try {
                     Field field = clzTry.getDeclaredField(names[i]);
                     assert field != null;
-                    // isAssignableFrom, not isInstance: isInstance(X.class) asks whether the
-                    // CLASS OBJECT is an instance, which is not the question here
                     Class<?> clzField = field.getType();
                     if (!AstNode.class.isAssignableFrom(clzField)
                             && !List.class.isAssignableFrom(clzField)) {
@@ -1902,7 +1900,6 @@ public abstract class AstNode
                         eOrig = e;
                     }
 
-                    // clzTry, not clz: clz is the parameter and is never null here
                     clzTry = clzTry.getSuperclass();
                     if (clzTry == null) {
                         throw new IllegalStateException(eOrig);
--- a/javatools/src/main/java/org/xvm/compiler/ast/AstNode.java
+++ b/javatools/src/main/java/org/xvm/compiler/ast/AstNode.java
@@ -1873,22 +1873,26 @@ public abstract class AstNode
      *
      * @return an array of fields corresponding to the specified names on the specified class
      */
-    protected static Field[] fieldsForNames(Class clz, String... names) {
+    protected static Field[] fieldsForNames(Class<?> clz, String... names) {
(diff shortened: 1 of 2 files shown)


lagergren:
Also this wouldn't even have compiled if you used the @NotNull annotation framework.

javatools/src/main/java/org/xvm/compiler/ast/AstNode.java
```diff
@@ -1873,22 +1873,26 @@ protected static <T> ArrayList<T> ensureArrayList(List<T> list) {
...
         NextField: for (int i = 0, c = fields.length; i < c; ++i) {
-            Class                clzTry = clz;
+            Class<?>             clzTry = clz;
             NoSuchFieldException eOrig  = null;
             while (clzTry != null) {
                 try {
                     Field field = clzTry.getDeclaredField(names[i]);
                     assert field != null;
-                    if (!field.getType().isInstance(AstNode.class) && field.getType().isInstance(List.class)) {
+                    // isAssignableFrom, not isInstance: isInstance(X.class) asks whether the
```
ggleyzer: While this comment is useful in this diff, it's not helpful, or even confusing in the resulting code. For the reader of the fixed version the history of the changes is invisible, and making the point "isAssignableFrom, not isInstance" doesn't do much good.

If someone wishes to see the history of changes, the PR itself would be a wonderful source of the information

ggleyzer (approved):

PR #559, opened 2026-08-29
Support the TimeZone: literal end to end, drop the RegEx arm that could only throw, and pin the two format lists together

lagergren:
This broke for me when fuzzing types, and please correct me if I am wrong, but to me it seems like a genuine outage / not implemented situation. Let me know if I have misunderstood anything. I did the fixes described below, and use that for the LSP compiler / runner code and it seems to work as intended for me, but please check the semantics over and tell me if they aren't supposed to be like e.g. Date and other first class citizens in the XTC language / code. 

Problem: A `TimeZone:` literal lexes and parses and then **dies in the constant pool**, so any module containing one fails to compile with an internal error rather than a diagnostic.

```
pool.ensureLiteralConstant(Format.Date,     "x")  -> Date{value="x"}
pool.ensureLiteralConstant(Format.Time,     "x")  -> Time{value="x"}
pool.ensureLiteralConstant(Format.Duration, "x")  -> Duration{value="x"}
pool.ensureLiteralConstant(Format.TimeZone, "x")  -> IllegalStateException: unsupported format: TimeZone
```

Its siblings `Date`/`TimeOfDay`/`Time`/`Duration` are plumbed through. `TimeZone` was missed in **seven** places across three modules:

| # | Site | Symptom if only the earlier ones are fixed |
| --- | --- | --- |
| 1 | `ConstantPool.ensureLiteralConstant` | throws on construction |
| 2 | `ConstantPool.disassemble` | *separate list* — constructible but unreadable |
| 3-4 | `LiteralConstant`'s two switches | rejected |
| 5 | `LiteralConstant.getType()` | falls through to `Constant.getType()`, which **throws** |
| 6 | `xConst`'s runtime literal switch | **compiles**, then dies at run time on `VAR_IN` |
| 7 | `NativeContainer.getConstType` | still failed after 1–6 were all fixed |

### ⚠️ For review: this touches the standard library

It also adds `construct(String)` to `lib_ecstasy`'s `TimeZone`. The runtime path calls exactly that constructor, and `Date`, `Time`, `TimeOfDay`, `Duration`, `Version` and `Path` **each already have one** — `TimeZone` had only `TimeZone(Int64 picos)` and a conditional `of(String)`. The new constructor delegates to `of()` and throws `IllegalArgument` on a bad string, matching `Date`'s style. Flagging it explicitly because it is not a javatools-only change.

### Verified end to end

Compiling *and running* a module containing `TimeZone:` literals:

```
utc=UTC
plus0130=+01:30
```

That mattered: the unit ratchet went green twice while the literal was still broken, and only running a module exposed sites 6 and 7.

### Test

`LiteralFormatPlumbingTest` — a six-stage ratchet over every literal format the compiler produces: the pool builds it, it maps to a type, `disassemble` reads it back, `xConst` materialises it, the Ecstasy class has the `construct(String)` that path calls, and `NativeContainer` knows its implementing type. It grew from three stages to six because each stage was added only after a run disproved the previous "fix" — which is the argument for keeping all six. 

Ideally we should have an Enum or a sealed class hierarchy or something to switch on so the test methodology is somewhat messy, but still sound. 

---

## Also in this PR: the `RegEx` arm, and a ratchet so these two switches cannot drift again

While adding `case TimeZone:` I found the neighbouring `case RegEx:` has the same shape as the bug this PR fixes, one step further along: **it can only ever throw.**

`ConstantPool.ensureLiteralConstant` lists `RegEx` in its `LiteralConstant` group, so the switch reads as though RegEx literals are supported. `LiteralConstant` has no RegEx case — in either its validating constructor or the private one it delegates to first — so the arm falls straight to `default: throw new IllegalStateException("unsupported format: " + format)`.

Deleting it rather than implementing it, because routing is not available:

- `RegExConstant extends ValueConstant`, not `LiteralConstant`, so the 2-arg `ensureLiteralConstant(format, s)` overload's `(LiteralConstant)` cast would `ClassCastException`;
- `RegExConstant`'s only constructors are `(pool, format, DataInput)` and `(pool, String, int nFlags)` — there is no flag-free literal form;
- `ensureRegExConstant` is the real, working producer and is unaffected. Nothing in the repo passes `Format.RegEx` to `ensureLiteralConstant`, and `disassemble`'s `case RegEx:` is untouched — RegEx is a genuine wire format.

### The ratchet is the point

`TimeZone` was in the grammar, the lexer and the compiler, and missing from one switch. `RegEx` is in one switch and missing from the class. Same defect class, opposite direction — two lists of "formats a `LiteralConstant` can hold" that nothing kept in agreement.

`poolOffersExactlyTheFormatsLiteralConstantAccepts()` measures both sets by driving every `Format.values()` entry through the two entry points; neither side is hardcoded. The subtle part is that **both switches throw with byte-identical text**, so the message cannot distinguish "the pool refused" from "the pool routed it and `LiteralConstant` refused" — the test disambiguates on the stack instead, counting a format as *offered* only if the call reached a `LiteralConstant.<init>` frame.

It was red on this branch before the fix:

```
ensureLiteralConstant hands [RegEx] to LiteralConstant, whose constructor rejects it,
so those arms can only ever throw
expected: <[IntLiteral, FPLiteral, Date, TimeOfDay, TimeZone, Time, Duration, Path]>
 but was: <[IntLiteral, FPLiteral, RegEx, Date, TimeOfDay, TimeZone, Time, Duration, Path]>
```

and it catches the drift from the other side too: deleting `case TimeZone:` from the pool switch — the exact mistake this PR fixes — makes it fail with *"LiteralConstant accepts [TimeZone] that ensureLiteralConstant has no arm for"*. So it would have caught the original bug, and it will catch the next one.

`Format.Version` correctly lands in neither set: `LiteralConstant` rejects it (`"version requires VersionConstant subclass"`) and the pool routes versions to `ensureVersionConstant`.

### DIFF
--- a/lib_ecstasy/src/main/x/ecstasy/temporal/TimeZone.x
+++ b/lib_ecstasy/src/main/x/ecstasy/temporal/TimeZone.x
@@ -32,6 +32,26 @@ import TimeOfDay.PicosPerSecond;
 const TimeZone(Int64 picos) {
 
     /**
+     * Construct a TimeZone from an ISO-8601 timezone indicator string, as produced by a
+     * `TimeZone:` literal.
+     *
+     * This exists for the same reason `Date`, `Time`, `TimeOfDay`, `Duration`, `Version` and
+     * `Path` each have a String constructor: it is the entry point the runtime uses to turn a
+     * literal constant into a value.
+     *
+     * @param tz  an ISO-8601 timezone indicator string; see [of]
+     *
+     * @throws IllegalArgument  if the string is not a valid ISO-8601 timezone indicator
+     */
+    construct(String tz) {
+        if (TimeZone timezone := TimeZone.of(tz)) {
+            construct TimeZone(timezone.picos);
+        } else {
+            throw new IllegalArgument($"invalid ISO-8601 timezone: \"{tz}\"");
+        }
+    }
+
+    /**
      * Internal constructor for the special "NoTZ" timezone.
      */
     private construct() {
--- a/javatools/src/main/java/org/xvm/asm/ConstantPool.java
+++ b/javatools/src/main/java/org/xvm/asm/ConstantPool.java
@@ -360,6 +360,7 @@ public class ConstantPool
         case Date:
         case TimeOfDay:
         case Time:
+        case TimeZone:
         case Duration:
         case Path:
         case RegEx: {
@@ -2597,6 +2598,7 @@ public class ConstantPool
             case Date:
             case TimeOfDay:
             case Time:
+            case TimeZone:
             case Duration:
             case Path:
                 constant = new LiteralConstant(this, format, in);
--- a/javatools/src/main/java/org/xvm/asm/ConstantPool.java
+++ b/javatools/src/main/java/org/xvm/asm/ConstantPool.java
@@ -362,8 +362,7 @@ public class ConstantPool
         case Time:
         case TimeZone:
         case Duration:
-        case Path:
-        case RegEx: {
+        case Path: {
             LiteralConstant constant = (LiteralConstant) ensureLocatorLookup(format).get(s);
             if (constant == null) {
                 constant = register(new LiteralConstant(this, format, s, oValue));
--- a/javatools/src/main/java/org/xvm/asm/constants/LiteralConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/LiteralConstant.java
@@ -84,6 +84,10 @@ public class LiteralConstant
             // TODO
             break;
 
+        case TimeZone:
+            // TODO
+            break;
+
         case Duration:
             // TODO
             break;
@@ -125,6 +129,7 @@ public class LiteralConstant
         case Date:
         case TimeOfDay:
         case Time:
+        case TimeZone:
         case Duration:
         case Version:
         case Path:
@@ -171,6 +176,7 @@ public class LiteralConstant
             case Date       -> pool.typeDate();
             case TimeOfDay  -> pool.typeTimeOfDay();
             case Time       -> pool.typeTime();
+            case TimeZone   -> pool.typeTimeZone();
             case Duration   -> pool.typeDuration();
             case Version    -> pool.typeVersion();
             case Path       -> pool.typePath();
--- a/javatools/src/main/java/org/xvm/runtime/NativeContainer.java
+++ b/javatools/src/main/java/org/xvm/runtime/NativeContainer.java
@@ -791,7 +791,7 @@ public class NativeContainer
         case          Dec32,   Dec64,   Dec128,   DecN:
 
         case Array, UInt8Array:
-        case Date, TimeOfDay, Time, Duration:
+        case Date, TimeOfDay, Time, TimeZone, Duration:
         case Range, Path, Version, RegEx:
         case Module, Package:
         case Tuple:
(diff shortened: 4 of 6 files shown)


lib_ecstasy/src/main/x/ecstasy/temporal/TimeZone.x
```diff
@@ -31,6 +31,26 @@ import TimeOfDay.PicosPerSecond;
  */
 const TimeZone(Int64 picos) {
 
+    /**
```
ggleyzer, line 34: Why do we need this constructor? Just to throw this?
    
     new IllegalArgument($"invalid ISO-8601 timezone: \"{tz}\"");

lagergren: Adding the mechanical reason on top of the design one (which is, as much of the runtime should be written in XTC as possible, and as little in Java as possible): the runtime looks this constructor up by name and signature, so its absence is what broke the literal.

`xConst.initNative` resolves it directly:

```java
MethodStructure timeZoneConstruct = f_container.getClassStructure("temporal.TimeZone")
        .findMethod("construct", 1, pool.typeString());
```

and the literal path then dispatches to it:

```java
case TimeZone:
    clz         = ensureClass(container, pool.typeTimeZone());
    constructor = info.timeZoneConstruct();
```

With no `construct(String)`, `findMethod` finds nothing and a `TimeZone:` literal compiles and then dies at run time with `Unexpected op execution failure ... op=VAR_IN`.

On master, every other type reachable from a literal already has one — `Date`, `Time`, `TimeOfDay`, `Duration`, `Version` and `Path`. `TimeZone` was the only one without, which is precisely the gap this fixes rather than a new convention.

On "just to throw": the body isn't only a throw — it delegates to the existing `TimeZone.of(tz)` and constructs from the result, with the throw as the `else`. What it adds over `of()` is a non-conditional path, which is what a literal needs, since a literal cannot be conditional. Routing the runtime to `of()` instead would mean special-casing `TimeZone` in the Java template, i.e. moving the validation the other way.

ggleyzer: I think it just proves that xConst.initNative() is implemented incorrectly. I don't see any reason it cannot call TImeZome.of()

cpurdy (changes requested):
That new constructor is definitely going in the wrong direction. We had originally approached these types similar to what Java does, e.g. `new Integer("1234");` but this means that the only way to report failure is through an exception, so if you want to test "is this string a valid int?", you have to try..catch and construct a new Integer. So we started switching to conditional static methods that would parse the string and return true on success with the resulting value.

I'd drop the constructor and call the "of(..)" static method instead from xConst. If you need help with that change, let us know.

lagergren:
> That new constructor is definitely going in the wrong direction. We had originally approached these types similar to what Java does, e.g. `new Integer("1234");` but this means that the only way to report failure is through an exception, so if you want to test "is this string a valid int?", you have to try..catch and construct a new Integer. So we started switching to conditional static methods that would parse the string and return true on success with the resulting value.
> 
> I'd drop the constructor and call the "of(..)" static method instead from xConst. If you need help with that change, let us know.

Why isn't there a TimeZone.parse if that is what you want to do? I merely made the TimeZone look like all the other Date objects do. If that is not what you plan to do, I would still favor it being consistent when it goes in, and we file an issue to change its semantic behavior (including the other Date style classes) later, which would be the more correct thing to do in our gitflow.

lagergren:
Yeah I don't know - I always prefer to implement a bug fix so that it is consistent with the rest of the world and how they do stuff like, the IllegalState and its similar mirroring in the other Date literals, which is there today, and then filing an issue that these are a bad way to go, and then fix that issue in a new PR. That is what most people would preserve, to avoid a refactoring at the same time as a bug fix.  Happy to do that you can approve so we can at least get a working TImeZone literal in, which is one of my blockers, and then do it "properly", including the other literals that have the same behavior. 

Otherwise - please take this over, do what you think should be done to it at the same time as the bug fix and merge when ready. You have my blessing. I have reassigned it to you @cpurdy

PR #560, opened 2026-08-31
Range-check an array index before narrowing it

lagergren:
`extractArrayValue` receives the index as a `long`, and two implementations narrowed it to `int` *before* checking it:

```java
int nIx = (int) lIndex;

return nIx < 0 || nIx >= ach.length
        ? frame.raiseException(xException.outOfBounds(frame, lIndex, ach.length))
        : frame.assignValue(iReturn, xChar.makeHandle(ach[nIx]));
```

`(int)` keeps only the low 32 bits, so an index of 2<sup>32</sup> + 4 becomes 4, satisfies the guard, and reads element 4.

### Reproduction

```
s = "abcdefgh"; arr = [10,11,12,13,14,15,16,17]; i = 4294967300   // 2^32 + 4

before:  s[i]   ->  'e'                                        // wrong element, no error
         arr[i] ->  raises "Index 4294967300 out of range 0..7"

after:   s[i]   ->  raises "Index 4294967300 out of range 0..7"
         arr[i] ->  unchanged
```

`xArray` checks the un-narrowed value and gets it right, so a `String` and an `Int[]` disagreed about the same out-of-range index.

The exception each affected site builds already reported `lIndex` rather than the narrowed variable, which suggests the check was always meant to be against the full value.

### Fix

`IndexSupport.checkedIndex(long, int)` performs the narrowing and the range check together, returning `-1` when out of range. That is what makes the defect hard to reintroduce: a caller gets back an `int` that has already been checked, so there is no narrowed value lying around to guard on by mistake. A valid index is never negative, so the caller branches on that without a second check.

`xString.extractArrayValue` and `xRTType.extractArrayValue` both go through it.

### Tests

Two, each proving something the other cannot.

**`IndexSupportTest`** — a plain unit test of `checkedIndex`. Its central case walks every slot of a container and asserts that `2^32 + slot` is rejected even though it narrows to a valid slot, with a precondition asserting that it *does* narrow that way, so the test states the arithmetic it is about rather than assuming it. Also covers `Long.MIN_VALUE`/`MAX_VALUE`, an empty container, and other container sizes.

**`TestArray.testIndexBounds`** — asserts the behaviour end to end, and is the one that **fails before and passes after**:

- unfixed runtime: `checkedGet(s, 4294967300) == True`, assertion fails
- fixed runtime: raises `OutOfBounds`, prints `index bounds ok`

It covers `Int[]` alongside `String` since the two disagreed, plus the ordinary cases (`0`, `4`, `7`, `8`, `-1`) to show they are untouched. Verified in both directions on the same build, rebuilding the XDK in between.

### Not affected

`BaseInt128`'s shift operators narrow the same way and are safe: `LongLong.shl` masks the count with `& 0x7F`, and truncation preserves the low 32 bits, so the low 7 survive. Calling that out because the sites look identical.

### Verification

- `:javatools:test :javatools_utils:test` — 443 tests, 0 failures
- `xdk:installDist` builds; the before/after above is from that build with and without the patch

### Note on the earlier CI failure

The first run failed in `:manualTests:testXtc`, and it is unrelated to this change. The cause is in `OSDirectory.files()`:

```
names()
  .filter(name -> { if (node := find(name)) { ... } else { return False; } })   // absence is normal
  .map(name    -> { assert node := find(name); ... })                          // absence is impossible
```

`find(name)` is called twice per name and the two calls disagree about whether absence is possible. Several `testXtc` tasks generate xunit modules into a shared `/tmp/xunit`, so a file can vanish between them — the log shows `IllegalState: "File|Directory node := find(name)": name=JsonTest_xunit.xtc`. This change only alters behaviour for an index whose value is at least 2^32, so it cannot reach that path.


manualTests/src/main/x/array.x
```diff
@@ -369,4 +370,58 @@ module TestArray {
...
+        assert !checkedGet(s, lowBitsInRange);
+        assert !checkedGet(arr, lowBitsInRange);
+
+        // and the ordinary out-of-range cases still behave
+        assert !checkedGet(s, 8);
+        assert !checkedGet(s, -1);
+        assert !checkedGet(arr, 8);
+        assert !checkedGet(arr, -1);
+
+        console.print("index bounds ok");
```
ggleyzer: we are trying to minimize the console output in the manual tests; that last line carries very little value

ggleyzer (approved):

manualTests/src/main/x/array.x
lagergren: The console isn't even piped to stdout by default from the build? I can remove it, but it's the wrong thing to solve, I think.

lagergren: Removed in cc0af13 — you're right here: the method already prints its `** testIndexBounds()` header like the rest of the file, so dropping the trailing line doesn't make the section silent, it just stops announcing that asserts didn't throw. Also rebased onto master to clear the conflict with #563/#564, keeping all three array tests.

PR #562, opened 2026-08-31
Look a directory entry up once, not twice, when listing dirs and files

lagergren:
`OSDirectory.dirs()` and `files()` were both built as `names().filter(...).map(...)`, with a `find(name)` call in each half:

```
.filter(name -> {
    if (File|Directory node := find(name)) { return node.is(File); }
    else { return False; }                     // absence is normal here
})
.map(name -> {
    assert File|Directory node := find(name);  // absence is impossible here
    return node.as(File);
});
```

The two halves disagree about whether a missing node is possible. The filter treats it as ordinary — its comment says *"that is most probably a broken link"* — while the map asserts it cannot happen. So a directory mutating under the scan trips the assertion instead of skipping the entry.

### Where it fired

CI, when several xunit runners generated modules into a shared build directory:

```
IllegalState: "File|Directory node := find(name)": name=JsonTest_xunit.xtc
    at fs.OSDirectory.files() (OSDirectory.x:51)
    at mgmt.DirRepository.isCacheValid() (DirRepository.x:125)
    at tools.ModuleGenerator.ensureModule(...) (ModuleGenerator.x:35)
```

That particular trigger is removed separately by #561, which stops the runners sharing a directory. This is the underlying fragility: any directory changing under a scan can still reach it.

### Fix

`flatMap` does the whole thing in one lookup. That resolves the disagreement by removing the second question, and halves the filesystem calls a listing makes:

```
Iterator<File> files() {
    return names().flatMap(name -> {
        if (File|Directory node := find(name), node.is(File)) {
            return [node];
        }
        return [];
    });
}
```

### On testing — no red-before test, and why

The two `find()` calls are **adjacent per element**, because `filter` and `map` are both lazy. Only a genuinely concurrent writer can land between them.

I tried a single-threaded reproduction that deletes entries mid-iteration; it **passes on the unfixed code too**, because by the time the pipeline reaches a deleted name the filter already sees the absence and drops it before the map runs. So I am not claiming a test that fails before and passes after — there isn't one short of real concurrency, which would be flaky in CI.

`TestFiles.testListing` is therefore a **guard for the rewrite**, not a regression test. It asserts the listings themselves (5 files, 1 directory) and that entries removed mid-iteration are skipped rather than raising. It passes before and after; its job is to keep the new implementation honest.

### Verification

- `:javatools:test :javatools_utils:test` green
- `xdk:installDist` builds
- `TestFiles` passes: `listing ok (files=5, dirs=1, after delete=3)`

---

### DIFF
--- a/javatools_bridge/src/main/x/_native/fs/OSDirectory.x
+++ b/javatools_bridge/src/main/x/_native/fs/OSDirectory.x
@@ -36,10 +36,6 @@ const OSDirectory
      * A name obtained from the listing may no longer resolve: it may be a broken link, or may have
      * been removed since the listing was taken. Such a name is simply not a node now, so it is
      * dropped.
-     *
-     * Each name is resolved exactly once. Resolving it a second time - to test it and then again
-     * to convert it - lets the two lookups disagree, and a directory changing under the scan then
-     * trips whichever of them assumed the node was still there.
      */
     private Iterator<FileNode> nodes() {
         return names().flatMap(name -> {
--- a/javatools_bridge/src/main/x/_native/fs/OSDirectory.x
+++ b/javatools_bridge/src/main/x/_native/fs/OSDirectory.x
@@ -21,36 +21,26 @@ const OSDirectory
 
     @Override
     Iterator<Directory> dirs() {
-        return names()
-            .filter(name -> {
-                if (File|Directory node := find(name)) {
-                    return node.is(Directory);
-                } else {
-                    // that is most probably a broken link
-                    return False;
-                }
-            })
-            .map(name -> {
-                assert File|Directory node := find(name);
-                return node.as(Directory);
-            });
+        return names().flatMap(name -> {
+            if (File|Directory node := find(name), node.is(Directory)) {
+                return [node];
+            }
+            // the name came from a listing, but the node may be a broken link, or may have been
+            // removed since that listing was taken; either way it is simply not a directory now
+            return [];
+        });
     }
 
     @Override
     Iterator<File> files() {
-        return names()
-            .filter(name -> {
-                if (File|Directory node := find(name)) {
-                    return node.is(File);
-                } else {
-                    // that is most probably a broken link
-                    return False;
-                }
-            })
-            .map(name -> {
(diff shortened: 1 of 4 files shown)


javatools_bridge/src/main/x/_native/fs/OSDirectory.x
```diff
@@ -21,36 +22,32 @@ const OSDirectory
...
+    }
+
+    /**
+     * The nodes that this directory currently contains.
+     *
+     * A name obtained from the listing may no longer resolve: it may be a broken link, or may have
+     * been removed since the listing was taken. Such a name is simply not a node now, so it is
+     * dropped.
+     *
+     * Each name is resolved exactly once. Resolving it a second time - to test it and then again
```
ggleyzer: The change looks very good! The only thing I'd suggest to omit is the second paragraph in the doc here. The comment talks about two lookups, but there is only one here. I believe this comment belongs to the PR description.

manualTests/src/main/x/files.x
```diff
@@ -144,4 +145,54 @@ module TestFiles {
         // this will force the caller to wait
         return done;
     }
+
+    /**
+     * Covers `Directory.dirs()` and `files()`, including an entry that disappears while the
+     * (lazy) iteration is in progress.
+     *
+     * Both were implemented as `names().filter(find).map(find)` - two lookups per name, where the
```
ggleyzer: Ditto here - the comment cans only be understood in a historical context and should be moved to the PR description instead

ggleyzer (approved):
Looks very good; just a couple of doc related comments

javatools_bridge/src/main/x/_native/fs/OSDirectory.x
lagergren: Done in 0fd9ac1 — the historical explanation is in the PR description instead.

manualTests/src/main/x/files.x
lagergren: Done in 0fd9ac1 — the historical explanation is in the PR description instead.

PR #563, opened 2026-08-31
Fix deleteAll(range): wrong elements on byte-backed arrays, crash on String arrays

lagergren:
Two defects in the same operation. They were found together, and they have the same cause.  (This issue also illustrates that it is actually quite problematic to bypass the Java type system, which would have facilitated this being a compile time error instead of two runtime issues, and we should really revisit that policy. It is not the first thing I ran into at runtime that could have been reduced to Java compile time during the LSP project.) 

## What breaks

Array storage is chosen per element type — a byte-backed delegate for `Int8`/`UInt8`, a `String[]` one for `String`, an `ObjectHandle[]` one for the rest — and each implements `deleteRangeImpl` itself.

**1. Byte-backed arrays silently return the wrong elements.**

```
Int8 [1,2,3,4,5].deleteAll(1..2)  ->  [1, 3, 4]      expected [1, 4, 5]
UInt8 likewise                    ->  0x010304       expected 0x010405
```

The copy starts at the wrong offset:

```java
System.arraycopy(abValue, nIndex + 1, abValue, nIndex, cSize - nIndex - nDelete);
//                        ^^^^^^^^^^ should be nIndex + nDelete
```

`+1` is correct in `deleteElementImpl` directly above, which is where this looks copied from. Deleting a *single* element agrees either way, which is why it survived. **No exception — just wrong data.**

**2. `String` arrays crash.**

```
ClassCastException: StringArrayHandle cannot be cast to GenericArrayDelegate
    at xRTDelegate.deleteRangeImpl(xRTDelegate.java:681)
    at xArray.invokeDeleteAll(xArray.java:765)
```

`xRTStringDelegate` has no `deleteRangeImpl`, so it inherits the one on `xRTDelegate` — which *is* the implementation for object arrays and starts by casting to `GenericArrayDelegate`. Again only for a multi-element range, since a single-element delete takes the `deleteElementImpl` branch, which `String` does implement.

Checked every element type with its own storage. `Int64`, `Int16`, `Int128`, `Float64`, `Char`, `Boolean`, `Bit`, `Nibble` and `Object` are correct; `Int8`, `UInt8` and `String` were not.

## Why both existed

Both are consequences of the storage protocol being declared in terms of a supertype that carries no information:

```java
protected void deleteRangeImpl(DelegateHandle hTarget, long lIndex, long cDelete)
```

Nine methods on `xRTDelegate` take a `DelegateHandle`, and twenty implementations across the hierarchy open by casting it back to the type they actually store. Two things follow directly, and both happened here:

- **A missing implementation is invisible.** `xRTStringDelegate` inherits an implementation written for a *different* storage type, and nothing objects, because `DelegateHandle` accepts it. Had the protocol been declared over the handle type each delegate stores, the inherited body would not have type-checked against `StringArrayHandle` and this would have been a compile error rather than a crash reported from user code.
- **Nothing relates one implementation to another.** The byte-backed and generic versions are the same algorithm over different storage, but the type system is told nothing that would let a reader — or a compiler — line them up. The `+1` sat next to a correct `+ nDelete` in a sibling file for as long as it took someone to delete two elements from an `Int8[]`.

This is the ordinary cost of routing a typed operation through an untyped parameter: the compiler is holding the information needed to catch both of these, and the signature throws it away. The fix here is the two lines; the reason there were two lines to fix is the signature.

`xRTDelegate` cannot be generified as it stands, because it is simultaneously the abstract base and the concrete object-array implementation, so its defaults construct a `GenericArrayDelegate`. Splitting that concrete role out is the prerequisite, and is a separate change.

## Tests

`TestArray.testDeleteRange` covers every element type that has its own storage, deleting a **multi-element** range because that is what neither defect survives. On the unfixed runtime it fails immediately:

```
IllegalState: "new Array<Int8>(Mutable, [1,2,3,4,5]).deleteAll(1..2) == [1, 4, 5]": ... = [1, 3, 4]
    at testDeleteRange() (array.x:390)
```

and the `String` case throws after it. It also keeps the single-element path, which always worked, and a range reaching the end of the array, where the copy is skipped entirely.

## Verification

- `:javatools:test :javatools_utils:test` green
- `xdk:installDist` builds
- the new test passes on this branch and fails on `master`, verified by running the same compiled module against both runtimes

### DIFF
--- a/manualTests/src/main/x/array.x
+++ b/manualTests/src/main/x/array.x
@@ -24,6 +24,7 @@ module TestArray {
 
         testComparable();
         testAggregation();
+        testDeleteRange();
 
         testIterators();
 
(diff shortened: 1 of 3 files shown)


manualTests/src/main/x/array.x
```diff
@@ -369,4 +370,45 @@ module TestArray {
         console.print($"col={col}");
         console.print($"bytes={col.contents}");
     }
+
+    /**
+     * `deleteAll(range)` across every element type that has its own storage.
+     *
+     * Array storage is chosen per element type - a byte-backed delegate for `Int8`, a `String[]`
```
ggleyzer, line 451: similar comment I made on the other PR. My guess is that this part of the comment was written by an agent and while absolutely correct, belongs to the PR description, but not here

ggleyzer (approved):

PR #564, opened 2026-08-31
Fix ClassCastException when comparing two array slices or views by reference

lagergren:
## The bug

Comparing the references of two slices, or of two views, kills the program with a raw Java `ClassCastException` escaping into Ecstasy instead of answering the comparison.

```
Int[] a  = [1, 2, 3, 4, 5];
Int[] s1 = a[1..3];
Int[] s2 = a[1..3];

s1 == s2      // True - fine
&s1 == &s2    // Unhandled exception: Run-time error: java.lang.ClassCastException
```

```
UInt8[] b = [1, 2, 3, 4];
&b.asBitArray() == &b.asBitArray()   // same crash, ByteBasedBitView$ViewHandle
```

Verified on clean master `145f12f51`:

```
java.lang.ClassCastException: class ...xRTSlicingDelegate$SliceHandle cannot be cast to
class ...xRTDelegate$GenericArrayDelegate
        at ...xRTDelegate.compareIdentity(xRTDelegate.java:266)
        at ...xArray.compareIdentity(xArray.java:555)
        at ...xRef$CompareReferents.doNext(xRef.java:1227)
```

## Cause

`xRTDelegate` is the base of every array delegate, but its `compareIdentity` is not a base implementation — it is the **object-array** implementation, and it opens with:

```java
GenericArrayDelegate h1 = (GenericArrayDelegate) hValue1;
```

`SliceHandle` holds `f_hSource`/`f_ofStart`; `ViewHandle` holds its source. Neither is a `GenericArrayDelegate`, and neither overrode `compareIdentity`, so both inherited a cast that cannot succeed.

`createDelegate`, `callEquals`, `insertElementImpl` and `deleteElementImpl` have the same problem, but those four are unreachable — `xArray` answers `callEquals` itself, and Ecstasy rejects `insert`/`delete` on a fixed-size slice before any delegate is asked. `compareIdentity` is the one that escapes.

Quick fix: two added overrides, no change to existing code. Each compares what the handle actually holds — same mutability and size, plus start offset and direction for a slice, then `compareIdentity` on the source delegate. Both guard with `instanceof` rather than casting, so a slice compared against a non-slice answers `false` instead of raising. 

## Second direction, second set of templates

Reviewing the above surfaced that the reverse direction crashes too, in *different* templates, so
this PR now covers both:

```
Char[] a = ['a','b','c','d','e'];
Char[] s = a[1..3];

&s == &a    // xRTSlicingDelegate  — the case above
&a == &s    // xRTCharDelegate     — this one
```

`xRef.CompareReferents` resolves the template from the **first** handle and hands both to it. Handle 1 is safe by construction: it chose the template. Handle 2 is arbitrary. Every concrete delegate cast both, so any pair of same-element-type arrays with different storage crashed in whichever direction put the concrete delegate first.

Seven templates had the identical two-cast opening — `xRTCharDelegate`, `xRTStringDelegate`, `xRTFloat64Delegate`, `ByteBasedDelegate`, `LongBasedDelegate`, `LongLongDelegate`, and `xRTDelegate` itself — and each now tests rather than casts. The contract is stated once, on `ClassTemplate.compareIdentity` where implementors will read it, rather than repeated above each guard.

## Testing

`TestArray.testSliceIdentity` covers the slice case, the view case, and all seven concrete delegates in both directions, with `assert` rather than printing.

- **Red on master:** dies at `array.x:84` with the `ClassCastException` above; the view case dies the same way with `ByteBasedBitView$ViewHandle`.
- **Each delegate verified individually** against clean master `145f12f51`, one module per element type so the first failure cannot mask the rest:

```
Char     master=CRASH  fixed=ok
String   master=CRASH  fixed=ok
Float64  master=CRASH  fixed=ok
Int8     master=CRASH  fixed=ok
Int      master=CRASH  fixed=ok
Int128   master=CRASH  fixed=ok
Object   master=CRASH  fixed=ok
```

- **Green after:** `&nums[1..3] == &nums[1..3]` → `True`, `&nums[1..3] == &nums[0..2]` → `False`, `&bytes.asBitArray() == &bytes.asBitArray()` → `True`, `&chars == &chars[1..3]` → `False`.
- Full `array.x` suite: 0 unhandled exceptions.
- `:javatools:test --rerun-tasks --no-build-cache`: 343 tests, 0 failures, 0 errors.

## Footnote: this could have been a compile error, and immediately spotted

This is worth noting because it is the same root cause as #563 and #560 before it: none of this would have had to show up at runtime. If the base declares the storage protocol `abstract` rather than supplying one subclass's implementation as the default, `javac` names every affected class directly. Applying that to `compareIdentity` alone, on current master, prints:

```
xRTSlicingDelegate.java:23: error: xRTSlicingDelegate is not abstract and does not override
  abstract method compareIdentity(ObjectHandle,ObjectHandle) in xRTDelegate
xRTViewToBit.java:21:     error: xRTViewToBit is not abstract and does not override ...
xRTViewFromByte.java:18:  error: xRTViewFromByte is not abstract and does not override ...
xRTViewFromBit.java:16:   error: xRTViewFromBit is not abstract and does not override ...
```

Four classes, named with file and line, on the first build. This is the actual output, and it is how #563 was found: the compiler pointed straight at the missing `deleteRangeImpl` the instant the base stopped covering for it.

Finding this one took the opposite route: noticing that a base method casts to one specific subclass's handle, enumerating the delegates that are not that subclass, working out which of the five affected methods are actually reachable from Ecstasy (four are not - `xArray` answers `callEquals` itself, and the fixed-size guard rejects `insert`/`delete`), then building a repro for the one that is. That is a fair amount of work to reach something the compiler would have handed over for free, and it is hard to be confident the sweep was exhaustive - honestly, it is exhaustive for these five methods, and the same shape may exist elsewhere.

This PR deliberately changes no architecture - it is two overrides, purely additive, safe on its own. But the "Object/instanceof/cast" design decision is getting increasingly expensive for me, which is worrying. We could absolutely re-type the template boundaries incrementally, so it is not a "must change the entire world" kind of thing. I will try to put up a few simple PRs later to show how I think we can rather easily tighten this up.

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTSlicingDelegate.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTSlicingDelegate.java
@@ -172,10 +172,7 @@ public class xRTSlicingDelegate
     }
 
     /**
-     * A slice does not own element storage, so the inherited implementation - which is the
-     * object-array one, and opens by casting to {@code GenericArrayDelegate} - cannot apply to a
-     * {@link SliceHandle}. A slice's identity is the source it projects plus the window it
-     * projects through.
+     * A slice's identity is the source it projects plus the window it projects through.
      */
     @Override
     public boolean compareIdentity(ObjectHandle hValue1, ObjectHandle hValue2) {
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTSlicingDelegate.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTSlicingDelegate.java
@@ -170,4 +170,31 @@ public class xRTSlicingDelegate
             return super.toString() + " @" + f_ofStart;
         }
     }
+
+    /**
+     * A slice does not own element storage, so the inherited implementation - which is the
+     * object-array one, and opens by casting to {@code GenericArrayDelegate} - cannot apply to a
+     * {@link SliceHandle}. A slice's identity is the source it projects plus the window it
+     * projects through.
+     */
+    @Override
+    public boolean compareIdentity(ObjectHandle hValue1, ObjectHandle hValue2) {
+        if (hValue1 == hValue2) {
+            return true;
+        }
+
+        if (!(hValue1 instanceof SliceHandle h1) || !(hValue2 instanceof SliceHandle h2)) {
+            return false;
+        }
+
+        DelegateHandle hSource1 = h1.f_hSource;
+        DelegateHandle hSource2 = h2.f_hSource;
+
+        return h1.getMutability() == h2.getMutability()
+            && h1.m_cSize         == h2.m_cSize
+            && h1.f_ofStart       == h2.f_ofStart
+            && h1.f_fReverse      == h2.f_fReverse
+            && hSource1.getTemplate() == hSource2.getTemplate()
+            && hSource1.getTemplate().compareIdentity(hSource1, hSource2);
+    }
 }
--- a/manualTests/src/main/x/array.x
+++ b/manualTests/src/main/x/array.x
@@ -97,14 +97,42 @@ module TestArray {
         console.print($"&bytes.asBitArray() == &bytes.asBitArray(): {&v1 == &v2}");
 
         // The reverse direction dispatches to the CONCRETE delegate, which resolves the template
-        // from the first handle only and used to cast the second one too.
-        Char[] chars = ['a', 'b', 'c', 'd', 'e'];
-        Char[] cs    = chars[1..3];
+        // from the first handle only and used to cast the second one too. Each element type below
+        // is backed by a different delegate, and every one had the same defect, so each needs its
+        // own case: array-vs-slice and slice-vs-array must both answer rather than raise.
+        Char[]    chars   = ['a', 'b', 'c', 'd', 'e'];
+        String[]  strs    = ["a", "b", "c", "d", "e"];
+        Float64[] floats  = [1.0, 2.0, 3.0, 4.0, 5.0];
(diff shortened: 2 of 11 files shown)


manualTests/src/main/x/array.x
```diff
@@ -68,6 +69,72 @@ module TestArray {
...
+
+        assert &charsSlice   != &chars;
+        assert &strsSlice    != &strs;
+        assert &floatsSlice  != &floats;
+        assert &int8sSlice   != &int8s;
+        assert &intsSlice    != &ints;
+        assert &int128sSlice != &int128s;
+        assert &objsSlice    != &objs;
+
+        console.print("array-vs-slice identity answers for all seven delegates");
```
ggleyzer: Similar to the other comment - not much value is the last "print"

javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTSlicingDelegate.java
```diff
@@ -170,4 +170,31 @@ public String toString() {
             return super.toString() + " @" + f_ofStart;
         }
     }
+
+    /**
+     * A slice does not own element storage, so the inherited implementation - which is the
```
ggleyzer: I think we could place this comment on the super; reducing the duplication of identical text

ggleyzer (approved):

manualTests/src/main/x/array.x
lagergren: Done — see 43b16d4.

javatools/src/main/java/org/xvm/runtime/template/_native/collections/arrays/xRTSlicingDelegate.java
lagergren: Done — see 43b16d4.

PR #565, opened 2026-09-01
Add support for immutable xvm structures

cpurdy:
This is the next chunk of work to help support bundles, efficient caching, etc. It's not a fix to anything in and of itself, but a building block that we need. As an aside, there are going to now be opportunities to efficiently replace things like array based APIs with e.g. List, now that we have mutable vs immutable in the structures.

### DIFF
--- a/javatools/src/main/java/org/xvm/asm/Component.java
+++ b/javatools/src/main/java/org/xvm/asm/Component.java
@@ -2054,7 +2054,7 @@ public abstract class Component
             List<Contribution> listClone = new ArrayList<>(listContribs.size());
 
             for (Contribution listContrib : listContribs) {
-                listClone.add((Contribution) listContrib.clone());
+                listClone.add(that.new Contribution(listContrib));
             }
             that.m_listContribs = listClone;
         }
@@ -2157,29 +2157,13 @@ public abstract class Component
     // ----- XvmStructure methods ------------------------------------------------------------------
 
     @Override
-    protected Component ensureMutable() {
-        return (Component) super.ensureMutable();
-    }
-
-    @Override
-    protected Component ensureReadOnly() {
-        for (Component sibling = getEldestSibling(); sibling != null;
-                sibling = sibling.getNextSibling()) {
-            if (!sibling.isReadOnly()) {
-                sibling.prepareReadOnly();
-                sibling.markReadOnly();
+    protected void markReadOnly() {
+        if (!isReadOnly()) {
+            List<Contribution> contributions = m_listContribs;
+            if (contributions != null) {
+                contributions.forEach(Contribution::markReadOnly);
             }
-        }
-        return this;
-    }
-
-    /**
-     * Detach any externally accessible mutable state before this component becomes read-only.
-     */
-    protected void prepareReadOnly() {
-        List<Contribution> contributions = m_listContribs;
-        if (contributions != null) {
-            contributions.forEach(Contribution::prepareReadOnly);
+            super.markReadOnly();
         }
     }
 
@@ -2698,8 +2682,26 @@ public abstract class Component
      * abstract sense, meaning any class, interface, mixin, const, enum, or service) can be composed
      * of any number of contributing components.
      */
-    public class Contribution
-            implements Cloneable {
+    public class Contribution {
+        /**
+         * Copy a Contribution onto this Contribution's containing Component.
+         *
+         * @param that  the Contribution to copy
+         */
+        private Contribution(Contribution that) {
+            m_composition  = that.m_composition;
+            m_typeContrib  = that.m_typeContrib;
+            m_constProp    = that.m_constProp;
+            m_annotation   = that.m_annotation;
+            m_mapParams    = that.m_mapParams == null
+                    ? null
+                    : new ListMap<>(that.m_mapParams);
+            m_constInjector = that.m_constInjector;
+            m_listInject    = that.m_listInject == null
+                    ? null
+                    : new ArrayList<>(that.m_listInject);
+        }
+
         /**
          * @see XvmStructure#disassemble(DataInput)
          */
@@ -3040,7 +3042,7 @@ public abstract class Component
         /**
          * Detach mutable values that may have been supplied by a caller.
          */
-        private void prepareReadOnly() {
+        private void markReadOnly() {
             if (m_listInject != null) {
                 m_listInject = Collections.unmodifiableList(new ArrayList<>(m_listInject));
             }
@@ -3290,15 +3292,6 @@ public abstract class Component
                     }
                 }
                 break;
-            }
-        }
-
-        @Override
-        protected Object clone() {
-            try {
-                return super.clone();
-            } catch (CloneNotSupportedException e) {
-                throw new IllegalStateException(e);
             }
         }
 
--- a/javatools/src/main/java/org/xvm/asm/Component.java
+++ b/javatools/src/main/java/org/xvm/asm/Component.java
@@ -2157,12 +2157,12 @@ public abstract class Component
     // ----- XvmStructure methods ------------------------------------------------------------------
 
     @Override
-    public Component ensureMutable() {
+    protected Component ensureMutable() {
         return (Component) super.ensureMutable();
     }
 
     @Override
-    public Component ensureReadOnly() {
+    protected Component ensureReadOnly() {
         for (Component sibling = getEldestSibling(); sibling != null;
                 sibling = sibling.getNextSibling()) {
             if (!sibling.isReadOnly()) {
@@ -2184,14 +2184,14 @@ public abstract class Component
     }
 
     @Override
-    protected Component findCorrespondingStructure(FileStructure file) {
-        Component parent     = (Component) getContaining();
-        Component parentCopy = parent.findCorrespondingStructure(file);
+    protected Component findThisIn(FileStructure thatFileStructure) {
+        Component thisParent = (Component) getContaining();
+        Component thatParent = thisParent.findThisIn(thatFileStructure);
 
         IdentityConstant idCopy = (IdentityConstant)
-                file.getConstantPool().getConstant(getIdentityConstant());
-        Component original = parent.getChild(getIdentityConstant());
-        Component copy     = parentCopy.getChild(idCopy);
+                thatFileStructure.getConstantPool().getConstant(getIdentityConstant());
+        Component original = thisParent.getChild(getIdentityConstant());
+        Component copy     = thatParent.getChild(idCopy);
 
         while (original != this && original != null && copy != null) {
             original = original.getNextSibling();
--- a/javatools/src/main/java/org/xvm/asm/Component.java
+++ b/javatools/src/main/java/org/xvm/asm/Component.java
@@ -2172,20 +2172,22 @@ public abstract class Component
         Component thisParent = (Component) getContaining();
         Component thatParent = thisParent.findThisIn(thatFileStructure);
 
-        IdentityConstant idCopy = (IdentityConstant)
-                thatFileStructure.getConstantPool().getConstant(getIdentityConstant());
-        Component original = thisParent.getChild(getIdentityConstant());
-        Component copy     = thatParent.getChild(idCopy);
+        IdentityConstant idThis = this.getIdentityConstant();
+        IdentityConstant idThat = (IdentityConstant)
+                thatFileStructure.getConstantPool().getConstant(idThis);
 
-        while (original != this && original != null && copy != null) {
-            original = original.getNextSibling();
-            copy     = copy.getNextSibling();
+        Component thisChild = thisParent.getChild(getIdentityConstant());
+        Component thatChild = thatParent.getChild(idThat);
+
+        while (thisChild != this && thisChild != null && thatChild != null) {
+            thisChild = thisChild.getNextSibling();
+            thatChild = thatChild.getNextSibling();
         }
 
-        if (original != this || copy == null) {
+        if (thisChild != this || thatChild == null) {
             throw new IllegalStateException("unable to locate component in FileStructure copy");
         }
-        return copy;
+        return thatChild;
     }
 
     @Override
@@ -2224,7 +2226,6 @@ public abstract class Component
 
     @Override
     protected void markModified() {
-        verifyMutable();
         m_fModified = true;
     }
 
@@ -2300,7 +2301,6 @@ public abstract class Component
     @Override
     protected void registerConstants(ConstantPool pool) {
         assert getContaining() == null || getContaining() instanceof Component;
-        verifyMutable();
 
         m_constId = pool.register(m_constId);
         m_cond    = pool.register(m_cond);
--- a/javatools/src/main/java/org/xvm/asm/Component.java
+++ b/javatools/src/main/java/org/xvm/asm/Component.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Set;
 
 import java.util.function.Consumer;
@@ -254,6 +255,8 @@ public abstract class Component
      * @param idNew  the new identity to use for this component
      */
     protected void replaceThisIdentityConstant(IdentityConstant idNew) {
+        verifyMutable();
+
         IdentityConstant idOld = m_constId;
         for (Iterator<Component> iter = siblings(); iter.hasNext(); ) {
             iter.next().m_constId = idNew;
@@ -271,6 +274,8 @@ public abstract class Component
      * @param idNew  the new identity to use instead of the old identity
      */
     protected void replaceChildIdentityConstant(IdentityConstant idOld, IdentityConstant idNew) {
+        verifyMutable();
+
         // nothing to do unless the name changed, which we don't support anyhow
         assert idOld.getName().equals(idNew.getName());
     }
@@ -298,6 +303,7 @@ public abstract class Component
         int nFlagsOld = m_nFlags;
         int nFlagsNew = (nFlagsOld & ~ACCESS_MASK) | (access.ordinal() << ACCESS_SHIFT);
         if (nFlagsNew != nFlagsOld) {
+            verifyMutable();
             m_nFlags = (short) nFlagsNew;
             markModified();
         }
@@ -319,6 +325,7 @@ public abstract class Component
         int nFlagsOld = m_nFlags;
         int nFlagsNew = (nFlagsOld & ~ABSTRACT_BIT) | (fAbstract ? ABSTRACT_BIT : 0);
         if (nFlagsNew != nFlagsOld) {
+            verifyMutable();
             m_nFlags = (short) nFlagsNew;
             markModified();
         }
@@ -340,6 +347,7 @@ public abstract class Component
         int nFlagsOld = m_nFlags;
         int nFlagsNew = (nFlagsOld & ~STATIC_BIT) | (fStatic ? STATIC_BIT : 0);
         if (nFlagsNew != nFlagsOld) {
+            verifyMutable();
             m_nFlags = (short) nFlagsNew;
             markModified();
         }
@@ -361,6 +369,7 @@ public abstract class Component
         int nFlagsOld = m_nFlags;
         int nFlagsNew = (nFlagsOld & ~SYNTHETIC_BIT) | (fSynthetic ? SYNTHETIC_BIT : 0);
         if (nFlagsNew != nFlagsOld) {
+            verifyMutable();
             m_nFlags = (short) nFlagsNew;
             markModified();
         }
@@ -382,6 +391,7 @@ public abstract class Component
         int nFlagsOld = m_nFlags;
         int nFlagsNew = (nFlagsOld & ~COND_RET_BIT) | (fConditional ? COND_RET_BIT : 0);
         if (nFlagsNew != nFlagsOld) {
+            verifyMutable();
             m_nFlags = (short) nFlagsNew;
             markModified();
         }
@@ -402,6 +412,7 @@ public abstract class Component
         int nFlagsOld = m_nFlags;
         int nFlagsNew = (nFlagsOld & ~AUXILIARY_BIT) | AUXILIARY_BIT;
         if (nFlagsNew != nFlagsOld) {
+            verifyMutable();
             m_nFlags = (short) nFlagsNew;
             markModified();
         }
@@ -466,11 +477,9 @@ public abstract class Component
      */
     public List<Contribution> getContributionsAsList() {
         List<Contribution> list = m_listContribs;
-        if (list == null) {
-            return Collections.emptyList();
-        }
-        assert (list = Collections.unmodifiableList(m_listContribs)) != null;
-        return list;
+        return list == null
+                ? Collections.emptyList()
+                : Collections.unmodifiableList(list);
     }
 
     /**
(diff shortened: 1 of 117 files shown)


javatools/src/main/java/org/xvm/asm/Component.java
```diff
@@ -2599,8 +2682,26 @@ public static Composition valueOf(int i) {
...
+        private Contribution(Contribution that) {
+            m_composition  = that.m_composition;
+            m_typeContrib  = that.m_typeContrib;
+            m_constProp    = that.m_constProp;
+            m_annotation   = that.m_annotation;
+            m_mapParams    = that.m_mapParams == null
+                    ? null
+                    : new ListMap<>(that.m_mapParams);
+            m_constInjector = that.m_constInjector;
+            m_listInject    = that.m_listInject == null
```
lagergren: I would use the empty map to mean absent instead of null. That way we can get rid of mutable Fields.

javatools/src/main/java/org/xvm/asm/FileStructure.java
```diff
@@ -1282,6 +1343,8 @@ protected void disassemble(DataInput in)
      * Re-registers all referenced constants with the pool.
      */
     public void reregisterConstants(boolean fOptimize) {
+        // this is not considered a "mutation" of the FileStructure, although there is the potential
+        // for internal changes
         ConstantPool pool = m_pool;
         pool.preRegisterAll();
         registerConstants(pool);
```
lagergren: I would make this chainable. 

return registerConstants(m_pool.preRegisterAll())

And instead of void, you could have a useable return value containing the reregistered constants or empty list if no work was done or something. That is a better API.

javatools/src/main/java/org/xvm/asm/Component.java
```diff
@@ -605,8 +623,10 @@ protected Contribution addContribution(Contribution contrib) {
      */
     public void removeContribution(Contribution contrib) {
         List<Contribution> list = m_listContribs;
-        if (list != null) {
+        if (list != null && list.contains(contrib)) {
```
lagergren: Why isn't listContribs always a final list instance. If it is empty it means "no contribs".  Would simplify the code and it would prevent unnecessary nullability issues and checks.

javatools/src/main/java/org/xvm/asm/Component.java
```diff
@@ -577,6 +593,8 @@ public Contribution addIncorporates(TypeConstant constClass,
      * @param contrib  the contribution to add to the end of the list
      */
     protected Contribution addContribution(Contribution contrib) {
+        verifyMutable();
+
         List<Contribution> list = m_listContribs;
         if (list == null) {
             m_listContribs = list = new ArrayList<>();
```
lagergren: Same here. Keep that list around - why check if it is null and ther create it - you get one bug prone non final field, you get unnecessary code, and any contribution can just go through the list even if it is empty, without creating nullability logic everywhere. This is common xtc pattern. 

Same goes for many other things in this file.

javatools/src/main/java/org/xvm/asm/PropertyStructure.java
```diff
@@ -197,7 +202,7 @@ public Annotation[] getPropertyAnnotations() {
             aAnno = m_aPropAnno;
             assert aAnno != null;
         }
-        return aAnno;
+        return isReadOnly() ? aAnno.clone() : aAnno;
     }
```
lagergren: Arrays.copyOf

javatools/src/main/java/org/xvm/asm/PropertyStructure.java
```diff
@@ -697,6 +708,27 @@ public void addAnnotation(Annotation annotation) {
...
+    @Override
+    protected void markReadOnly() {
+        if (!isReadOnly()) {
+            if (m_aPropAnno != null) {
+                m_aPropAnno = m_aPropAnno.clone();
+            }
+            if (m_aRefAnno != null) {
+                m_aRefAnno = m_aRefAnno.clone();
+            }
+            super.markReadOnly();
```
lagergren: Get rid of the clones

lagergren:
This is a building block I want, and some of it is more careful than it looks — `verifyMutable()` is called as a statement rather than inside an `assert`, so it survives `-da`, and `Contribution.markReadOnly()` *copies* the collections, which quietly neutralises any reference a caller took while the structure was still mutable. I nearly filed that second one as a leak before working out why it's safe. 

One question decides most of my comments, and I'd rather ask than assume: **can a structure be frozen after it has been shared, or is the rule "build, freeze, then publish"?**

If it's build-freeze-publish, then `m_fReadOnly` as a plain `boolean` is *right* and cheaper than `volatile` (even though the overhead is negligible) — publication supplies the happens-before edge for the flag and everything reachable from it. I'd only want that contract written down, because nothing states it today and the API makes the other order equally easy to write: `ensureReadOnly()` acts on a live object, `markReadOnly()` is itself a mutation, and `ensureMutable()`/`clone()` mint a mutable structure from a frozen one at an arbitrary later time.

Worth flagging either way: #549 is open and argues `MethodStructure.m_fNative` and `m_code` must be `volatile` because they're "written by one thread and read by others". `MethodStructure` is one of the classes this PR adds `m_fReadOnly` to. Those two PRs should agree before either lands — and if #549's premise is right, this one's answer follows.

The larger thing, not blocking: this makes immutability *advisory* — a runtime check every current and future mutator has to remember, with no way to say "I require an immutable structure" in a signature. That's fine as a first step, but it's worth deciding now whether the destination is structural (a distinct immutable type, or freeze-by-construction returning a new instance with `final` fields), because that changes what's worth building on top of this. I would DEFINITELY like something that has an implicit before and after edge like finals rather than volatiles or "these mutable fields are definitely not going to be in a racy situation ever."

javatools/src/main/java/org/xvm/asm/XvmStructure.java
```diff
@@ -587,4 +688,9 @@ public String toString() {
      * The containing XVM structure.
      */
     private XvmStructure m_xsParent;
+
+    /**
+     * True iff this XVM structure cannot be modified.
+     */
+    private boolean m_fReadOnly;
```
lagergren, line 695: This is the field the whole feature rests on, and it is the one thing I can't derive the contract for from the diff.

If the lifecycle is strictly build -> freeze -> publish, a plain `boolean` is genuinely fine and I'd only ask for that contract to be written down here. If a structure can be frozen *after* being shared, two things break: another thread can keep observing `false` and mutate through `verifyMutable()`, and even once it observes `true`, nothing orders the writes to the structure's *other* fields against that thread's reads — so it can see a frozen structure with stale contents.

The reason I don't think this is hypothetical for these exact classes: #549 argues that `MethodStructure`'s fields need `volatile` precisely because they are touched from more than one thread, and `MethodStructure` is one of the classes that inherits `m_fReadOnly` here. Both can't be right about the same objects.

Cheapest resolution if the ordering isn't already guaranteed: make it `volatile`. On a field this cold the cost isn't measurable and it removes the question. The better one is freeze-by-construction — `toImmutable()` returning a new instance whose fields are `final` — because then the ordering can't be got wrong by anyone later.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
```diff
@@ -575,6 +668,14 @@ public int hashCode() {
     @Override
     public abstract boolean equals(Object obj);
 
+    @Override
+    protected XvmStructure clone()
```
lagergren, line 672: `Object.clone()` is shallow, so this hands back a "mutable" structure that shares every referenced sub-object with the frozen original — the only difference between the two is the boolean cleared on the next line. Anything a subclass doesn't explicitly deep-copy is now writable through the clone while the original still answers `isReadOnly() == true`.

It's also the opposite direction from where the rest of the codebase is heading. A copy constructor per type gives a correct return type instead of a cast from `Object`, can't silently share what the author forgot about, and doesn't quietly start sharing one more thing the day someone adds a field.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
```diff
@@ -160,6 +163,96 @@ public Iterable<? extends XvmStructure> getContained() {
...
+
+    /**
+     * Obtain a mutable version of this XVM structure. If this structure is read-only, then the
+     * entire containing FileStructure is copied and the corresponding structure from that mutable
+     * copy is returned.
+     *
+     * @return this XVM structure if it is mutable, otherwise its counterpart in a mutable copy of
+     *         the FileStructure that contains this XVM structure
+     */
+    protected XvmStructure ensureMutable() {
```
lagergren, line 195: `new FileStructure(getFileStructure())` copies the **entire containing file** and then searches the copy for the counterpart. So calling something named `ensureMutable` on a single `Parameter` copies a whole module, and nothing at the call site hints at that.

Could the cost show up in the name or the return type? Failing that, a bulk entry point so a caller converting many structures pays once — and at minimum a first javadoc sentence that says it copies. The current one is just "Obtain a mutable version of this XVM structure."

javatools/src/main/java/org/xvm/asm/XvmStructure.java
```diff
@@ -160,6 +163,96 @@ public Iterable<? extends XvmStructure> getContained() {
...
+    }
+
+    /**
+     * Verify that this XVM structure can be modified.
+     *
+     * @return true
+     *
+     * @throws IllegalStateException  if this XVM structure is read-only
+     */
+    protected final boolean verifyMutable() {
```
lagergren, line 180: Every call site in this PR uses this as a statement, which is right and is the careful part. My worry is the shape rather than the current uses: a method that returns `true` and otherwise throws looks exactly like something meant to be written `assert verifyMutable();` — and written that way it silently stops checking anything under `-da`.

There's precedent for that going wrong a few lines from here. `Contribution.getTypeParams()` (Component.java:3038) still has `assert (map = Collections.unmodifiableMap(map)) != null;` — an assignment inside an assert, so the map stops being protected at all when assertions are off.

Making this `void` would mean the mistake isn't expressible.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
```diff
@@ -160,6 +163,96 @@ public Iterable<? extends XvmStructure> getContained() {
...
+                markReadOnly();
+            }
+        }
+        return this;
+    }
+
+    /**
+     * Mark this structure and all structures contained by it as read-only.
+     */
+    protected void markReadOnly() {
```
lagergren, line 224: Freezing here covers exactly what `getContained()` reports. Any state a subclass holds that isn't reachable that way stays quietly mutable inside a structure that answers `isReadOnly() == true`.

That's an easy thing to get wrong once and never notice, and it fails open rather than loudly. Worth either a per-subclass test, or a ratchet that enumerates declared fields and asserts each is frozen, immutable by type, or explicitly exempt.

javatools/src/main/java/org/xvm/asm/Component.java
```diff
@@ -2926,10 +3032,25 @@ public Map<StringConstant, TypeConstant> getTypeParams() {
...
+                return Collections.unmodifiableMap(map);
+            }
             assert (map = Collections.unmodifiableMap(map)) != null;
             return map;
         }
 
+        /**
+         * Detach mutable values that may have been supplied by a caller.
+         */
+        private void markReadOnly() {
```
lagergren, line 3045: `m_listInject` is copied **and** wrapped unmodifiable; `m_mapParams` is only copied. The read accessors wrap both, so I don't think this is a hole today — but the asymmetry is how holes get born, and a reader can't tell which of the two is the intended pattern here.

The copy itself is the good part and worth keeping either way: it detaches any reference a caller got hold of while the structure was still mutable.

javatools/src/main/java/org/xvm/asm/Component.java
```diff
@@ -2829,7 +2932,9 @@ public SingletonConstant getInjector() {
          *         indicates that all injections are handled by the injector
          */
         public List<Injection> getInjections() {
```
lagergren, line 2934: Related to the asymmetry just below: while the contribution is mutable this returns the raw `m_listInject` to the caller, and only wraps it once `isReadOnly()` is true.

I think this is safe *in this PR* — `markReadOnly()` copies the list before wrapping, so a reference handed out earlier points at the old list and can't be used to mutate the frozen one. But the safety comes from the copy at freeze time, three thousand lines away, not from anything visible here. As written the accessor reads like a leak, and the next person to touch either half has to reconstruct that argument to know it isn't.

Worth a comment saying the copy at freeze is what makes this safe — or just returning an unmodifiable view unconditionally.

lagergren (commented):
Line-level notes to go with my summary comment above. The first one is the only one I'd call blocking; the rest are readability and contract questions.

cpurdy:
> One question decides most of my comments, and I'd rather ask than assume: **can a structure be frozen after it has been shared, or is the rule "build, freeze, then publish"?**

Ideally, when we load from disk (or from byte[] or whatever), we will either get a read-only form, or we will need to say "instantiate this thing in a mutable form" or "instantiate this thing in a read-only form".

When we create one from scratch (e.g. a compiler), it is a mutable form. Obviously.

From a mutable form, it can be transitioned to a read-only form. From a read-only form, it must be copied to obtain a mutable form. 

When mutable, the data model is not intended to be (nor is in any way) thread safe. Trying to use it in a multi-threaded manner when it is mutable is a fool's errand. Don't do that.

When read-only, the data model should be thread safe. I cannot (yet) warrant that it is, but it should be reasonably close outside of tests explicitly intended to show otherwise.

Again, the intention is simple: single threaded if mutable, and only after explicitly made immutable should any thread even get to look at the structure.

> If it's build-freeze-publish, then `m_fReadOnly` as a plain `boolean` is _right_ and cheaper than `volatile` (even though the overhead is negligible) — publication supplies the happens-before edge for the flag and everything reachable from it.

Something like this. I'm trying to retrofit this on top of an "organically evolved" data structure. It's not going to immediately come out as perfect.

> `markReadOnly()` is itself a mutation

Sure. But note that it does nothing to something that is already read-only. So the model is consistent.

> Worth flagging either way: #549 is open and argues `MethodStructure.m_fNative` and `m_code` must be `volatile` because they're "written by one thread and read by others".

Actually, that issue is one of the reasons that I did this project and did it *now*. The idea of _"written by one thread and read by others"_ is obviously incompatible with what I've described above.

> The larger thing, not blocking: this makes immutability _advisory_ — a runtime check every current and future mutator has to remember, with no way to say "I require an immutable structure" in a signature.

Correct. This is Java, so we're largely out of luck in this regard anyhow.

The point is: As we utilize the read-only option, e.g. by default on loading, tools that need a mutable form will need to explicitly ask for a mutable form, or they'll blow up. I'm ok with that as a short term problem to fix. And when we do fix it, it should eliminate some of the bleed-through issues that either already do happen or that we fear could happen.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
cpurdy: I don't view this as being an issue. If it's read-only, the reads are safe. If it's not read-only, you can't be reading it from another thread. These data structures are for us, and used only by us; these are not for random use. As long as we build following these rules, we're fine. If you want to add text to the comment to explain this in more detail, do so, but keep it small and precise.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
cpurdy: Your point is reasonable. That would have been a fine alternative implementation. I'm not going to re-implement it at this point. This work was done in a hurry, to help you achieve your goals without completely derailing the JIT project, but it's already kept me off the JIT project for too long.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
cpurdy: You're correct. Don't call ensureMutable() on a Parameter if you don't know what it does. That could be bad.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
cpurdy: Focusing just on what was changed in this PR: Being able to add verifyMutable() at the top of any mutating method to enforce the contract is a reasonable approach. You're not going be mutating something that's read-only, and if you do, your hand will get crushed by the lid of the cookie jar coming down. That's the goal. This is us, protecting us, from us.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
cpurdy: In general: If getContained() is wrong, then nothing works. These are our structures; we need to make sure that they work.

javatools/src/main/java/org/xvm/asm/Component.java
cpurdy: We should clean up and tighten these as we visit them. I'm not willing to rewrite the entirety of the package as part of this PR. We need to get reasonably close with this first go, i.e. to prevent us from hurting us by not following the contract that we are mandating on ourselves. This isn't (as of yet anyhow) a shut-down of a CVE; it's just us putting in place a contract for ourselves to follow to avoid stepping on our own feet.

javatools/src/main/java/org/xvm/asm/Component.java
cpurdy: You're welcome to add a comment to it. If you had the question now, it's likely you'd have the same question in the future. Just keep the comments precise and targeted (i.e. low cost for the reader, but containing the necessary info).

javatools/src/main/java/org/xvm/asm/XvmStructure.java
lagergren: Yeah but I want the lid of the cookie jar to shout that that is exactly what I am doing wrong when I am doing it, not any moment later manifesting itself as cursed weird runtime behavior, is all.

lagergren:
> > One question decides most of my comments, and I'd rather ask than assume: **can a structure be frozen after it has been shared, or is the rule "build, freeze, then publish"?**
> 
> Ideally, when we load from disk (or from byte[] or whatever), we will either get a read-only form, or we will need to say "instantiate this thing in a mutable form" or "instantiate this thing in a read-only form".
> 
> When we create one from scratch (e.g. a compiler), it is a mutable form. Obviously.
> 
> From a mutable form, it can be transitioned to a read-only form. From a read-only form, it must be copied to obtain a mutable form.
> 
> When mutable, the data model is not intended to be (nor is in any way) thread safe. Trying to use it in a multi-threaded manner when it is mutable is a fool's errand. Don't do that.
> 
> When read-only, the data model should be thread safe. I cannot (yet) warrant that it is, but it should be reasonably close outside of tests explicitly intended to show otherwise.
> 
> Again, the intention is simple: single threaded if mutable, and only after explicitly made immutable should any thread even get to look at the structure.
> 
> > If it's build-freeze-publish, then `m_fReadOnly` as a plain `boolean` is _right_ and cheaper than `volatile` (even though the overhead is negligible) — publication supplies the happens-before edge for the flag and everything reachable from it.
> 
> Something like this. I'm trying to retrofit this on top of an "organically evolved" data structure. It's not going to immediately come out as perfect.
> 
> > `markReadOnly()` is itself a mutation
> 
> Sure. But note that it does nothing to something that is already read-only. So the model is consistent.
> 
> > Worth flagging either way: #549 is open and argues `MethodStructure.m_fNative` and `m_code` must be `volatile` because they're "written by one thread and read by others".
> 
> Actually, that issue is one of the reasons that I did this project and did it _now_. The idea of _"written by one thread and read by others"_ is obviously incompatible with what I've described above.
> 
> > The larger thing, not blocking: this makes immutability _advisory_ — a runtime check every current and future mutator has to remember, with no way to say "I require an immutable structure" in a signature.
> 
> Correct. This is Java, so we're largely out of luck in this regard anyhow.
> 
> The point is: As we utilize the read-only option, e.g. by default on loading, tools that need a mutable form will need to explicitly ask for a mutable form, or they'll blow up. I'm ok with that as a short term problem to fix. And when we do fix it, it should eliminate some of the bleed-through issues that either already do happen or that we fear could happen.

This is the contract I wanted, and I guess I am just looking for ways that the code clearly enforces it, so that I cannot accidentally not do this.  But as I've said all the time - this is fine in the current environment and merge it as is.  Would be cool to show you how a "replaced representation" example will "work" now e.g. Lists instead of Arrays in ways it didn't "work" before. And some unit tests that exercise the API would be nice mainly as documentation.

lagergren (approved):

lagergren:
> > > One question decides most of my comments, and I'd rather ask than assume: **can a structure be frozen after it has been shared, or is the rule "build, freeze, then publish"?**
> > 
> > 
> > Ideally, when we load from disk (or from byte[] or whatever), we will either get a read-only form, or we will need to say "instantiate this thing in a mutable form" or "instantiate this thing in a read-only form".
> > When we create one from scratch (e.g. a compiler), it is a mutable form. Obviously.
> > From a mutable form, it can be transitioned to a read-only form. From a read-only form, it must be copied to obtain a mutable form.
> > When mutable, the data model is not intended to be (nor is in any way) thread safe. Trying to use it in a multi-threaded manner when it is mutable is a fool's errand. Don't do that.
> > When read-only, the data model should be thread safe. I cannot (yet) warrant that it is, but it should be reasonably close outside of tests explicitly intended to show otherwise.
> > Again, the intention is simple: single threaded if mutable, and only after explicitly made immutable should any thread even get to look at the structure.
> > > If it's build-freeze-publish, then `m_fReadOnly` as a plain `boolean` is _right_ and cheaper than `volatile` (even though the overhead is negligible) — publication supplies the happens-before edge for the flag and everything reachable from it.
> > 
> > 
> > Something like this. I'm trying to retrofit this on top of an "organically evolved" data structure. It's not going to immediately come out as perfect.
> > > `markReadOnly()` is itself a mutation
> > 
> > 
> > Sure. But note that it does nothing to something that is already read-only. So the model is consistent.
> > > Worth flagging either way: #549 is open and argues `MethodStructure.m_fNative` and `m_code` must be `volatile` because they're "written by one thread and read by others".
> > 
> > 
> > Actually, that issue is one of the reasons that I did this project and did it _now_. The idea of _"written by one thread and read by others"_ is obviously incompatible with what I've described above.
> > > The larger thing, not blocking: this makes immutability _advisory_ — a runtime check every current and future mutator has to remember, with no way to say "I require an immutable structure" in a signature.
> > 
> > 
> > Correct. This is Java, so we're largely out of luck in this regard anyhow.
> > The point is: As we utilize the read-only option, e.g. by default on loading, tools that need a mutable form will need to explicitly ask for a mutable form, or they'll blow up. I'm ok with that as a short term problem to fix. And when we do fix it, it should eliminate some of the bleed-through issues that either already do happen or that we fear could happen.
> 
> This is the contract I wanted, and I guess I am just looking for ways that the code clearly enforces it, so that I cannot accidentally not do this. But as I've said all the time - this is fine in the current environment and merge it as is. Would be cool to show you how a "replaced representation" example will "work" now e.g. Lists instead of Arrays in ways it didn't "work" before. And some unit tests that exercise the API would be nice mainly as documentation.

BTW - There are ways to do som of this in Java.  I would define a trivial wrapper FrozenArray<ElementTypeGeneric>, which will, in all situations be absolutely as performant as a raw array, but with none of the mutability and Object:s.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
lagergren: It is protected, so it is fine, but these are things we should think about should not be physically or API-wise possible, and not just "if you dont know this explodes, that is bad", but this if fine for now.

javatools/src/main/java/org/xvm/asm/XvmStructure.java
lagergren: We need tests for getContained in various problematic envs in the future.

PR #568, opened 2026-09-01
jsondb: do not discard a rollback failure that follows a failed commit

### LINKED ISSUE #567: jsondb: a rollback failure after a failed commit is silently discarded
## Summary

When a commit fails, `Client.commit()` runs a compensating rollback — and discards any exception that rollback throws:

`lib_jsondb/src/main/x/jsondb/Client.x` (around line 1427):

```
} catch (Exception e) {
    log($"Exception during commit of {this}: {e}");
    result = DatabaseError;
    try {
        txManager.rollback(writeId_);
    } catch (Exception ignore) {}     // ← the secondary failure disappears
}
```

The commit failure is logged. The rollback failure is not.

## Why it matters

These two failures are not equivalent, and the second one is the more serious.

A commit that fails and rolls back cleanly is a *handled* outcome: the transaction did not happen, and `DatabaseError` says so accurately. A commit that fails and whose **rollback also fails** is different in kind — at that point the client no longer knows the transaction's disposition. It may be partially applied. `DatabaseError` is still returned, so the caller cannot tell the two situations apart.

`TxManager` owns recovery, and this issue is not proposing that `Client` take that over. But the secondary failure is precisely the evidence a health check or a recovery path needs in order to know that recovery is required at all — and today it is discarded at the only place it is observable.

The failure mode is silence: no log line, no exception, no signal. If it ever happens in production the first symptom is inconsistent data with nothing in the log to explain it.

## Suggested fix

Log the rollback failure, carrying the original commit failure as context, so both are visible. That is a one-line change in the catch and does not alter control flow or the returned `CommitResult`.

## Related

- #541 documents that `Catalog.log()` performs `@Inject Console` on every call and that injector contention under concurrent load can form a permanent wait cycle. Anything that adds a `log()` call — including the fix suggested above — should be aware of it. The call here is asynchronous (`catalog.log^(msg)`) and only on a double-failure path, so the added pressure is minimal, but caching that injection (recommendation 1 in #541) would remove the question entirely.
- #543 covers the same theme — a failure that disappears instead of reaching a caller — in the embedding and hot-reuse path rather than in a library.

## Notes

Found while auditing swallowed exceptions across the runtime and libraries. This is a latent defect: I have not reproduced a rollback failure in practice, and I am not claiming an observed data-loss incident. What is certain is that if it does occur, nothing records it.

lagergren:
Fixes #567.

When a commit fails, `Client.commit()` runs a compensating rollback and throws away anything that rollback raises:

```
} catch (Exception e) {
    log($"Exception during commit of {this}: {e}");
    result = DatabaseError;
    try {
        txManager.rollback(writeId_);
    } catch (Exception ignore) {}     // ← the secondary failure disappears
}
```

The commit failure is logged. The rollback failure is not.

## Why the second failure is the one worth keeping

These two outcomes are not equivalent, and the current code cannot tell them apart.

A commit that fails and rolls back cleanly is *handled*: the transaction did not happen, and `DatabaseError` describes it accurately. A commit that fails and whose **rollback also fails** is different in kind — the client no longer knows the transaction's disposition, and it may be partially applied. Both return `DatabaseError`, so a caller sees the same thing either way.

`TxManager` owns recovery and this change does not touch that. But the secondary failure is exactly the evidence that tells a health check or a recovery path that recovery is *needed*, and it was being discarded at the only point where it is observable.

The failure mode is silence — no log line, no exception, no signal. If it happens, the first symptom is inconsistent data with nothing in the log to explain it.

## The change

One catch. Control flow and the returned `CommitResult` are unchanged; the rollback failure is logged with the original commit failure as context, so both are visible together:

```
log($"Exception during rollback after failed commit of {this}: {e2} (commit failure: {e})");
```

## A note on #541, since this adds a `log()` call

#541 documents that `Catalog.log()` does `@Inject Console` on every call, and that injector contention under concurrent load can form a permanent wait cycle. This change adds a `log()` on a path that reaches it.

I do not think it is a meaningful contribution to that pressure: `Client.log` dispatches asynchronously (`catalog.log^(msg)`), so the committing fiber does not block on it, and this line only runs when a commit **and** its compensating rollback have both failed. But it is worth naming rather than leaving for a reviewer to notice, and #541's own first recommendation — caching the injected `Console` per `Catalog` service instance — would remove the question entirely.

## Testing, stated honestly

`JsondbClientRollbackFailureTest` fails on unmodified master and passes after:

```
org.opentest4j.AssertionFailedError: rollback failure after commit failure
must not be swallowed ==> expected: <false> but was: <true>
```

**It is a source-shape test** — it reads `Client.x` off disk and asserts on the shape of the error path, rather than provoking a real rollback failure. I would rather say that plainly than let it read as a behavioural reproduction. Inducing a genuine `TxManager.rollback` failure underneath a genuine commit failure needs fault injection that does not exist in this codebase today; the defect is a discarded exception whose absence is not otherwise observable. `:javatools:test` goes from 348 to 349 tests with the same 40 pre-existing environment-gated skips, and `xdk:installDist` compiles `lib_jsondb` clean.

## Not included

There are nine other `catch (… ignore)` sites across `lib_json`, `lib_jsondb/tools`, `lib_web`, `lib_xenia`, `lib_xunit_engine` and `manualTests`. They are unrelated to this path and would be a separate tidy — the house idiom for a genuinely unused catch variable is `catch (X _)`, which `Client.x` already uses at line 218.


javatools/src/test/java/org/xvm/lib/jsondb/JsondbClientRollbackFailureTest.java
ggleyzer, line 1: Marcus, did you write this test? I don't think so.

My next question is: did you read this test?

ggleyzer:
> A commit that fails and whose rollback also fails is different in kind — the client no longer knows the transaction's disposition, and it may be partially applied

I don't think this is a correct statement. A failed rollback doesn't leave a partially "applied" transaction. The only value in logging the exception is for developer's sake

ggleyzer:
Closing as "no fix"

PR #570, opened 2026-09-01
Keep one failing alarm from killing every timer in the JVM

lagergren:
`xLocalClock.TIMER` is a **single process-wide `java.util.Timer`** — one thread, shared by every alarm in every container in the JVM:

```java
public static Timer TIMER = new Timer("ecstasy:LocalClock", true);   // xLocalClock.java:282
```

The only thing scheduled on it is `Alarm.Trigger`, and its `run()` had no guard:

```java
protected static class Trigger extends TimerTask {
    @Override
    public void run() {
        f_alarm.run();          // nothing catches anything
    }
}
```

`java.util.Timer`'s own thread catches nothing. **An exception escaping `TimerTask.run()` terminates that thread and cancels the Timer permanently**, so every later `TIMER.schedule(...)` — at `:128`, `:226` and `:239` — throws `IllegalStateException: Timer already cancelled`.

So one bad alarm silently disables **every alarm in the whole process**.

## Why this is nasty rather than merely wrong

The failure does not surface where it happens. A dead `Timer` means alarms simply never fire, so what you observe is a **hang** — in a different container, arbitrarily later, with nothing in the log connecting it to the alarm that died. There is no error, no degraded mode, no signal at all.

It is reachable today. `Alarm.run()` calls `WeakCallback.extractCallback()`, which on master ends in a bare `throw new IllegalStateException()` whenever the context is gone or the callback id was already removed — precisely the shutdown/cancel race a timer thread is most likely to hit.

## The fix

Guard at the `TimerTask` boundary so the timer thread survives, and **record** the failure rather than discarding it.

Deliberately **not** a blanket `catch` around `context.callLater(...)`: that would swallow real errors at the one place a diagnostic is useful. The guard belongs at the boundary that must not die, not around the call that might legitimately fail.

**Both triggers are guarded, not just `xLocalClock`'s.** `xNanosTimer`'s `Trigger` schedules onto the same static `TIMER` and calls the same throwing method, so fixing one would leave the identical JVM-wide kill switch armed. Its test is red on master too.

31 lines added, 9 removed, across 2 files. Nothing changes for a program that works today — this only adds a guard where there was none.

## Proof

`AlarmTimerSurvivesCallbackFailureTest` — **8 failures / 8 runs** on master, **0 / 8** after. The test log shows the thread dying:

```
Exception in thread "test-alarm-timer" java.lang.IllegalStateException
  at org.xvm.runtime.WeakCallback.extractCallback(WeakCallback.java:37)
  at ...xLocalClock$Alarm.run(xLocalClock.java:232)
  at ...xLocalClock$Alarm$Trigger.run(xLocalClock.java:267)
  at java.base/java.util.TimerThread.mainLoop(Timer.java:572)
```

It is not timing-luck. `java.util.Timer` runs everything on one thread, so the test schedules a real alarm on a **private** Timer at 500 ms, drops the callback at ~1 ms (a 500× margin), queues a marker task at 1500 ms and awaits 10 s — the marker can only fire if that thread outlived the failure. The timer thread is a daemon with a bounded join, so a failure ends the test cleanly rather than hanging CI.

`:javatools:test` 348 → 350, same 40 pre-existing environment-gated skips.

## Relationship to the other clock PR

This is independent of the callback-registry work in the companion PR, in both directions, and either can land first — they conflict textually in `xLocalClock.java` and `xNanosTimer.java`, so whichever goes second wants a rebase.

The distinction is worth stating: that PR removes the throwers we currently know about. This one says the timer thread must survive one we do not. If a future thrower appears on this path — from `callLater`, from the fiber machinery — the kill switch is still armed without this change.

### DIFF
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/temporal/xLocalClock.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/temporal/xLocalClock.java
@@ -264,7 +264,19 @@ public class xLocalClock
 
             @Override
             public void run() {
-                f_alarm.run();
+                try {
+                    f_alarm.run();
+                } catch (Throwable e) {
+                    // must not happen - and must not escape either. TIMER is a single
+                    // process-wide java.util.Timer with one thread, shared by every alarm in
+                    // every container, and java.util.Timer kills that thread and cancels the
+                    // Timer permanently if a TimerTask throws. Letting this out would silently
+                    // disable every alarm in the JVM, surfacing much later as an unrelated hang.
+                    // TODO: report this through a runtime failure channel once one exists,
+                    //       rather than to stderr
+                    System.err.println("Unexpected alarm failure on the shared LocalClock timer");
+                    e.printStackTrace(System.err);
+                }
             }
 
             private final Alarm f_alarm;
--- a/javatools/src/main/java/org/xvm/runtime/template/_native/temporal/xNanosTimer.java
+++ b/javatools/src/main/java/org/xvm/runtime/template/_native/temporal/xNanosTimer.java
@@ -452,14 +452,24 @@ public class xNanosTimer
 
                 @Override
                 public void run() {
-                    Alarm alarm       = m_alarm;
-                    long cExtraMillis = alarm.checkReadyMillis();
-                    if (cExtraMillis == 0) {
-                        m_alarm = null;
-                        alarm.run();
-                    } else {
-                        // reschedule
-                        xLocalClock.TIMER.schedule(alarm.createTrigger(), cExtraMillis);
+                    try {
+                        Alarm alarm       = m_alarm;
+                        long cExtraMillis = alarm.checkReadyMillis();
+                        if (cExtraMillis == 0) {
+                            m_alarm = null;
+                            alarm.run();
+                        } else {
+                            // reschedule
+                            xLocalClock.TIMER.schedule(alarm.createTrigger(), cExtraMillis);
+                        }
+                    } catch (Throwable e) {
+                        // see xLocalClock.Alarm.Trigger.run(): these triggers are scheduled onto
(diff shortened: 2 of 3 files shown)


javatools/src/main/java/org/xvm/runtime/template/_native/temporal/xLocalClock.java
```diff
@@ -264,7 +264,19 @@ protected Trigger(Alarm alarm) {
 
             @Override
             public void run() {
-                f_alarm.run();
+                try {
```
ggleyzer, line 267: I disagree with this solution; `f_alarm.run()` doesn't run users' code, but our own carefully written logic. As you wrote - it must not throw. If it does, it's a our bug. 

What is the correct way to report our bugs? What good does it do for Michael to see this message
    "Unexpected alarm failure on the shared LocalClock timer"
in the terminal console of a Docker log file? Should we introduce an internal service that sends a http "post" to xqiz.it? Maybe. I'm willing to discuss all this when we are live with our product.
Is this a good investment of our time now - IMHO absolutely not

javatools/src/main/java/org/xvm/runtime/template/_native/temporal/xNanosTimer.java
```diff
@@ -452,14 +452,24 @@ protected Trigger(Alarm alarm) {
...
                 public void run() {
-                    Alarm alarm       = m_alarm;
-                    long cExtraMillis = alarm.checkReadyMillis();
-                    if (cExtraMillis == 0) {
-                        m_alarm = null;
-                        alarm.run();
-                    } else {
-                        // reschedule
-                        xLocalClock.TIMER.schedule(alarm.createTrigger(), cExtraMillis);
+                    try {
```
ggleyzer, line 455: ditto

ggleyzer (commented):
My recommendation is: defer.

Issue #572, opened 2026-09-01
Owner-bearing values in mutable public statics make containers last-writer-wins


lagergren:
## Summary

Native templates hold **owner-bearing values in mutable `public static` fields**, assigned from per-container `initNative()`. With more than one container in a JVM — sequentially or concurrently — the last container to initialise wins, and every earlier container's code then reads a value belonging to a different owner.

This is the residue left after the `INSTANCE` template cache is removed. That change routes *template lookup* through a container-owned table, but it does not touch this population, so "which container does this value belong to" is still answered by a static.

## The shape

All of these are `public static`, non-final, assigned during `initNative()`, and all of them carry an owner (a `TypeComposition` knows its container; a handle knows its composition):

| field | file |
|---|---|
| `ClassComposition CLASS` | `template/xObject.java:17` |
| `StringHandle EMPTY_STRING`, `EMPTY_ARRAY` | `template/text/xString.java:441-442` |
| `BooleanHandle TRUE`, `FALSE` | `template/xBoolean.java:19-20` |
| `EnumHandle NULL` | `template/xNullable.java:16` |
| the `*_ARRAY_CLZ` compositions | `template/collections/xArray.java` |
| `DELEGATES` | `template/_native/collections/arrays/xRTDelegate.java` |

…and others in the same style.

`xString.java:65` is representative: `EMPTY_STRING = new StringHandle(getCanonicalClass(), new char[0]);` — `getCanonicalClass()` is per-container, the field is not.

## Why it matters

Two failures, and the first needs no concurrency at all:

**Sequential.** Build container A, then container B in the same JVM. B's `initNative()` overwrites the statics. A is still live, and code acting on A's behalf now reads values whose composition — and therefore whose container, pool and type system — belong to B.

**Concurrent.** The fields are neither `volatile` nor `final`, and they are written during initialisation and read during execution, so there is no happens-before edge. A reader can observe a partially-initialised value, or a stale one indefinitely.

## Two places this shows through a public API

Worth naming because they look correct at the call site:

- **`xString.makeHandle(Container, char[])`** takes a container, but returns the global `EMPTY_STRING` for a zero-length array (`xString.java:409`) — so the argument is silently ignored in exactly the case where the caller was most explicit about ownership.
- **`xArray.makeBitArrayHandle` and friends** derive their owner *from* a static `TypeComposition`, so passing a container does not determine the result's owner either.

[...]

PR #573, opened 2026-09-01
Resolve native templates per container instead of through mutable statics

lagergren:
## The defect

Native templates published themselves into mutable `public static` fields from their constructors:

```java
public class xEnum extends xConst {
    public static xEnum INSTANCE;                      // plain mutable static

    public xEnum(Container container, ClassStructure structure, boolean fInstance) {
        super(container, structure, false);
        if (fInstance) {
            INSTANCE = this;                           // this-escape, during construction
        }
    }
}
```

**139 sites did this. 196 sites read the result.** None of the fields were `volatile` or `final`.

Templates are constructed **per container** (`NativeContainer:172-175`), and a JVM can build several — `InterpreterConnector:36` does `new NativeContainer(runtime, repository)`, and so does every embedding host and several tests.

### It is wrong in series, with no concurrency involved

Every `Connector` builds its own root: `InterpreterConnector`'s constructor is `new Runtime()` plus `new NativeContainer(...)`, and `Runner` builds a Connector per run. A host that runs several modules in one JVM therefore has several native containers alive at once — the Gradle plugin's direct mode does exactly this, in one cached classloader.

All ~139 native templates are constructed by a native container (`newInstance(this, structClass, Boolean.TRUE)`), so each new root overwrites the statics. Two native containers are **unrelated** — neither is an ancestor or descendant of the other — so code acting on root A's behalf then reads a template owned by root B: wrong container, wrong pool, wrong type system.

To be precise about what is *not* the problem: nested containers under a single root never construct native templates, and `getTemplate(String)` resolves through `getNativeContainer()`, so under one root `INSTANCE` holds an ancestor-owned template and reading it is legitimate. This bug needs a second root.

The clearest case is `Proxy.makeAsyncNativeHandle`, which does `new AsyncHandle(INSTANCE.f_container, method)`: **it takes a `Container` out of a JVM-global mutable static** — and with more than one root, that container can belong to an entirely different runtime.

Note that the reachability sweep cannot catch this: it deliberately does not enumerate `static` roots. That is a reason to delete the static, not to trust a clean sweep.

`xEnum.initNative()` also branches on `if (this == INSTANCE)`, so which object won the assignment decides whether native rebase initialisation runs at all. That is behaviour, not caching.

Add threads and it is additionally a `this`-escape publishing a partially-constructed object to a shared static with no safe publication.

## The change

Template lookup moves into one container-owned table (`NativeTemplates`), reached from the container that is asking rather than from a static. The dangerous pattern is expressed once, in one reviewable place, instead of being pasted 139 times.

**Zero `INSTANCE = this` and zero mutable `public static … INSTANCE` declarations remain.** Net **−75 lines** (187 files, +1639/−1714) despite adding the module.

Seven commits, each building independently: the table; the migration; a note recording why the `fInstance` constructor flag has to stay (it does not become redundant, contrary to my first assumption); two readability passes (container accessors, and renaming that flag to `fBaseTemplate`); a dead cast; and the test rework described below.

## Verification

- `./gradlew build` green; `:javatools:test` 351 tests, 0 failures, 40 skips (the standing set).
- End-to-end `xcc` + `xec` on a real module: output **byte-identical to master**.
- Commits 1 and 2 each verified to build on their own.

A review pass over my own change caught **two regressions I had introduced**, both from a scripted rewrite threading a container or frame that master deliberately leaves null: an NPE on `xRTType.makeHandle`'s foreign-handle path (three callers pass `null` intentionally), and an NPE on every native-throwable translation in `xException`, because `Utils.translate` passes a null frame. Both are fixed, folded into the commits that introduced them, and pinned by `ownerlessFactoryPathsStillWork`.

That pass also found composition caches I had bound to the *first caller's* container — which could be a short-lived nested container pinned forever in a static — where master bound them to the long-lived native container. Restored.

## Readability: say which container is being asked

Two accessors, in their own commit, because the migration otherwise made the diff harder to read than the code it replaced. Call sites had become:

```java
frame.f_context.f_container.nativeTemplates().get(xListMap.class)
```

Four hops to name one template — and when the lookup is four hops through another object's fields, a reader cannot see *whose* table is being asked, which is the question this PR exists to make answerable.

- `Frame.container()` — three lines; master had no such accessor.
- `Container.nativeTemplate(Class<T>)` and `Frame.nativeTemplate(Class<T>)` — a shorthand for the case that is 91% of all uses.

On the lines this PR introduces: `f_context.f_container` went **89 → 1**, and `nativeTemplates().get(...)` went **256 → 1**. Both survivors are deliberate — one chain starts from a `ServiceHandle`, which has no `container()`, and the other is `Container.nativeTemplate`'s own body.

`NativeTemplates.of(...)` keeps all four overloads; they are layered rather than redundant, and each now says so in its javadoc. `of(TypeComposition)` alone has 21 call sites, and neither it nor `of(ClassTemplate)` can be replaced by the shorthand, because those types have no owner accessor of their own. `ClassTemplate` deliberately did **not** gain a `nativeTemplate(...)` — `isNativeInstance` now reads `this == f_container.nativeTemplate(clz)`, which names the owner at the identity comparison instead of hiding it.

The ~198 pre-existing `f_context.f_container` sites elsewhere in the tree are untouched. That sweep is mechanical and obvious, but it does not belong inside a 181-file diff.

## `fInstance` renamed to `fBaseTemplate`

The flag used to mean "assign yourself to `INSTANCE`". That mechanism is gone, so the name was a fossil pointing at a removed mechanism — and with 151 declarations against 6 real consumers, a reader's first conclusion is that it is dead.

It is not. Removing it was attempted, and fails during bootstrap: `xService` is constructed both as the rebased base template over the `Service` interface and as the template for every concrete service class, and only the first may build a `NativeRebaseConstant`, whose constructor asserts its argument is an interface. The table cannot answer the question either, and that is by design — it is populated *after* a template finishes constructing, which is precisely the property that stops templates escaping mid-construction.

`fBaseTemplate` rather than `fRebase`, because five consumers turn it into a rebase but the sixth does not: `xRTConnector` passes `false` to its super and uses its own flag to build a process-wide user-agent string. "This is the container's base template for this class" is true at all 151 sites; "rebase" is not.

## What this does NOT fix — see #572

This removes the `INSTANCE` template cache. It does **not** end last-writer-wins across containers, because a large population of owner-bearing values is still held in mutable `public static` fields assigned from per-container `initNative()`: `xObject.CLASS`, `xString.EMPTY_STRING`/`EMPTY_ARRAY`, `xBoolean.TRUE`/`FALSE`, `xNullable.NULL`, `xArray`'s `*_ARRAY_CLZ` compositions, `xRTDelegate.DELEGATES`, and others.

Two places that leaks through a public API even after this change:

- `xString.makeHandle(Container, char[])` returns the global `EMPTY_STRING` for a zero-length array, ignoring the container it was handed;
- `xArray.makeBitArrayHandle` and friends derive their owner *from* a static `TypeComposition`.

Those are pre-existing and out of scope here, but they mean the guarantee this change provides is narrower than its name suggests, and I would rather say so than let the name carry more weight than the code. Filed as #572.

### DIFF
--- a/javatools/src/main/java/org/xvm/api/InterpreterConnector.java
+++ b/javatools/src/main/java/org/xvm/api/InterpreterConnector.java
@@ -95,7 +95,7 @@ public class InterpreterConnector
                 TypeConstant typeArg = method.getParam(0).getType();
 
                 assert typeStrings.isA(typeArg);
-                ahArg = new ObjectHandle[]{xString.makeArrayHandle(asArg)};
+                ahArg = new ObjectHandle[]{xString.makeArrayHandle(m_containerMain, asArg)};
             }
             break;
 
@@ -103,7 +103,7 @@ public class InterpreterConnector
             TypeConstant typeArg = method.getParam(0).getType();
             assert typeStrings.isA(typeArg);
             // the method requires an array that we can supply
-            ahArg = new ObjectHandle[]{xString.makeArrayHandle(asArg)};
+            ahArg = new ObjectHandle[]{xString.makeArrayHandle(m_containerMain, asArg)};
             break;
         }
         }
--- a/javatools/src/main/java/org/xvm/asm/OpVar.java
+++ b/javatools/src/main/java/org/xvm/asm/OpVar.java
@@ -147,7 +147,8 @@ public abstract class OpVar
         if (clzArray == null || !typeList.equals(typePrev)) {
             TypeConstant typeEl = typeList.resolveGenericType("Element");
 
-            clzArray = xArray.INSTANCE.ensureParameterizedClass(context.f_container, typeEl);
+            clzArray = context.f_container.nativeTemplates().get(xArray.class).
+                    ensureParameterizedClass(context.f_container, typeEl);
 
             context.setOpInfo(this, Category.Composition, clzArray);
             context.setOpInfo(this, Category.Type, typeList);
--- a/javatools/src/main/java/org/xvm/asm/OpVar.java
+++ b/javatools/src/main/java/org/xvm/asm/OpVar.java
@@ -147,7 +147,7 @@ public abstract class OpVar
         if (clzArray == null || !typeList.equals(typePrev)) {
             TypeConstant typeEl = typeList.resolveGenericType("Element");
 
-            clzArray = context.f_container.nativeTemplates().get(xArray.class).
+            clzArray = context.f_container.nativeTemplate(xArray.class).
                     ensureParameterizedClass(context.f_container, typeEl);
 
             context.setOpInfo(this, Category.Composition, clzArray);
--- a/javatools/src/main/java/org/xvm/asm/constants/TypeConstant.java
+++ b/javatools/src/main/java/org/xvm/asm/constants/TypeConstant.java
@@ -7812,7 +7812,7 @@ public abstract class TypeConstant
         }
 
         // don't cache a "foreign" handle
-        return xRTType.makeForeignHandle(this);
+        return xRTType.makeForeignHandle(container, this);
     }
 
     /**
--- a/javatools/src/main/java/org/xvm/asm/op/Assert.java
+++ b/javatools/src/main/java/org/xvm/asm/op/Assert.java
@@ -126,7 +126,7 @@ public class Assert
         ClassConstant    constClz    = (ClassConstant) idConstruct.getNamespace();
         ClassTemplate    template    = frame.ensureTemplate(constClz);
         ClassComposition clzTarget   = template.getCanonicalClass(frame.f_context.f_container);
-        StringHandle     hMsg        = xString.makeHandle(sMsg);
+        StringHandle     hMsg        = xString.makeHandle(frame, sMsg);
         ObjectHandle[]   ahArg       = new ObjectHandle[construct.getMaxVars()];
 
         ahArg[0] = hMsg;
--- a/javatools/src/main/java/org/xvm/asm/op/MoveRef.java
+++ b/javatools/src/main/java/org/xvm/asm/op/MoveRef.java
@@ -18,6 +18,7 @@ import org.xvm.javajit.RegisterInfo;
 
 import org.xvm.javajit.registers.Ref;
 
+import org.xvm.runtime.Container;
 import org.xvm.runtime.Frame;
 import org.xvm.runtime.ObjectHandle;
 import org.xvm.runtime.TypeComposition;
@@ -71,8 +72,9 @@ public class MoveRef
                     typeReg = infoSrc.getType();
                 }
             } else {
-                TypeComposition clzRef = xRef.INSTANCE.ensureParameterizedClass(
-                        frame.f_context.f_container, infoSrc.getType());
+                Container       container = frame.f_context.f_container;
+                TypeComposition clzRef    = container.nativeTemplates().get(xRef.class).
+                        ensureParameterizedClass(container, infoSrc.getType());
                 hRef    = new RefHandle(clzRef, frame, m_nFromValue);
                 typeReg = hRef.getType();
             }
--- a/javatools/src/main/java/org/xvm/asm/op/MoveRef.java
+++ b/javatools/src/main/java/org/xvm/asm/op/MoveRef.java
@@ -72,8 +72,8 @@ public class MoveRef
                     typeReg = infoSrc.getType();
                 }
             } else {
-                Container       container = frame.f_context.f_container;
-                TypeComposition clzRef    = container.nativeTemplates().get(xRef.class).
+                Container       container = frame.container();
+                TypeComposition clzRef    = container.nativeTemplate(xRef.class).
                         ensureParameterizedClass(container, infoSrc.getType());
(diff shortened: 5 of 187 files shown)


lagergren:
Made a mistake. Will reopen later.

ggleyzer:
I completely disagree with the analysis. The "fInstance" argument is passed as "true" only during the native container initialization, which is by design a single-threaded action.

There is a good point regarding `Proxy.makeAsyncNativeHandle()`. I'll take it under consideration.

My review decision: close; do not apply

ggleyzer:
I reviewed `Proxy.makeAsyncNativeHandle()` use and I think the code is correct. This method is only called for native methods, which means the methods that belong to types in the native type system. Therefore using the native container is the correct choice - it's used to create/cache class compositions that belong there

lagergren:
Rewrote the tests in this PR after re-reading them. Two problems with what was there, both mine.

**They argued the weaker of this change's two cases.** The tests built two native containers in series and checked the first still resolved its own templates. That depends on a container topology you've said goes against the design assumptions — and if that topology is ruled out, the defect those tests demonstrate is unreachable, so they could never settle anything. Meanwhile the case that needs no such agreement went untested: 139 copy-pasted mutable statics became one container-owned table.

**And they never ran.** They gated on compiled XDK output via `assumeTrue`, but in the javatools build the XDK is downstream — javatools is what produces the compiler that builds it. So they silently skipped and CI stayed green. That was true of the whole batch; I found it only when I forced one to execute, and it failed immediately on a real bug (`FileStructure.merge` asserting when a module already present is merged — it supersedes a fingerprint but not a real module). Ten days behind a green suite.

What's here now is two tests, both reading javatools' **compiled classes** — not XTC modules, nothing from the XDK. `compileJava` is upstream of `test`, so the inputs exist by construction and there is nothing to assume:

- **`theTemplateTreeIsFreeOfInstanceStatics`** — across all 318 compiled template classes, none declares an `INSTANCE` static and no code in the tree reads one. Red on master, green here.
- **`componentNamesFollowTheLoadersRule`** — the class-to-component-name mapping the table resolves through.

The reason for reading class files rather than source is your review time. You read four of the 139 sites, recognise the pattern, and the scan guarantees the other 135 didn't deviate — including a site that kept the old shape under a different spelling, which text-matching would miss. It turns a wide, shallow diff from a reading task into one file plus a proof.

Two things I want to be straight about.

**The scan keys on the name `INSTANCE`.** The stronger invariant — no static field of a `ClassTemplate` type anywhere in the tree — is what I tried first, and it fails: `xPackage.LIST_MAP_TEMPLATE`, `xEnum.RANGE_TEMPLATE`, `xRTComponentTemplate.MULTI_METHOD_TEMPLATE`, `xArray.MUTABILITY` and about six more are the same pattern under other names. They're worth converting and they are **not** in this PR. The javadoc names them so the test pins what was done rather than failing on work nobody claimed. I'd rather you hear that from me than find it yourself and read the PR as incomplete.

**`NativeTemplates.resolvedNames()` has no test here.** It's the capability the table was wanted for — asking which templates a container has actually pulled in is one call, where against 139 statics it wasn't expressible at all. Exercising it needs a live container, so that test belongs in the xdk build where the distribution exists, not here.

lagergren:
Correcting myself on two points in the comment above.

**"They never ran" was wrong.** I checked by restoring the old test file and running it in this branch's tree: all four tests ran, none skipped. What is true is narrower — they gate on `assumeTrue(systemModulesAvailable())`, so whether they run depends on build state rather than on anything the test controls. They do skip in some trees (a linked worktree, where the checkout walk looked for `.git` as a directory when it is a file there), and CI runs `distZip check`, so I cannot tell you from the outside whether `:javatools:test` fires before or after `lib_ecstasy`'s output exists. "Can silently skip depending on build state" is what I should have written; "never ran" overstated it.

**I removed a test I should have kept.** Trimming the class to what runs without a container also removed `ownerlessFactoryPathsStillWork`, which pins the two NPEs this PR introduced and then fixed - `xRTType.makeForeignHandle` and `Utils.translate`, both of which build a handle with no frame to take an owner from. Nothing else covers those paths. It is restored in b00ce2eab, so the class is now three tests: the two class-file scans plus that one, which builds a single container and makes no claim about how many a host may create.

Its repository helper is also no longer read-through. With read-through on, `LinkedRepository` answers a hit in a later repository by cloning the module into the first one - and the first here is a read-only build output directory, so the store fails and a module that *was* found comes back as null. That is worth knowing independently of this PR; it is a live way for a lookup to report "not found" for something it located.

The description is updated for the commit count, the diffstat and the test numbers, which had all gone stale.

lagergren:
This is still a problem, even with the Tooling API, and I would very much like to get this fixed in a single PR doing nothing else. I still rely on the interpreter to get LSP working.  Less importantly: I would really like a single implementation of the INSTANCE cache so that I can instrument it and use it for LSP footprint callbacks, so even unmodified it is really causing trouble for me that it exists copied into 146 places.  I will get back with an attempt to make this cleaner, but it is still going to wide but very shallow.

PR #575, opened 2026-09-01
Do not alias a caller's tuple storage as a callee's register file

lagergren:
A callee can overwrite its caller's tuple. This is reachable from ordinary
single-threaded Ecstasy code today, through reflection, and it writes through a
value the language considers immutable.

## The mechanism

`Utils.ensureSize(ahArg, cVars)` is **grow-only**. When the array it is given is
already big enough, it returns *that very array*:

```java
public static ObjectHandle[] ensureSize(ObjectHandle[] ahArg, int cVars) {
    int cArgs = ahArg.length;
    if (cArgs < cVars) {
        ObjectHandle[] ahVar = new ObjectHandle[cVars];
        System.arraycopy(ahArg, 0, ahVar, 0, cArgs);
        return ahVar;
    }

    return ahArg;      // <-- the caller's own array, handed straight back
}
```

Every tuple-argument call path then installs that result as the **callee frame's
register file** (`f_ahVar`). Put those two facts together:

> Whenever the callee needs no more registers than it has parameters, the
> caller's tuple storage *is* the callee's register file.

A frame's registers are where parameters live, and a parameter is an ordinary
assignable local. So a parameter reassignment in the callee is a store into the
caller's tuple.

The condition is not exotic — "needs no more registers than it has parameters"
is every small method that doesn't declare locals.

## Reproducing it

`manualTests/src/main/x/reflect.x`, `testInvokeTupleAliasing` — added by this PR:

```ecstasy
class Probe {
    void setB(String a, Int b) {
        a = "MUTATED";
        b = -1;
    }
}

Tuple<String, Int> args = ("origB", 7);
Method<Probe, <String, Int>, <>> m = Probe.setB;
m.invoke(p, args);

assert args[0] == "origB" && args[1] == 7;
```

On master:

```
** testInvokeTupleAliasing
args after invoke=(MUTATED, -1)
IllegalState: callee wrote through into the caller's tuple: (MUTATED, -1)
	at testInvokeTupleAliasing() (reflect.x:230)
	at run() (reflect.x:16)
```

With this PR it prints `args after invoke=(origB, 7)` and passes.

Note what the caller did wrong: nothing. It built a tuple and passed it. Tuples
are values; `args` is not declared mutable and the language offers no way to
write to it from there. The write arrives through the runtime, underneath the
type system.

## Why this is a master bug and not a theoretical one

Three independent signals, all already in master:

1. **`xRTFunction.invokeInvoke` already copies** for exactly this case
   (`cArgs == cVars ? ahArg.clone() : Utils.ensureSize(...)`). The hazard was
   known and handled on the function path.
2. **`xRTMethod.invokeInvoke` did not**, and carried the question in the source:
   `ObjectHandle[] ahPass = hTuple.m_ahValue;  // TODO GG+CP do we need to check these?`
   This PR answers that TODO. Yes.
3. The reflection path is **live today** — the reproducer above needs nothing but
   `Method.invoke`.

## Scope: one fix, twelve sites

The firing site is `xRTMethod.invokeInvoke`. The other eleven — `Call_T0/T1/TN/TT`,
`Invoke_T0/T1/TN/TT`, `Construct_T`, `New_T`, `NewG_T` — reach the identical
`ensureSize`-into-a-frame shape and carry the identical defect.

They are **latent, not firing**: the `InvocationExpression` / `NewExpression`
branches that would emit them are gated on an `m_fTupleArg` flag that is never
assigned anywhere, and `Construct_T` is currently an outright
`UnsupportedOperationException`. But they decode from any `.xtc`, and they go
live the moment that flag is wired up.

Fixing all twelve keeps **one rule for one hazard**, rather than leaving eight of
twelve to be rediscovered later by whoever turns that flag on. `Construct_T` is
the worst of the latent cases: it passes the array on to `Utils.makeFinalizer`,
so the caller's tuple storage would be retained by the finalizer well after the
constructor returned.

`xRTFunction` is deliberately left alone — its existing copy is correct and
predates this.

## The fix

One sanctioned way to take a tuple's elements toward a frame:

```java
public ObjectHandle[] valuesCopy() {
    ObjectHandle[] ahValue = m_ahValue;
    return Arrays.copyOf(ahValue, ahValue.length);
}
```

and every tuple-argument extraction routed through it.

**Cost**: one array copy of `cArgs` references per tuple-argument call. This is
the same copy `xRTFunction` has always made, and it is bounded by a method's
parameter count. On the paths where growth was needed, `ensureSize` was
allocating a new array anyway, so those are unchanged.

## Tests

- **`reflect.x testInvokeTupleAliasing`** — the end-to-end proof above, red on
  master, green here.
- **`MethodInvokeArgumentAliasingTest.ensureSizeAliasesWhenNoGrowthIsNeeded`** —
  pins the aliasing contract that makes the copies load-bearing. If `ensureSize`
  ever changes to always copy, this fails and the copies become redundant rather
  than wrong.
- **`MethodInvokeArgumentAliasingTest.noTupleArgumentSiteReadsTupleStorageDirectly`** —
  parses the *compiled* classes with `java.lang.classfile` and fails on a
  `getfield m_ahValue` in any of the twelve. Bytecode rather than source text, so
  it cannot be fooled by how the read is spelled, and it stays silent about the
  many legitimate uses of that field elsewhere. Red on master, naming all 31
  reads.

`:javatools:test` — 350 tests, 0 failures, 0 errors.

## Known limitation

`TupleHandle.m_ahValue` stays public, because plenty of code reads it for reasons
that never reach a frame. The bytecode ratchet is what keeps the twelve
frame-bound sites honest; it does not stop a *thirteenth* site being written. If
reviewers prefer, the field could be encapsulated in a follow-up, but that is a
wider change than this defect needs.


javatools/src/main/java/org/xvm/runtime/template/collections/xTuple.java
```diff
@@ -670,6 +670,21 @@ public static class TupleHandle
             extends ObjectHandle {
         public ObjectHandle[] m_ahValue;
 
+        /**
```
ggleyzer, line 673: Too verbose. All we need to say - "Clone the values array".

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTMethod.java
```diff
@@ -203,7 +203,12 @@ public int invokeBindTarget(Frame frame, MethodHandle hMethod, ObjectHandle hTar
     public int invokeInvoke(Frame frame, MethodHandle hMethod, ObjectHandle[] ahArg, int iReturn) {
         ObjectHandle   hTarget = ahArg[0];
         TupleHandle    hTuple  = (TupleHandle) ahArg[1];
-        ObjectHandle[] ahPass  = hTuple.m_ahValue;  // TODO GG+CP do we need to check these?
+        // the call chain grows the argument array into the callee frame's register file only when
```
ggleyzer: this comment is unnecessary

ggleyzer (approved):

javatools/src/main/java/org/xvm/runtime/template/collections/xTuple.java
lagergren: Done - cut down to `/** Clone the values array. */`.

javatools/src/main/java/org/xvm/runtime/template/_native/reflect/xRTMethod.java
lagergren: Done - removed.

PR #583, opened 2026-09-02
Fix a SharedContext propagataion bug

ggleyzer:
While running manualTests/service.x test on a virtual machine Cam was intermittently getting an exception:

Service "TestServices" (id=99) at ^TestServices (CallLaterRequest: native), fiber 354: Unhandled exception: IllegalState: "String user := UserId.hasValue()": UserId=(name=userId, defaultValue=Null, SharedContext.hash=5309077122121599543)
at TestService.validateContext() (services.x:144)
at ^TestService (validateContext())
    =========
at testSharedContext()

It was caused by a race in async SharedContext propagation.

This change contains a reproducer and the corresponding fix.


cpurdy (approved):
Wow! It's hard to believe that we never ran into this before. Looks good.

PR #584, opened 2026-09-03
Fix potential deadlock in XUnit when obtaining injectables during test execution

thegridman:
- Remove invoke methods from the ExecutionContext so that the context is not blocked during test invocation. Instead, the MethodExecutor is exposed as a property on the ExecutionContext.
- Annotate the EngineExecutionContext service with @Concurrent
- Fix the XUnit -t option to work for executing a single test method

### DIFF
--- a/lib_xunit_engine/src/main/x/xunit_engine/executor/BaseExecutionLifecycle.x
+++ b/lib_xunit_engine/src/main/x/xunit_engine/executor/BaseExecutionLifecycle.x
@@ -2,6 +2,8 @@ import extensions.ExtensionRegistry;
 
 import xunit.MethodOrFunction;
 
+import xunit.executor;
+
 /**
  * A base class for `ExecutionLifecycle` implementations.
  */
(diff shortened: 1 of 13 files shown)


lib_xunit_engine/src/main/x/xunit_engine/executor/BaseExecutionLifecycle.x
```diff
@@ -35,7 +35,7 @@ import xunit.MethodOrFunction;
 
         if (MethodOrFunction constructor := model.constructor.is(MethodOrFunction)) {
             // ToDo: call pre/post fixture constructor extensions
-            if (Object fixture := context.invokeSingleResult(constructor)) {
+            if (Object fixture := context.methodExecutor.invokeSingleResult(constructor, context)) {
```
lagergren: There are are a lot of "context.methodExecutor.*" stuff that has become longer than in the last branch, cosmetically, perhaps a getter or shorthand to do this ? 

Fixtures are Object? Should they at least be some kind of interface or is that the whole point that they can be anything or that the invoke return doesnt allow you a better return type?  That could be a different PR though , if relevant.

lagergren (approved):
Looks right to me and I understand the underlying reason for the change.

lib_xunit_engine/src/main/x/xunit_engine/executor/BaseExecutionLifecycle.x
thegridman: The "fixture" is the instance of the thing that contains the test method or function being invoked, so it could be anything, a class, const, service, module, package, etc.

Regarding the method executor change, I'm not 100% sure I like it either. The MethodExecutor is a property on the context. I guess it does not have to be. In fact MethodExecutor is just a const with some method that know how to invoke test methods and bid parameters to those methods. In fact it could just be a number of static methods. I'll look at that.

PR #585, opened 2026-09-03
Fix StackOverflowError compiling an assignable synthetic expression, and run Dec28 as a manual test

lagergren:
`Dec28.x` was excluded from the manualTests source set because compiling it killed the compiler with a `StackOverflowError`. This fixes the compiler defect, puts the file back, and turns it into a test that can actually fail.

## The defect

`Expression` implements the L-value generation methods as a pair of bridges, so a subclass only has to write one of them:

```java
public Assignable generateAssignable(...) {
    if (hasMultiValueImpl()) {
        return generateAssignables(ctx, code, errs)[0];   // -> plural
    }
    throw notImplemented();
}

public Assignable[] generateAssignables(...) {
    if (hasSingleValueImpl() && isSingle()) {
        return new Assignable[] { generateAssignable(ctx, code, errs) };  // -> singular
    }
    ...
}
```

Both are documented as "must be overridden by any expression that is assignable". A class that reports **both** implementations, is single, is assignable, and overrides **neither** makes those two call each other until the stack is exhausted.

`SyntheticExpression` is a delegating wrapper. It forwarded every assignability *question* to the wrapped expression - `isAssignable`, `requireAssignable`, `markAssignment` - but neither of the two methods that do the assigning, so it inherited both bridges. Overriding `isAssignable` is precisely what moved it into the position where the mutual recursion is reachable.

The fix delegates the two generation methods the same way the class delegates everything else.

## Why it was not noticed

`SyntheticExpression` is synthesized by the compiler, never written by hand: `ConvertExpression` (inserted by `Expression.convertTo` when a value needs a conversion method) and `UnpackExpression` (inserted when a tuple is unpacked into multiple values). Reaching the bug needs one of these wrappers to be the *target of an assignment*, which is uncommon - `Dec28.x` was the only file in the repository that produced it, and it was excluded, so nothing exercised the path.

The failure mode was a `StackOverflowError` out of the compiler rather than a diagnostic.

## Test

`Dec28.x` comes off the exclude list and is added to `testModuleNames` - one entry, the same way the other modules are registered - so it runs in `runSequential` and therefore in CI.

`run()` previously only printed its results, so a wrong rendering would have scrolled past unnoticed. It now pairs each literal with the rendering it must produce and asserts. Verified to have teeth by corrupting one expectation:

```
IllegalState: Dec28(".00001234500") rendered "0.000012345", expected "0.00001234X"
```

which fails the task with a non-zero exit.

The debug assert at the top of `run()` is commented out. It is a link-time no-op unless the `debug` condition is defined - it compiles to a jump over the assert - so it was harmless while nobody ran this module, but now that CI does, an enabled break-point here would hang the run.

## Verification

- `Dec28.xtc` compiles (it did not before).
- `:manualTests:runOne -PtestName=TestDec28` passes; all eight literals round-trip.
- Negative check above fails the build as it should.


lagergren:
This would have been a compile time error if you had used sealed classes. 

Assume that SyntheticExpression is sealed with implicit "permits ConvertExpression, UnaryOpExpression, UnpackExpression". The Java type hierarchy is closed at compile time and it's a compile error. One exhaustiveness test on permitted subclasses would have found it mechanically. 

I feel there are also a several things pointing to that you should experimentally seal AstNodes and ConstantPool entries and see what falls out. I think some things are hiding undetectable here, since you do all your type checking at runtime, so in order to fully utilize it, there would probably be rewrites to use -> switches with types rather than runtime casts and instanceofs on what at compile time can only be known as "Objects".  Might be a good direction to go in. Again - it is hard to go back to types from Objects, and it won't be a totally free bonus, but definitely worth investigating.

javatools/src/main/java/org/xvm/compiler/ast/SyntheticExpression.java
```diff
@@ -71,6 +72,26 @@ public void requireAssignable(Context ctx, ErrorListener errs) {
         expr.requireAssignable(ctx, errs);
     }
 
+    /**
```
ggleyzer: This is a pattern Cam came up with about ten years ago and we use it all over the place. It was clearly an omission in this class, but I don't think the comment belong here.

manualTests/src/main/x/Dec28.x
```diff
@@ -2,12 +2,20 @@ module TestDec28 {
     @Inject Console console;
 
     void run() {
-        assert:debug;
-
-        for (String s : ["0", "1", "123", "123.45", "1.234567", "0.0001234567",
-                         ".00001234500", "1234000"]) {
-            Dec28 dec = new Dec28(s);
-            console.print($"s={s}, dec={dec}");
+        String[] literals = ["0", "1", "123", "123.45", "1.234567", "0.0001234567",
```
ggleyzer: Why don't we move it to numbers.x? 

Also, we are trying to get away from console.print() as the means of verification as much as possible. If there any way to replace it with as assert - it would be much preferable

ggleyzer (changes requested):

javatools/src/main/java/org/xvm/compiler/ast/SyntheticExpression.java
lagergren: Removed. Agreed - the pattern is used throughout, this class just missed it, so the two overrides now read like the `requireAssignable`/`markAssignment` delegations right above them.

manualTests/src/main/x/Dec28.x
lagergren: Both done.

Moved into `numbers.x` as `testDec28()`. That is strictly better than what I had: `TestNumbers` is already in `testModuleNames`, so there is no build wiring at all now - `build.gradle.kts` only loses the dead exclude for a file that no longer exists, and `Dec28.x` is gone.

Verification is by assert. Each literal is paired with the rendering it must produce; the only `console.print` left is the `** testDec28()` section header, to match the other tests in the file. Confirmed it actually fails by corrupting one expectation:

```
IllegalState: Dec28(".00001234500") rendered "0.000012345", expected "0.00001234X"
```

non-zero exit, failed task.

ggleyzer (approved):

PR #586, opened 2026-09-03
Fix a concurrency issue where JsonMapStoreTest fails to load off heap  values owned by pending commits

thegridman:
There was a concurrency issue in JsonMapStore where the store could not load a value marked as OffHeap that was owned by a transaction not yet stored on disc.

### DIFF
--- a/manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
+++ b/manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
@@ -674,7 +674,7 @@ class JsonMapStoreTest {
                             schema.people.put(idx, new Person($"First{workerId}", Null, $"Last{i}"));
                         }
                     }
-                } catch (Exception e) {
+                } catch (Exception _) {
                     // Concurrent writers hitting the same keys will legitimately conflict
                     // sometimes (e.g. oodb:CommitFailed/ConcurrentConflict) -- that is correct
                     // OODB behavior, not the bug being probed for.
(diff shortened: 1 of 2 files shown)


manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
```diff
@@ -592,4 +592,100 @@ class JsonMapStoreTest {
...
+                            schema.people.remove(idx);
+                        } else {
+                            schema.mapData.put(stringKey, $"v-{workerId}-{i}");
+                            schema.people.put(idx, new Person($"First{workerId}", Null, $"Last{i}"));
+                        }
+                    }
+                } catch (Exception e) {
+                    // Concurrent writers hitting the same keys will legitimately conflict
+                    // sometimes (e.g. oodb:CommitFailed/ConcurrentConflict) -- that is correct
+                    // OODB behavior, not the bug being probed for.
```
lagergren, line 680: Would it still be worth logging this as a warning? I have had some bad experiences with swallowed exceptions that suddenly start swallowing them for a different reason rendered undetectable. 

Is it possible to use _ instead of e in Exctasy same as in java/kotlin/python for "ignored" otherwise?

lagergren (approved):

manualTests/src/main/x/jsondb/jsondb_test/storage/JsonMapStoreTest.x
ggleyzer: yes, we do support _ as a "ignored"; LSP barfs on it sometimes though

ggleyzer: we could count the number of ConcurrentConflict vs CommitFailed and report at the end

thegridman: This was Michael's original test that he put in issue #541 that we copied verbatim into this PR as it highlighted the concurrency issue. I will change `} catch (Exception e) {` to `} catch (Exception _) {`

PR #593, opened 2026-09-10
Enforce source whitespace with Spotless and fix existing files

lagergren:
# Summary

## Why?

- Java and XTC sources need consistent whitespace rules that are enforced independently of editor settings.
- Developers should get automatic formatting fixes locally, while CI rejects violations in committed files.

## How?

- Configure Spotless to remove trailing whitespace, expand indentation tabs, collapse consecutive blank lines, and ensure a final newline across all Java/XTC sources.
- Make local root `check` and `build` apply fixes before verification; retain read-only verification in CI and for explicit `spotlessCheck` invocations.
- Keep the machinery and the cleanup of 39 source files in separate commits for review.

## Implications

- Local `check` and `build` can modify source files; review and commit the resulting diff.
- Existing LF or CRLF conventions are retained, and generated output and hidden directories, including local worktrees, are excluded.
- Tabs remaining outside indentation require manual repair, so literal string contents are not automatically replaced with spaces.
- Javac warning enforcement is outside this change.

## Changes

- `build.gradle.kts` — centralize Spotless rules and local-versus-CI task ordering.
- `gradle/libs.versions.toml` — add Spotless 8.10.2.
- `.github/workflows/commit.yml` — clarify that the existing root `check` invocation enforces Spotless without applying fixes in CI.

## Manual verification

1. Run `./gradlew spotlessCheck`; all 2,152 Java/XTC source files pass.
2. Add trailing spaces or remove a final newline from a source file. `./gradlew spotlessCheck` fails without editing it; `./gradlew spotlessApply` repairs it. Local root `check` and `build` also apply those repairs automatically.
3. With the same violation, `CI=true ./gradlew spotlessCheck --no-scan` fails without changing the file.

Validation exercised temporary fixtures for each rule, LF/CRLF handling, empty files, exclusions, new-source detection, idempotence, and configuration-cache reuse. Root `check` was exercised with unrelated included-build checks detached: local execution repaired the fixture, while CI execution failed without editing it. Local and CI `build` task graphs were also verified. The full XVM build/test suite was not run. All 39 source edits preserve non-whitespace content.


ggleyzer:
Two comments:

1) I don't think there is any reason to change anything in "archive/" directory; that's why it's called "archive" :-)
2) I really don't like introducing yet another dependency for something that trivial; either Claude or Chet would do it in a minute

ggleyzer (approved):

Issue #603, opened 2026-09-16
Floating-point defects in the reference compiler and interpreter


lagergren:
An audit of the FP codecs, prompted by finding that every non-zero `Float8e4`/`Float8e5` literal compiled to `+0.0`, turned up several more defects of the same family. Some are fixed in #598; the rest are recorded here.

The recurring shape is worth naming, because four separate codecs had a version of it: **a hand-rolled conversion, verified (if at all) by a round-trip test — which cannot see the bug, because an exactly representable value needs no rounding.** The interesting inputs are the ones *between* representable values, and nothing was checking those.

## Fixed in #598

Listed for context; each has an exhaustive unit test.

| | defect |
|---|---|
| `Float8e4Constant` / `Float8e5Constant` | `toBits` returned `0x00` for every non-zero finite value; `toFloat` threw for all but four patterns |
| all four FP8/FP16 codecs | `floatToIntBits` collapses NaN onto the canonical positive one, so a negative NaN lost its sign |
| `Float16Constant.toFloat` | a stray `\| 0x3ff` made every half with a zero significand — every power of two, `1.0` included — decode 1023 ULPs high (58 of 65536 patterns) |
| `BFloat16Constant.toHalf` | "rounded" by multiplying the value by `1.001957f`; 12.5% of values needing rounding landed on the wrong neighbour |
| `xFloat8e4`/`xFloat8e5`/`xFloat16` | interpreter arithmetic did not round results back into the format, so values escaped their own type |
| `BaseBinaryFP` `split()` | `(l & EXP_MASK >>> 52)` parses as `l & (EXP_MASK >>> 52)`, so `split()` reported exponent `0` for every `Float64`, and with it `FPNumber.exponent` |

### 1. BFloat16 cannot execute at all

```
BFloat16 a = 1.0;
→ Unknown constant: BFloat16{value=1.0}
```

There is no `xBFloat16` template in `runtime/template/numbers`, so no BFloat16 value can be created. This is the same state `Float8e4`/`Float8e5` were in before #598: a constant encoder with no runtime behind it.

`manualTests/numbers.x` carries a `testBFloat16()` with the assertions ready, deliberately not called from `run()`, with a TODO naming this.

### 2. `split()` is still wrong for BFloat16, Float128 and FloatN

[...]

PR #604, opened 2026-09-16
Fix four FP math defects: cos() unimplemented, atanh() computing arcoth, scaleByPow() raising to a power, log2() inexact

lagergren:
Found while auditing what else is unimplemented in the compiler and interpreter, following the floating-point work in #598. Both defects are in the FP maths shared by every numeric type, binary and decimal.

## `cos()` threw for every floating point type

```
Float64 x = 1.0;
x.sin()  ->  0.8414709848078965
x.tan()  ->  1.5574077246549023
x.cos()  ->  NotImplemented
```

It was missing at all three levels:

- `BaseFP.initNative` never passed `"cos"` to `markNativeMethod`, so the call fell through to the `TODO` body in the Ecstasy source
- neither `BaseBinaryFP` nor `BaseDecFP` had a `"cos"` case to dispatch to
- `Decimal` had no `cos()` at all

The marking list does have `"asinh"` **twice**, immediately after `"tanh"`, which is where `"cos"` looks to have gone:

```java
markNativeMethod("tanh"       , VOID, THIS);
markNativeMethod("asinh"      , VOID, THIS);   // <- already marked six lines above
markNativeMethod("acosh"      , VOID, THIS);
```

The duplicate is removed.

## `atanh()` returned NaN for every valid input

It computed

```java
0.5 * Math.log((d + 1.0) / (d - 1.0))
```

which is **arcoth**, not artanh. Over atanh's own domain of `|x| < 1` the argument to the logarithm is negative, so the result was always NaN — and in the decimal path `BigDecimal.valueOf(NaN)` throws, so it surfaced as a raw Java exception:

```
Run-time error: java.lang.NumberFormatException: Infinite or NaN
```

Now `0.5 * log((1 + d) / (1 - d))`, fixed in both `BaseBinaryFP` and `Decimal`:

```
Float64 0.5 .atanh()    before  NaN    now  0.5493061443340548
```

## One helper for the double-based math

`Decimal`'s maths methods were all the same six lines with one word changed. Nineteen of them now share:

```java
private Decimal applyToDouble(DoubleUnaryOperator op, Decimal decNonFinite) {
    return isFinite()
            ? fromBigDecimal(BigDecimal.valueOf(op.applyAsDouble(toBigDecimal().doubleValue())))
            : decNonFinite;
}
```

so `sin()` is `return applyToDouble(Math::sin);`.

The non-finite result is a parameter because it genuinely differs per operation — `nan()` for the trigonometric ones, `this` for `deg2rad`/`nextUp`, `isSigned() ? nan() : this` for the logarithms, `isSigned() ? zero(false) : this` for `exp`. That behaviour is preserved exactly:

```
inf:   sin=NaN  deg2rad=Infinity  log=Infinity  exp=Infinity
-inf:  log=NaN  exp=0             deg2rad=-Infinity
```

**Not folded in**, deliberately: `abs`, `floor`, `ceil` and `sqrt` work on the `BigDecimal` rather than on a double.

## `scaleByPow()` raised the value to a power instead of scaling it

Found by the coverage below, immediately after writing it.

```java
case "scaleByPow":
    return frame.assignValue(iReturn, makeHandle(Math.pow(d, l)));
```

`FPNumber` documents it as *"this value scaled by `radix` raised to the power of `n`"*, and describes it as analogous to shifting the decimal point. So `2.0.scaleByPow(3)` should be `2 × 2³ = 16.0`. It gave `2³ = 8.0` — the name appears to have been read as "pow" rather than as "scale by". The decimal path had the same defect via `dec.pow(n)`.

Binary FP now uses `Math.scalb`, which is exactly "multiply by 2ⁿ". `Decimal` gains `scaleByPow`, built on `BigDecimal.scaleByPowerOfTen`, the radix-10 equivalent. The tests pin **both** radices, since that difference is the entire point of the operation:

```
Float64 2.0 .scaleByPow(3)  == 16.0     (radix 2)
Dec64   2.0 .scaleByPow(3)  == 2000.0   (radix 10)
```

## `log2()` was not exact for powers of two

The `new BigDecimal(double)` inconsistency noted above turned out to sit on top of a real defect. `log2` was computed as `log10(x) × (1/log10(2))` — exact in real arithmetic, but not in floating point:

```
Float64 8.0 .log2()     before  2.9999999999999996   now  3.0
Float64 2^-14 .log2()   before -13.999999999999998   now -14.0
```

**13 of the 41** exact powers of two between 2⁻²⁰ and 2²⁰ came back as non-integers. Dividing the natural logarithms is exact for all 41, and is the better approximation in general too — over 200k random finite doubles, comparing which candidate reproduces `x` more closely under `2^result`, `log(x)/log(2)` wins 59,425 to 47,828 with 86,904 identical.

`LOG2_10` existed in two copies, in `BaseFP` and in `Decimal`, used for nothing but this; both become `LOG_2`.

And with `new BigDecimal(double)` gone — it was arriving as `2.999999999999999555910790149937383830547332763671875`, fifty digits of noise to be rounded away — `log2` joins the other nineteen methods on the helper.

**The test now asserts exactness, not closeness.** The previous assertion compared within `1e-9` and passed on the wrong answer, which is exactly how this survived being tested.

## Why these sat here

`numbers.x` made **one** transcendental call in total. That is how `cos()` came to throw for every floating point type, and `atanh()` came to compute arcoth, without anything noticing.

`testFPMathSurface` now checks all 28 functions `FPNumber` declares against known values: the trigonometric functions and their inverses, the hyperbolics and theirs, `exp` and the three logarithms, `sqrt` and `cbrt`, the angle conversions round-tripping through each other, `floor`/`ceil`/`round`/`round(TowardZero)`, `nextUp`/`nextDown` round-tripping, `pow` including its agreement with `sqrt` at exponent ½, and `atan2`.

Four defects in 28 functions, in code that had effectively no coverage.

## Tests

`manualTests/numbers.x` gains `testFPMath`: `cos` on Float64, Float32, Float16, Dec64 and Dec32; the `sin² + cos² = 1` identity; `atanh` over its domain and that `tanh` undoes it; plus `asinh` and `acosh`, which share `atanh`'s shape and were correct.

The Float16 case compares through Float64 on purpose. A Float16 literal whose significand is zero still decodes 1023 ULPs high on master, so `f16.cos() == 1.0` would compare a correctly computed `1.0` against a literal that is not quite `1.0` — the test fails for a reason that has nothing to do with `cos`. That decoder defect is fixed in #598; this branch stays independent rather than carrying the same fix twice, and the comparison can be simplified once #598 lands.

## Verification

- javatools + javatools_utils: 40 suites, 469 tests, 0 failures, 42 skipped
- `manualTests runSequential`: 21/21 SUCCESS

Related: #603 catalogues the remaining floating-point gaps.


javatools/src/main/java/org/xvm/runtime/template/numbers/BaseDecFP.java
```diff
@@ -138,7 +138,10 @@ public int invokeNative1(Frame frame, MethodStructure method, ObjectHandle hTarg
             Decimal dec  = ((DecimalHandle) hTarget).getValue();
             long    lPow = ((JavaLong) hArg).getValue();
 
-            return frame.assignValue(iReturn, makeHandle(dec.pow((int) lPow)));
+            // scale by the radix raised to n -- 10^n for a decimal FP type -- rather than
```
ggleyzer, line 141: As I mentioned many times, I would prefer avoiding references to the replaced (incorrect) logic in the comments; the reader would lack the context. Explanation of what the code does is sufficient. The critique of the old code belongs on the PR, not the code

javatools/src/main/java/org/xvm/runtime/template/numbers/BaseBinaryFP.java
```diff
@@ -120,7 +120,12 @@ public int invokeNative1(Frame frame, MethodStructure method, ObjectHandle hTarg
             double d = ((FloatHandle) hTarget).getValue();
             long   l = ((JavaLong) hArg).getValue();
 
-            return frame.assignValue(iReturn, makeHandle(Math.pow(d, l)));
+            // scale by the radix raised to n -- for a binary FP type that is 2^n, which is
+            // exactly what scalb does. This used to be Math.pow(d, l), raising the value to the
```
ggleyzer, line 124: ditto

manualTests/src/main/x/numbers.x
```diff
@@ -399,6 +400,157 @@ module TestNumbers {
...
+
+        Dec64 hd = 0.5;
+        assert (hd.atanh().toFloat64() - 0.5493061443340549).abs() < 0.000000001;
+
+        // the neighbouring inverse hyperbolics, which share the shape and were correct
+        assert (h.asinh() - 0.48121182505960347).abs() < 0.000000001;
+        Float64 two = 2.0;
+        assert (two.acosh() - 1.3169578969248166).abs() < 0.000000001;
+
+        console.print($"cos(0)={zero.cos()} atanh(0.5)={a} acosh(2)={two.acosh()}");
```
ggleyzer, line 454: I would prefer avoiding adding new console output and find a way to "assert" instead

ggleyzer (changes requested):

javatools/src/main/java/org/xvm/runtime/template/numbers/BaseDecFP.java
lagergren: Yeah sorry it's something I do for my own note taking when trying to fix a bug and reason about the code. I usually try to get rid of them but I am far from successful.

lagergren: Fixed in 7278c4bbc4be10fd233396bddf1e2c229021eb72. The comment now describes scaling by 10^n. I also removed the historical comparisons elsewhere in this PR's FP comments and tests, along with the stale LOG2_10 documentation.

javatools/src/main/java/org/xvm/runtime/template/numbers/BaseBinaryFP.java
lagergren: Fixed in 7278c4bbc4be10fd233396bddf1e2c229021eb72. The comment now describes scaling by 2^n, with the discussion of the previous implementation removed.

manualTests/src/main/x/numbers.x
lagergren: Fixed in 7278c4bbc4be10fd233396bddf1e2c229021eb72. Removed all three console.print calls added by the FP math tests. The printed values already had assertions; all 53 FP math assertions are preserved. TestNumbers passed with --rerun-tasks --no-build-cache, and spotlessCheck passed.

ggleyzer (approved):

PR #605, opened 2026-09-17
Support Float8e4/Float8e5 in the JIT, and keep Float16 results in format

lagergren:
Adds `Float8e4`/`Float8e5` to the JIT back-end, and fixes `Float16` results
escaping their own format. This is the JIT half of #598 (which did the
constant encoding, the `Float16` decoder, and the interpreter templates);
that has now landed, so this branch is rebased to sit directly on master
and contains only the JIT-side work.

### FP8 in the JIT

An FP8 value is carried as its 8-bit encoding in an `int`, so it shares the
`"I"` carrier with the small integer types while sharing none of their
arithmetic or ordering. Every primitive operation therefore has to ask the
Ecstasy type rather than the carrier before dispatching:

- `NumberSupport` routes FP8 add/sub/mul/div/mod/rem to `$add`/`$div`/... on
  the jitbridge class instead of the integer opcodes.
- `TypeConstant.buildCompare()` routes FP8 comparison through `$compare`.
  The encoding is sign-magnitude, so comparing it as an `int` is wrong:
  `-4.0` encodes as `0xC8` and `4.0` as `0x48`.
- `Builder` loads FP8 constants as their encoding, and gains the field
  carrier, `$box` and `ClassDesc` entries; `JitTypeDesc` gains the `JIT_LIST`
  entry; `ConstantPool` gains `typeFloat8e4()`/`typeFloat8e5()` and lists
  both as JIT primitives.
- The `Float8e4`/`Float8e5` jitbridge classes get the actual codec and
  arithmetic helpers.

### Float16 narrowing

`Float16` shares the `"F"` carrier with `Float32`, so an operation on it was
performed at float precision and the result kept whatever float precision
produced — a `Float16` register could hold a value `Float16` cannot
represent. Binary FP ops now round back into the format via
`floatToFloat16`/`float16ToFloat`. `Float32` and `Float64` fill their
carriers exactly and are untouched.

### Tests

New `Float8e4Tests`, `Float8e5Tests` and `Float16Tests` under
`manualTests/src/main/x/jit/jit_tests/numbers/` — 197 asserts covering
comparison, fields, nullable fields, const fields and constructors, params,
returns, conditional returns, arrays, rounding, negation and arithmetic.

`xtc run --jit src/main/x/jit/jit_tests.x` reports **All JIT tests passed**,
and `:javatools:test` is 392 tests / 0 failures / 0 errors.

Since the JIT runners print nothing on success, I sanity-checked that the
new cases genuinely execute rather than silently passing: temporarily
breaking one `Float8e4` assert makes the suite fail at that exact frame and
report `One or more JIT tests failed`.


javatools/src/main/java/org/xvm/asm/constants/TerminalTypeConstant.java
```diff
@@ -79,7 +79,7 @@ public TerminalTypeConstant(ConstantPool pool, Constant constId) {
      *
      * @throws IOException if an issue occurs reading the Constant value
      */
-    public TerminalTypeConstant(ConstantPool pool, Format format, DataInput in)
+    public TerminalTypeConstant(ConstantPool pool, @SuppressWarnings("unused") Format format, DataInput in)
```
ggleyzer: I don't see any reason to clobber the code with `@SuppressWarnings("unused")`. 

We have hundreds of places where some arguments are not used (by design). Fixing juts one seems arbitrary and unnecessary.

javatools/src/main/java/org/xvm/javajit/JitTypeDesc.java
```diff
@@ -55,11 +55,13 @@ public static ClassDesc getJitClass(Builder builder, TypeConstant type) {
      * @return the primitive ClassDesc if the specified type is optimizable to a primitive Java
      *         class; null otherwise
      */
-    public static ClassDesc getJavaPrimitive(TypeConstant type) {
+    public static ClassDesc findJavaPrimitive(TypeConstant type) {
```
ggleyzer: I don't see why this method is better be called "find" rather than "get". It doesn't search for anything - simply returns a one-to-one mapping

javatools/src/main/java/org/xvm/javajit/builders/CommonBuilder.java
```diff
@@ -124,7 +124,7 @@ public CommonBuilder(TypeSystem typeSystem, Artifact art) {
 
     /**
      * List of constant properties for every class name this builder assembles.
-     *
+     * <p>
```
ggleyzer: Similar to my other comment, I don't see a reason to fix just one place out of hundreds where we don't format the docs "correctly". We don't have any plans to produce a proper Javadoc anyway... This simply clobbers the diff and adds no value IMHO

javatools/src/main/java/org/xvm/javajit/builders/EnumerationBuilder.java
```diff
@@ -26,7 +26,7 @@
 
 /**
  * The builder for Enumeration types.
- *
+ * <p>
```
ggleyzer: ditto

javatools/src/main/java/org/xvm/javajit/builders/EnumerationBuilder.java
```diff
@@ -91,17 +91,13 @@ private void assembleCountProp(ClassBuilder classBuilder) {
         JitMethodDesc jmDesc     = prop.getGetterJitDesc(this);
 
         classBuilder.withMethodBody(getterName + OPT, jmDesc.optimizedMD, ClassFile.ACC_PUBLIC,
-                code -> {
-            code.loadConstant((long) enumValues.length)
-                .lreturn();
-        });
+                code -> code.loadConstant((long) enumValues.length)
+                    .lreturn());
```
ggleyzer: maybe add a tab to align the ops (the way it used to be written)?

javatools/src/main/java/org/xvm/javajit/builders/EnumerationBuilder.java
```diff
@@ -91,17 +91,13 @@ private void assembleCountProp(ClassBuilder classBuilder) {
...
 
         classBuilder.withMethodBody(getterName, jmDesc.standardMD, ClassFile.ACC_PUBLIC,
-                code -> {
-            code.loadConstant((long) enumValues.length)
-                .invokestatic(CD_Int64, "$box", MethodTypeDesc.of(CD_Int64, CD_long))
-                .areturn();
-        });
+                code -> code.loadConstant((long) enumValues.length)
+                    .invokestatic(CD_Int64, "$box", MethodTypeDesc.of(CD_Int64, CD_long))
+                    .areturn());
```
ggleyzer: ditto

javatools/src/main/java/org/xvm/javajit/builders/EnumValueBuilder.java
```diff
@@ -16,7 +16,7 @@
 
 /**
  * The builder for Enum value types.
- *
+ * <p>
```
ggleyzer: ditto

javatools/src/main/java/org/xvm/javajit/builders/ExceptionBuilder.java
```diff
@@ -75,7 +73,7 @@ public void assembleCreateException(ClassBuilder classBuilder) {
      *     }
      *</pre></code>
      */
-    public void assembleJavaException(String className, ClassBuilder classBuilder) {
+    public void assembleJavaException(@SuppressWarnings("unused") String className, ClassBuilder classBuilder) {
```
ggleyzer: ditto

ggleyzer (changes requested):
Reviewing meaningful changes along with dozens of tiny formatting trivial ones makes the process very time consuming. I'd suggest to simplify the process by splitting in two. 1) The meaningful functionality changes
2) A separate "reformat code" change.

I don't mind fixing a typo here and there when you work on a class, but changing them randomly just due to one's personal preference seems to be a questionable process

javatools/src/main/java/org/xvm/asm/constants/TerminalTypeConstant.java
lagergren: Agreed, reverted — see #610.

34 of the 66 `Constant` subclasses taking `(pool, Format, DataInput)` ignore that parameter: `disassemble()` calls them all through one shape and their format is implied by their class. Annotating one was arbitrary.

It cannot be dropped or renamed to `_` either — JEP 456 excludes method and constructor parameters, which is why `catch (IOException _)` is fine elsewhere in this codebase but this is not. Left exactly as master has it.

javatools/src/main/java/org/xvm/javajit/JitTypeDesc.java
lagergren: Fair — "find" implied a search where there is only a one-to-one mapping. Renamed in #610.

`getJavaPrimitive` and `getPrimitiveFieldClass` keep their original names and their nullable contracts, so their existing callers are untouched, and the non-null variants are `requireJavaPrimitive` / `requirePrimitiveFieldClass`, named alongside `Objects.requireNonNull`.

The reason for two methods at all is that most callers had already ruled the null out with `isJavaPrimitive()` and were carrying `assert cd != null` to say so.

javatools/src/main/java/org/xvm/javajit/builders/CommonBuilder.java
lagergren: Agreed, reverted in #610 — 29 of them across eight files. Changing comment formatting in whatever files a branch happens to touch just hides the real diff.

javatools/src/main/java/org/xvm/javajit/builders/EnumerationBuilder.java
lagergren: Same — reverted in #610.

javatools/src/main/java/org/xvm/javajit/builders/EnumerationBuilder.java
lagergren: Done in #610 — the chained ops line up under `code.` again, as they did before the lambda bodies were flattened.

javatools/src/main/java/org/xvm/javajit/builders/EnumerationBuilder.java
lagergren: Same — realigned in #610, all three lambdas in this file.

javatools/src/main/java/org/xvm/javajit/builders/EnumValueBuilder.java
lagergren: Same — reverted in #610.

javatools/src/main/java/org/xvm/javajit/builders/ExceptionBuilder.java
lagergren: Same — reverted in #610. The `@SuppressWarnings("unused")` on `assembleJavaException`'s `className` is gone too.

lagergren:
Done — the PR is split. This one is now only the Float8e4/Float8e5 JIT support and its tests: 10 commits, 19 files. Everything else moved to #610, which targets this branch so it can be reviewed separately once this lands.

Two test fixes were kept here rather than moved, on the grounds that leaving broken tests in place is worse than a clean split:

- `LiteralFormatPlumbingTest` had four tests that read `.java` files as text and asserted on substrings — including the literal string `"protected void disassemble(DataInput in)"`. They assert formatting rather than behaviour. Replaced with a real write and read-back round trip; removing `TimeZone` from the disassemble switch now fails it with `Unsupported constant format: 33`.
- `DirRepositoryConcurrentScanTest` copied `.xtc` files out of the XDK build and `assumeTrue`d past their absence, so it reported green having run nothing on a checkout without the distribution built. It writes its own modules now, and runs in 0.157s.

All eight inline comments are answered and resolved; each fix is on #610 and I verified it there before resolving.

ggleyzer (approved):
Looks good

PR #608, opened 2026-09-17
ListMap: isolate copied entries, drop the raw casts, and add tests (#607)

### LINKED ISSUE #607: ListMap: copy constructor shares mutable entries, plus three soundness issues and no tests
`javatools_utils/src/main/java/org/xvm/util/ListMap.java` has one real correctness bug and three
soundness/clarity problems. It has no tests at all.

## 1. The copy constructor shares mutable entries with the original (correctness bug)

```java
public ListMap(ListMap<K, V> map) {
    m_list = new ArrayList<>(map.m_list);   // line 42 - copies references to the SimpleEntry objects
}
```

`ArrayList(Collection)` copies element *references*, and the elements are mutable
`AbstractMap.SimpleEntry` objects. `put` on a key that already exists goes through
`entry.setValue(value)` rather than adding a new entry:

```java
public V put(K key, V value) {
    Entry<K,V> entry = getEntry(key);
    if (entry != null) {
        return entry.setValue(value);       // mutates an entry the original map still holds
    }
    ...
```

So writing an existing key in a copy silently writes it in the original. Confirmed with:

```java
var original = new ListMap<String, String>();
original.put("k", "original");

var copy = new ListMap<>(original);
copy.put("k", "changed");

assertEquals("original", original.get("k"));
// org.opentest4j.AssertionFailedError:
//   expected: <original> but was: <changed>
```

Every caller of this constructor is making a defensive copy, and none of them get one:

| Call site | Intent |
| --- | --- |
| `ClassStructure.java:551` | hands a copy of `m_mapParams` to callers |
| `ClassStructure.java:3331` | `markReadOnly()` - copies in order to freeze |
| `Component.java:2688` | clones a Contribution's params from `that` |
| `Component.java:3040` | `markReadOnly()` again |
| `DifferenceTypeConstant.java:382` | merges child maps |
| `TypeConstant.java:5701` | derives a child map from another TypeInfo |

`Component.java:3037` is the clearest illustration - the line immediately above copies a list
correctly and then the map copy shares state:

```java
m_listInject = Collections.unmodifiableList(new ArrayList<>(m_listInject));   // isolated
if (m_mapParams != null) {
    m_mapParams = new ListMap<>(m_mapParams);                                 // not isolated
}
```

A `markReadOnly()` that produces a map whose entries stay writable through the old reference is the
inverse of its purpose.

Whether this bites today depends on whether anything writes an *already present* key into one of
those copies; I have not traced that. The isolation those call sites are asking for is not there
either way.

Fix: copy the entries, not the list.

```java
public ListMap(ListMap<K, V> map) {
    m_list = new ArrayList<>(map.size());
    for (Entry<K, V> entry : map.m_list) {
        m_list.add(new SimpleEntry<>(entry.getKey(), entry.getValue()));
    }
}
```

## 2. `asList()` is immutable only when assertions are enabled

```java
public List<Entry<K,V>> asList() {
    List<Entry<K,V>> list = (List) m_list;
    assert (list = Collections.unmodifiableList(list)) != null;   // line 83 - assignment in an assert
    return list;
}
```

The wrapper is applied inside the assert expression, so `asList()` returns an **unmodifiable** view
under `-ea` and a **live, mutable** view of the internal list in production. The javadoc says
"read-only". Two behaviours from one method depending on a JVM flag is a bad place to be, and the
production one exposes the map's internals.

Both callers - `Component.java:3209` and `NameExpression.java:1672` - only read, so making it
unconditionally unmodifiable is safe.

## 3. Two raw casts, from one avoidable cause

```java
List<Entry<K,V>> list = (List) m_list;              // line 82
return (Iterator) m_list.iterator();                // line 136
```

Both exist because the field is declared as the implementation type while the API speaks the
interface type, and generics are invariant:

```java
private final ArrayList<SimpleEntry<K, V>> m_list;  // line 127
```

Declaring it `ArrayList<Entry<K, V>>` removes the need for both casts entirely - `put` still adds
`new SimpleEntry<>(...)`, since `SimpleEntry` is an `Entry`. The only signature that changes is
`protected SimpleEntry<K,V> getEntry(Object)` (line 105), which becomes `Entry<K,V>`; there are no
subclasses of `ListMap` in the repo, so that is free.

## 4. Immutability is signalled by array identity

```java
public ListMap(int cInitSize) {
    m_list = cInitSize >= 0
        ? new ArrayList<>(cInitSize)
        : (ArrayList<SimpleEntry<K,V>>) EMPTY_ARRAY_LIST;   // line 33 - unchecked, unsuppressed
}
...
if (m_list == EMPTY_ARRAY_LIST) {                            // line 63
    throw new UnsupportedOperationException();
}
```

A negative capacity means "immutable", encoded as identity against a shared `ArrayList<?>`. That
costs an unchecked cast and hides the protocol in a comparison. An explicit `private final boolean`
removes both, and then `EMPTY_ARRAY_LIST` can go away. Nothing outside the class passes a negative
capacity - the only such caller is the internal `EMPTY` instance.

## 5. No tests

There is no `ListMapTest`. Whatever is done here should start with one: insertion order, `put`
replacing versus appending, the copy constructor's isolation, `asList` immutability, `entryAt`, and
the empty map rejecting writes.

## Suggested order

1. `ListMapTest` covering the above, with the copy-constructor case failing.
2. Fix the copy constructor.
3. Field type to `ArrayList<Entry<K, V>>`, dropping both raw casts.
4. Explicit immutability flag, dropping the unchecked cast and `EMPTY_ARRAY_LIST`.
5. Make `asList()` unconditionally unmodifiable.

Blast radius is small: no subclasses, two `asList()` callers (both read-only), six copy-constructor
callers, and no caller passing a negative capacity.

lagergren:
Fixes #607.

`ListMap` had one correctness bug and three soundness problems, and no tests.

### The copy constructor shared mutable entries

```java
m_list = new ArrayList<>(map.m_list);   // copies references to the SimpleEntry objects
```

`put` replaces a known key through `entry.setValue()`, so writing an existing key in a copy wrote
it in the original. Every caller of that constructor is making a defensive copy - and
`ClassStructure.getTypeParams()` even documents its result as *"an ordered read-only map"* - so none
of them were getting what they asked for. `Component` line 3037 is the clearest case: the line above
it isolates a list correctly with `Collections.unmodifiableList(new ArrayList<>(...))`, and the map
copy on the next line shared state.

The constructor now copies the entries.

### Equality is untouched, and the tests prove it rather than assert it

This was the thing worth being careful about, so the equality tests were written and run **before**
any change, against the unmodified class. `ListMap` declares no `equals` or `hashCode`, so it
inherits `AbstractMap`'s, which compare keys to values through `entrySet()` - it never compared the
backing `ArrayList`. A copied entry is `equals` to the one it came from, so a copy still equals, and
hashes like, its original.

Baseline run against the unmodified class: 11 passed, and exactly the two copy-isolation tests
failed. All three equality tests were green before and after.

### Two raw casts, from one cause

```java
List<Entry<K,V>> list = (List) m_list;        // asList()
return (Iterator) m_list.iterator();          // the entry set
```

Both existed because the field was declared as `ArrayList<SimpleEntry<K, V>>` while the API speaks
`Entry`, and generics are invariant. Declaring the field `ArrayList<Entry<K, V>>` removes the need
for either - `put` still adds a `SimpleEntry`, which is an `Entry`. `getEntry`'s protected return
type narrows with it, which is free as nothing subclasses `ListMap`.

### Immutability by array identity

A negative capacity meant "immutable", encoded as identity against a shared `ArrayList<?>`, which
cost an unchecked cast and hid the protocol inside a comparison. An explicit `boolean` replaces it
and `EMPTY_ARRAY_LIST` goes away. The check also moves ahead of the key lookup, so an immutable map
rejects a write rather than quietly taking one through `setValue` - unreachable today, since
immutable maps are always empty.

### asList() was read-only only under -ea

```java
assert (list = Collections.unmodifiableList(list)) != null;
```

The wrapper was applied inside an assert, so the method returned an unmodifiable list under `-ea`
and a live view of the internals in production, contradicting its own javadoc. It is now
unconditional. Both callers only read.

### Tests

`ListMapTest` is new - there were none. 13 tests: equality and hashCode (against another ListMap,
against a `HashMap`, and across a copy), insertion order, replace-versus-append, copy isolation in
three forms, the shared empty map, and `asList`.

### Also here

The first commit lifts `ListMap.empty()` over from the FP8 JIT branch, where it was unrelated to
that work. `EMPTY` was declared raw, so each of its 13 uses was an unchecked raw-to-parameterized
conversion; it becomes a private `ListMap<?, ?>` reached through a generic `empty()`, so the one
unchecked cast lives in one place.

### Verification

`javatools` 392 tests, `javatools_utils` 120, no failures, forced with
`--rerun-tasks --no-build-cache`. Full `build` green, which compiles the whole XDK and so exercises
`ClassStructure`, `Component` and `TypeInfo` - the copy constructor's callers - end to end.

### Proof that compiler behaviour is unchanged

Tests passing would not catch code that silently *relied* on the shared entries - that would change
what the compiler emits, not crash it. So the whole XDK was built twice, once with master's
`ListMap` and once with this one, and every compiled module hashed:

```
find xdk/build/install/xdk/lib -name '*.xtc' | sort | xargs shasum
```

All 22 modules are byte-identical. Compiling the XDK exercises `ClassStructure`, `Component`,
`TypeInfoReal` and `TypeConstant` - every copy-constructor call site - so identical output means
nothing depended on a copy aliasing its original.

Note also that `SimpleEntry` is not replaced: both entry-creation sites still do
`new SimpleEntry<>(...)`, and only the list's declared type widens from
`ArrayList<SimpleEntry<K, V>>` to `ArrayList<Entry<K, V>>`. The stored objects, and so their
`equals` and `hashCode`, are unchanged.


javatools_utils/src/main/java/org/xvm/util/ListMap.java
```diff
@@ -105,15 +119,20 @@ protected SimpleEntry<K,V> getEntry(Object key) {
...
      * The contents of the map are stored in an ArrayList of SimpleEntry
      * objects.
      */
-    private final ArrayList<SimpleEntry<K, V>> m_list;
+    private final ArrayList<Entry<K, V>> m_list;
+
+    /**
+     * True iff this map refuses to take entries; see {@link #ListMap(int)}.
+     */
+    private final boolean f_fImmutable;
```
ggleyzer, line 135: I don't see a reason for adding this field. The only place it's set as "true" is the constructor with a negative initial size. I would suggest adding a private constructor taking a boolean that is used by EMPTY. Then the old logic:

    if (m_list == EMPTY_ARRAY_LIST) {
            throw new UnsupportedOperationException();
        }

would be sufficient

ggleyzer (approved):

PR #609, opened 2026-09-17
Disallow property narrowing

ggleyzer:
The compiler used to allow property narrowing in the situation like one found in Version.x

        if (parent != Null && parent.form == Number) {
            buf.add(this.form == Number ? '.' : '-');
        }

Our reasoning was that since a Version is a constant, the value cannot change from one line to another.

However, Cam made an observation, that if a property getter is overridden on a subclass, this assumption may no longer be correct.

In that light, we decided to harden the compiler until a better way to express the writer's intent/assumptions is found. That caused the need to fix a number of places in Ecstasy code where that assumption no longer holds.

In the large number of places the change looks good, even better than original. Some are on par and only few are a bit more verbose.


lib_web/src/main/x/web.x
```diff
@@ -589,9 +589,8 @@ module web.xtclang.org {
          */
         ResponseOut makeResponse() {
             import responses.SimpleResponse;
-            return text == Null
-                    ? new SimpleResponse(status)
-                    : new SimpleResponse(status, text);
+            return new SimpleResponse(status, text?) :
```
cpurdy, line 592: reads better now, I think

lib_xenia/src/main/x/xenia/Catalog.x
```diff
@@ -93,7 +93,7 @@ const Catalog(WebApp webApp, WebServiceInfo[] services, Class[] sessionAnnos) {
         /**
          * The HTTP Method.
          */
-        HttpMethod? httpMethod.get() = method.is(Observe|Intercept) ? method.httpMethod : Null;
+        HttpMethod? httpMethod.get() = method.is(Observe|Intercept)?.httpMethod : Null;
```
cpurdy, line 96: 👍

lib_oodb/src/main/x/oodb/DBProcessor.x
```diff
@@ -465,10 +465,7 @@ interface DBProcessor<Message extends immutable Const>
          *         message processing has not completed before the next iteration would begin
          */
         conditional (Duration repeatInterval, Policy repeatPolicy) isRepeating() {
-            if (repeatInterval != Null) {
-                return True, repeatInterval, repeatPolicy;
-            }
-
+            return True, repeatInterval?, repeatPolicy;
```
cpurdy, line 468: I remember early on we had a test or example like this, with two returns. Actually, it may have been in the core lib.

lib_net/src/main/x/net/Uri.x
```diff
@@ -550,12 +550,12 @@ const Uri
...
     @RO Position endPosition.get() {
-        return fragment  != Null ? Section.Fragment .end(fragment)
-             : opaque    != Null ? assert
-             : query     != Null ? Section.Query    .end(query)
-             : path      != Null ? Section.Path     .end(path)
-             : authority != Null ? Section.Authority.end(authority)
-             : scheme    != Null ? Section.Scheme   .end(scheme)
+        return Section.Fragment .end(fragment?)
+             : opaque != Null   ? assert
+             : Section.Query    .end(query?)
```
cpurdy, line 555: This one does raise a question in my mind: Where the grounding point is for the short-circuit. Is it the same colon used by the ternary, or is it the expression as a whole? I worry a tiny bit when I have to think about it ...

lib_ecstasy/src/main/x/ecstasy/reflect/Version.x
```diff
@@ -695,7 +695,7 @@ const Version
     Appender<Char> appendTo(Appender<Char> buf, Boolean suppressBuild=False) {
         parent?.appendTo(buf, True);
 
-        if (parent != Null && parent.form == Number) {
+        if (parent?.form == Number) {
```
cpurdy, line 698: much better

lib_ecstasy/src/main/x/ecstasy/reflect/Argument.x
```diff
@@ -22,7 +22,7 @@ const Argument<Referent extends Passable>(Referent value, String? name = Null) {
             return 2;
 
         default:
-            return value.is(Stringable) ? value.estimateStringLength() : 0;
+            return value.is(Stringable)?.estimateStringLength() : 0;
```
cpurdy, line 25: much better

lib_ecstasy/src/main/x/ecstasy/lang/Error.x
```diff
@@ -61,7 +61,10 @@ const Error {
     /**
      * The string identity of the error code corresponding to the error.
      */
-    String code.get() = errorCode.is(ErrorCode)?.code : errorCode;
+    String code.get() {
+        var errorCode = this.errorCode;
```
cpurdy, line 65: as discussed ... still need to think about this scenario some more, but this will work for now

lib_ecstasy/src/main/x/ecstasy/lang/src/Lexer.x
```diff
@@ -2483,6 +2483,7 @@ class Lexer
 
         @Override
         public String toString() {
+            Object value = this.value;
```
cpurdy, line 2486: reminds me of:
 val val = val;
🤣

lib_ecstasy/src/main/x/ecstasy/lang/src/ast/NamedTypeExpression.x
```diff
@@ -125,16 +125,14 @@ const NamedTypeExpression(Token[]?          moduleNames,
...
             access.id.text.appendTo(buf);
         }
 
-        if (noNarrow != Null) {
-            noNarrow.id.text.appendTo(buf);
-        }
+        noNarrow?.id.text.appendTo(buf);
 
-        if (params != Null) {
+        if (var params ?= this.params) {
```
cpurdy, line 135: 👍

javatools_bridge/src/main/x/_native/reflect/RTParameter.x
```diff
@@ -7,9 +7,8 @@ const RTParameter<ParamType>(Int ordinal, String? name, Boolean formal, Boolean
         implements Parameter<ParamType> {
     @Override
     conditional String hasName() {
-        return name == Null
-                ? False
-                : (True, name);
+        return (True, name?);
```
cpurdy, line 10: I wonder if we could alternatively do:
return (True, name?) : False;
might read better

cpurdy (approved):
Necessary change. Still some improvements/refinements to consider long term.

javatools_bridge/src/main/x/_native/reflect/RTParameter.x
ggleyzer: I agree it would. However, it doesn't compile :)
We may want to improve the compiler for this one

lib_net/src/main/x/net/Uri.x
ggleyzer: Hmm.. may need to test more thoroughly

PR #612, opened 2026-09-18
Migrate the remaining MethodTypeDesc.of calls to md()

lagergren:
Mechanical, no behaviour change. Stacked on #610 — **review and merge that first.**

`Builder.md()` was introduced alongside the JIT builder work in #610, where 51 call sites moved to it: a spelled-out `MethodTypeDesc.of(...)` is the bulkiest part of an `invokestatic` and kept forcing the call onto a second line.

```java
.invokestatic(CD_JavaFloat, "float16ToFloat", md(CD_float, CD_short));
```

Verified: full `build`, `spotlessCheck` run on its own, and the JIT test suite all green. Ju


javatools/src/main/java/org/xvm/asm/op/FBind.java
```diff
@@ -50,6 +49,7 @@
 
 import static org.xvm.util.Handy.readPackedInt;
 import static org.xvm.util.Handy.writePackedLong;
```
ggleyzer, line 52: We keep imports in alphabetical order; this should go up

javatools/src/main/java/org/xvm/asm/op/I_Set.java
```diff
@@ -37,6 +36,7 @@
 
 import static org.xvm.util.Handy.readPackedInt;
 import static org.xvm.util.Handy.writePackedLong;
```
ggleyzer, line 39: ditto

ggleyzer (changes requested):
I like it; please make sure the imports are alphabetical

javatools/src/main/java/org/xvm/asm/op/FBind.java
lagergren: Fixing

lagergren:
Had merge conflicts because problem with brain. Moved everything to a new PR including alphabetical import order.

PR #614, opened 2026-09-18
fix: avoid cloning children twice when merging an existing module

### LINKED ISSUE #579: loadModule() of an already-present module fails an adoption assertion in FileStructure.merge
## Summary

`InterpreterConnector.loadModule(ECSTASY_MODULE)` fails an assertion while merging the module into
the native container's `FileStructure`:

```
java.lang.AssertionError
    at org.xvm.asm.Component.adoptChildren(Component.java:1002)
    at org.xvm.asm.Component.linkSibling(Component.java:990)
    at org.xvm.asm.Component.addChild(Component.java:878)
    at org.xvm.asm.Component.cloneChildren(Component.java:2166)
    at org.xvm.asm.FileStructure.merge(FileStructure.java:214)
    at org.xvm.runtime.NativeContainer.createFileStructure(NativeContainer.java:868)
    at org.xvm.api.InterpreterConnector.loadModule(InterpreterConnector.java:73)
```

The assertion is that an adopting component has no children of its own:

```java
protected void adoptChildren(Component that) {
    // the child can't have any of its own children; that "merge" functionality is simply
    // not supported by this operation
    assert m_abChildren  == null;
    assert m_childByName == null;
```

`merge` supersedes an existing *fingerprint* of the same id, but not an existing real module:

```java
var moduleExist = getModule(moduleClone.getIdentityConstant());
if (moduleExist != null && moduleExist.isFingerprint()) {
    removeChild(moduleExist);
}
addChild(moduleClone);
moduleClone.cloneChildren(module.children());
```

so cloning children into a structure that already holds that module reaches `addChild` with a name
that already has a sibling, and the deep-cloned newcomer is not childless.

The trigger is loading a module the native container already has: ecstasy is present from bootstrap,
so merging ecstasy collides with itself. An ordinary application module does not, which is why the
CLI path does not hit this.

## Reproducer

Construct one `InterpreterConnector` over a repository containing `ecstasy.xtc`,
`javatools_bridge.xtc` and `javatools_turtle.xtc` (the extracted XDK layout), and call
`loadModule(Constants.ECSTASY_MODULE)`. It fails on the **first** connector.

`InterpreterConnectorTest.parallelConnectorsLoadIndependentNativeContainers` fails for this reason,
4 runs out of 4 — the parallelism in that test is incidental.

## Applies to master

Verified: the identical failure, with the same stack, reproduces on `origin/master` against the same
repository directory. Only the line numbers differ.

## Note on why this hid

The test is gated on `assumeTrue(systemModulesAvailable())`, so a working tree without a compiled XDK
skips it and the suite still reports zero failures.

## Correction

This issue was originally filed as a parallel-construction race. That was wrong: it is sequential,
reproducible on the first connector, and present on master. Corrected here rather than opening a
second issue. For the record, the single-plane parallel case measured healthy — six compiles and
four runs issued concurrently against one engine complete cleanly with a clean ownership sweep.

---

## Update 2026-09-18 — still reproduces; original repro path is now stale

Re-verified on `origin/master` (`794bf23e6`). The defect is unchanged, but this report's entry
points have aged and no longer show it in the best light. Restating against the current world:

**`LspSupport` is gone**, replaced by `org.xvm.api.EmbeddingSupport`, and the
`InterpreterConnectorTest` cited above no longer exists. So the original reproducer reads as if it
depended on retired internals. It does not — the defect is reachable from today's public API and
from the shipped CLI.

**Reproduces through the current EmbeddingApi.** Using only `EmbeddingSupport.configure(...)` and
`EmbeddingSupport.run(...)`, no internal classes:

```
---- EmbeddingSupport.run(ecstasy.xtclang.org) ----
   THREW: java.lang.AssertionError
      at org.xvm.asm.Component.adoptChildren(Component.java:947)
      at org.xvm.asm.Component.linkSibling(Component.java:935)
      at org.xvm.asm.Component.addChild(Component.java:892)
      at org.xvm.asm.Component.cloneChildren(Component.java:2073)
      at org.xvm.asm.FileStructure.merge(FileStructure.java:241)
      at org.xvm.runtime.NativeContainer.createFileStructure(NativeContainer.java:763)
```

Note `InterpreterControl.prepareModule` serialises the module and reads it back before calling
`createFileStructure`, so this is not an aliasing artifact — identity alone is enough to collide.

**Reproduces from the shipped CLI**, with no embedding and no test code at all:

```
$ xec crypto.xtclang.org
java.lang.AssertionError
    at org.xvm.asm.Component.adoptChildren(Component.java:947)
    ...
    at org.xvm.runtime.NativeContainer.createFileStructure(NativeContainer.java:763)
```

**Scope is wider than ecstasy.** `NativeContainer.createFileStructure` pre-merges turtle, crypto,
net, web, native and then the application module, so *any* of those six crashes when it is itself
the application module. Ordinary application modules never collide, which is why normal use — and
`LspTest`, which only ever runs compiled user modules — never sees it.

**The API cannot contain it.** `EmbeddingSupport.run` guards with `catch (RuntimeException e)` and
logs `ERR_CREATE_APP_CONTAINER`. `AssertionError` is an `Error`, so it escapes that catch and
propagates to the embedder rather than becoming a logged error and a `null` return. Worth fixing
independently of the merge defect: an embedder should not receive an `Error` because a module name
was a library.

**Severity, stated plainly.** Nothing *needs* to run a system module as an application, so this is
robustness rather than a blocker. The correct outcome is the graceful `Missing method "run"` that
these modules produce once the merge is fixed — not an assertion failure.

**Proposed fix:** #614. It resolves both reproductions above. One open question there is what
`merge()` should do when the incoming module's children *differ* from the module already present —
the current proposal silently drops them — which is the semantics question raised on that PR.

mvanhorn:
# Description

First reproduce the structural defect in FileStructureTest with an in-memory module containing a package and nested class, then merge the module into a copy that already contains it. In FileStructure.merge, distinguish a newly inserted module (including fingerprint replacement) from a real-module sibling whose children addChild/linkSibling already adopted, and avoid cloning that shared child hierarchy a second time. Preserve the existing sibling model rather than removing every real module with the same identity; keep constant registration, synthesis, and fTakeFile primary-module/file metadata processing active instead of returning early.

Loading ecstasy through InterpreterConnector reaches NativeContainer.createFileStructure, which already copied ecstasy before merging the requested module again. FileStructure.merge adds the incoming module as a sibling, thereby adopting the existing child map, and then deep-clones children into that same map; nested duplicates reach Component.adoptChildren with populated children and fail an assertion. The report explicitly corrects its original race diagnosis: a first, sequential connector load reproduces it. The current clone retains that call chain, although the cited InterpreterConnectorTest no longer exists and LspSupport has been replaced by EmbeddingSupport.

Fixes #579

# Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Documentation or formatting/non-functional change

# Resolves/Fixes

Not applicable to this change.

# How Has This Been Tested?

Ungated, assertion-enabled unit regression: build a module with a package containing a class, copy it, and merge the same source module again; assert no adoption failure, intact nested children, and unchanged source hierarchy.
Repeat the merge and check nested child sibling chains for accidental duplication, not merely top-level module counts; valid module siblings may remain under the existing model.
Exercise both fTakeFile values, including a different initial primary module: true selects the merged module and its file metadata, false retains the previous primary. Verify constant ownership and validateConstants where applicable.
Merge an absent module and replace an existing fingerprint with a populated real module; confirm neither path skips child cloning or dependency fingerprint propagation.
Preserve conditional/version siblings and FileStructure(FileStructure) copy behavior; include structural and serialization round-trip checks using existing test utilities where valid.
Downstream validation: run `./gradlew :javatools:test --tests org.xvm.asm.FileStructureTest --rerun-tasks --no-build-cache`, then the existing XvmStructureTest and BundlerTest suites. Inspect JUnit XML for actual executed cases, failures, and skips; do not rely on a green console summary.
For the reported integration case, build `./gradlew :xdk:installDist` and use the current extracted core libraries (including crypto, net, and web, now explicitly loaded by NativeContainer). Run a sequential first connector load with assertions enabled, compare an ordinary application load, and exercise the documented LspTest/EmbeddingSupport runner path. A successful runner path alone does not prove that loading ecstasy directly is supported; report this distinction to the maintainer. Do not reintroduce a parallel-race test as the primary regression.
Run `./gradlew spotlessCheck` and inspect git status after implementation. No builds or tests were executed during this read-only planning lane.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

AI was used for assistance.


lagergren:
Thanks for this — I reproduced the defect independently and without AI, and it's real, not a test artifact.

It's also reachable from the shipped CLI with no test code involved. On master, any module that `NativeContainer.createFileStructure` has already merged fails when it is itself the application module. This is a rather strange use case, and I am not sure if it is intended, but I am trying to figure out if this would actually implicate a real problem with merge and module children on something that is not recursively build like the XDK, and I think it does. I am not the expert here, though, so I will defer opinion to @cpurdy.

```
$ xec crypto.xtclang.org
java.lang.AssertionError
    at org.xvm.asm.Component.adoptChildren(Component.java:947)
    at org.xvm.asm.Component.linkSibling(Component.java:935)
    at org.xvm.asm.Component.addChild(Component.java:892)
    at org.xvm.asm.Component.cloneChildren(Component.java:2073)
    at org.xvm.asm.FileStructure.merge(FileStructure.java:241)
    at org.xvm.runtime.NativeContainer.createFileStructure(NativeContainer.java:763)
```

Same for `ecstasy.xtclang.org`. With your change both get through to a benign "Missing method run". Your analysis of the mechanism matches what I see.

I ran the suite (the description notes tests weren't executed in your lane, so for the record): with the fix, 19 tests / 0 failures / 6 skipped — the 6 being the pre-existing XDK-dependent ones. Reverting only the one-line change gives 6 failures, all at `Component.adoptChildren:947`. So the new tests are genuinely load-bearing and worth keeping.

## Where I think this may not be the whole fix

Every new test merges **identical** content. `testMergePreservesVersionSiblingsAndFileCopy` looks like it covers divergence, but v1.0 and v2.0 are both `createNestedModule("Test")`, so their children are the same. That means the suite can't distinguish "correctly skipped a redundant clone" from "silently discarded the incoming hierarchy".

Here is an example: 

```
incoming children  = [other]
dest BEFORE merge  = [util]
dest AFTER  merge  = [util]        <- "other" is gone
  module sibling children = [util]
  module sibling children = [util]
```

For this kind of layout, destination holds real module `Test` containing package `util`; incoming `Test` contains package `other`:

For #579 that's harmless — the incoming module *is* the existing one, so there's nothing to lose. But `merge()` doesn't document a same-content precondition, and nothing warns when it's violated. The loud crash has become a silent drop.

## A second candidate fix (maybe)

@cpurdy Please weigh in on this. Sketch is as follows: `addChild` routes name collisions to `linkSibling`, so the component model already intends colliding children to become siblings. The assertion fires only because `cloneChild` populates grandchildren *before* the child is added, so `adoptChildren` sees a non-empty map. Adding the child first and cloning grandchildren afterwards also fixes the crash:

|                                       | this PR (skip clone) | reorder `cloneChild` |
|---------------------------------------|----------------------|----------------------|
| crash fixed                           | yes                  | yes                  |
| divergent children                    | silently dropped     | preserved as siblings |
| child `siblingCount` after same merge | 1                    | 2                    |
| this PR's tests                       | pass                 | 6 fail, `expected: <1> but was: <2>` |

So the `assertEquals(1, siblingCount(...))` assertions aren't neutral — they encode "skip" as the intended semantics. That may well be the right choice, since duplicating identical children into 2-long sibling chains is just bloat. But it's worth deciding deliberately rather than having a test settle it.

## Suggestion

Keep the 8 tests. Before merging, could we settle what `merge()` should do when the module is already present as a real module:

1. no-op on the children (this PR), and document that on `merge()` — plus ideally fail loudly rather than silently when the incoming content differs; or
2. link the incoming children as siblings, which preserves them.

@cpurdy — what are the intended semantics here? Specifically: when `FileStructure.merge()` is handed a module whose identity already exists in the file as a *real* (non-fingerprint) module, should the merge be a no-op for that module's children, or should the incoming children be linked into the existing shared child map as sibling chains? And relatedly, is merging a same-identity module with *different* content a supported call at all, or should it be rejected outright? The answer decides which of the two fixes above is correct, and what the new tests should assert.

Whichever we pick, a divergent-content test pins it. Something like:

```java
@Test
public void testMergeDivergentModuleDoesNotSilentlyLoseChildren() {
    FileStructure dest = new FileStructure(createNestedModule("Test").getModule(), false);

    FileStructure    incomingFile = new FileStructure("Test");
    PackageStructure other        = incomingFile.getModule()
            .createPackage(Constants.Access.PUBLIC, "other", null);
    other.createClass(Constants.Access.PUBLIC, Component.Format.CLASS, "Thing", null);

    dest.merge(incomingFile.getModule(), true, false);

    boolean found = false;
    for (Component m = dest.getModule(); m != null && !found; m = m.getNextSibling()) {
        found = m.getChildByNameMap().containsKey("other");
    }
    assertTrue(found, "merged module's own children were discarded");
}
```

That currently fails against this PR and passes under the reorder variant, so it's a clean discriminator once the intended behaviour is agreed.

Also minor: `hasChildren()` is a slightly indirect guard for the stated intent. It's true only because the clone adopted a sibling's map, so testing "was this linked as a sibling rather than inserted fresh" would say what the code means more directly.

mvanhorn:
Thanks for reproducing it independently, and for running the suite. The `xec crypto.xtclang.org` stack is the same path I hit, and having the revert numbers on record (6 failures, all at `adoptChildren:947`) is more than I could show from my lane.

On the semantics question my vote is option 1, no-op on the children, plus failing loudly when the incoming content differs.

The AssertionError exists because the clone adopted a sibling's child map, so two components shared one map and the second adopt tripped the invariant. Option 2 preserves the incoming children by linking them into that same shared map, which keeps the aliasing that caused the crash and moves the ambiguity to lookup time: after a divergent merge you would have two `other` packages reachable through the sibling chain with nothing saying which wins. That seems worse than refusing the operation.

Which leads to the second half of your question, and I think it is the more important one: merging a same-identity module with different content looks like a caller bug rather than a supported call. `NativeContainer.createFileStructure` is merging what it believes is the same module, so if the content diverges something upstream is wrong, and a silent no-op hides that just as much as silently dropping children does. So: no-op for the already-present real module, an explicit failure when the incoming content is not equivalent, and both documented on `merge()`.

If that is the direction @cpurdy confirms, your discriminator test is still the right test, just inverted: it should assert the divergent merge is rejected rather than that the children survive. Happy to add it either way once the semantics are settled, since as you say it fails against this PR today and I would rather not pin in an assertion before we agree what it should assert.

Agreed on `hasChildren()`. It is true only as a side effect of the adopted map, so it tests a symptom. I will change it to check what the code actually means.

@cpurdy, the question is above if you have a view.

javatools/src/main/java/org/xvm/asm/FileStructure.java
```diff
@@ -238,7 +238,14 @@ public void merge(ModuleStructure module, boolean fSynthesize, boolean fTakeFile
         }
 
         addChild(moduleClone);
-        moduleClone.cloneChildren(module.children());
+        // Newly inserted modules, including fingerprint replacement, still have an empty child
```
ggleyzer: I would suggest a simpler a comment. For example:

// prevent duplication of already cloned children

ggleyzer (changes requested):

javatools/src/main/java/org/xvm/asm/FileStructure.java
mvanhorn: Done, taken as suggested in 5a4420e.

ggleyzer (approved):

mvanhorn:
Thanks for merging, @ggleyzer. Reproducing it in `FileStructureTest` first made the newly-inserted vs already-present distinction much easier to argue.

PR #615, opened 2026-09-18
Keep the text an IntLiteral was written as

lagergren:
This was discovered when I was hooking up LSP integration and ran the JIT test with the interpreter - they work fine with the JIT, but we were missing the hText for IntNHandle for the code on the interpreter side. 

`IntNHandle`'s constructor accepts the literal's text handle and drops it:

```java
public IntNHandle(TypeComposition clazz, PackedInteger piValue, StringHandle hText) {
    super(clazz);
    assert piValue != null;
    m_piValue = piValue;
    // hText is never assigned
}
```

`m_hText` is therefore always null, so `getText()` always takes its lazy
fallback and regenerates the text from the `PackedInteger`:

```java
m_hText = hText = xString.makeHandle(m_piValue.toString());
```

That loses everything the numeric value does not carry.

### Impact

`IntLiteral` is documented as `const IntLiteral(String text)` with
`String toString() = text`, so it is meant to preserve exactly how it was
written. Interpreted, it did not:

| written  | before   | after    |
|----------|----------|----------|
| `+1`     | `1`      | `+1`     |
| `+1_000` | `1000`   | `+1_000` |
| `0x1F`   | `31`     | `0x1F`   |
| `+0x1F`  | `31`     | `+0x1F`  |
| `-1`     | `-1`     | `-1`     |
| `0`      | `0`      | `0`      |

`-1` and `0` were only correct by luck, since `PackedInteger.toString()`
happens to reproduce them.

The rest of the chain was already right: the compiler stores the source text
in the pool and `createConstHandle` fetches it with
`constVal.getStringConstant()` — the constructor just threw it away. The JIT
reads the same constant through a different path and prints `+1` correctly,
so the two backends disagreed.

`xFPLiteral.FPNHandle` already assigns the field, which is why `FPLiteral`
was unaffected (`+1.5` printed as `+1.5` throughout).

### Fix

One line — assign `m_hText` in the constructor.

### Test

`testLiteralText()` in `manualTests/src/main/x/literals.x`, covering the forms
above, the `Signum.factor` path, and two `FPLiteral` cases so the working side
stays working.

`TestLiterals` is in `testModuleNames`, so it runs through
`runSequential` → `runCiTestTasks` **interpreted**. That matters: there is
already an equivalent assertion in `jit_tests/basic/enumTests.x`
(`Signum.Positive.factor.toString() == "+1"`), but that suite only runs under
`--jit`, where the bug does not exist — which is why this went unnoticed.

Verified the test is load-bearing by reverting the one-line fix and re-running:

```
IllegalState: "plusOne.toString()  == \"+1\"": plusOne.toString()=1
    at testLiteralText() (literals.x:431)
BUILD FAILED
```

`./gradlew spotlessCheck` clean and full `./gradlew build` green with the fix.


javatools/src/main/java/org/xvm/runtime/template/numbers/xIntLiteral.java
```diff
@@ -440,6 +440,7 @@ public IntNHandle(TypeComposition clazz, PackedInteger piValue, StringHandle hTe
             assert piValue != null;
 
             m_piValue = piValue;
+            m_hText   = hText;
```
cpurdy, line 443: useful

cpurdy (approved):
nice catch

PR #616, opened 2026-09-18
Contain a failing compile or module start inside the embedding API

lagergren:
`EmbeddingSupport` has two entry points that run caller-supplied work, and both
promise a `null` on failure with the reason reported to `errs`. Neither guard
holds:

```java
// run()      -- an Error is not a RuntimeException
} catch (RuntimeException e) { ... return null; }

// compile()  -- narrower still; a plain RuntimeException escapes too
} catch (LauncherException e) { ... return null; }
```

So a failure can reach the host as a bare `Error` instead of the
`ERR_CREATE_APP_CONTAINER` / `ERR_INTERNAL` report the API otherwise gives.

This is not hypothetical. #579 makes it happen today:

```
---- EmbeddingSupport.run(ecstasy.xtclang.org) ----
   THREW: java.lang.AssertionError
      at org.xvm.asm.Component.adoptChildren(Component.java:947)
      ...
      at org.xvm.runtime.NativeContainer.createFileStructure(NativeContainer.java:763)
```

And `AssertionError` is a production error class here, not a test artifact: the
shipped `xec` launcher hardcodes `DEFAULT_JVM_OPTS='"-ea" ...'`, and there are
~1,600 `assert` statements across `asm`, `compiler`, `runtime` and `javajit`.

### Change

Catch `AssertionError` alongside `RuntimeException` in both methods, and document
the `null` return that neither javadoc mentioned. `LauncherException` already
extends `RuntimeException`, so `compile()` loses nothing by widening.

### What is deliberately *not* caught

`Error` is not caught wholesale. A `VirtualMachineError` says the JVM is in
trouble, not that this input was bad:

- `OutOfMemoryError` — the handler would have to allocate to describe it, so the
  recovery can fail while recovering, and the failure is not attributable to this
  call anyway.
- `StackOverflowError` — `compile()` can be made to throw one with ~3,000 nested
  parens. Tempting to report as "expression too deeply nested", but it is the
  same category of unstable-to-handle condition, so it is left to propagate.

That is a deliberate line: no `VirtualMachineError` is explicitly handled.

### Verification

`run()`, against the live #579 fault:

```
   returned  : null
   errs count: 1
   logged    : EMB-3: Unable to create the application container:
               Unable to start ecstasy.xtclang.org. Cause: java.lang.AssertionError
```

`compile()`, probed with nine pathological sources — empty, garbage, no `module`
keyword, unterminated block, duplicate modules, bad annotation, recursive typedef,
3,000-deep nesting, missing module name. Eight return `null` with errors logged;
the ninth is the `StackOverflowError` above, left to propagate by design.

`./gradlew spotlessCheck` clean and full `./gradlew build` green.

### On testing

No unit test, and I would rather say why than quietly omit one. Reproducing the
`run()` fault needs a compiled XDK *and* a way to inject a failure into the
container start, and `run()` has no seam for the latter — `ensureConnector()` and
`*Control.create()` are called directly. The only deterministic `Error` available
today is the #579 bug itself, which disappears once #614 lands, so a test written
against it would quietly become vacuous.

If you want this pinned permanently, the options are a small injection seam on the
control-creation step, or an integration check alongside `LspTest` (already a
`main()` rather than a JUnit test, and already requiring the XDK). Happy to add
either — just say which you prefer.


thegridman (approved):

lagergren:
Thanks!

